PR #147, opened 2017-04-28
Makes CustomTabsServiceConnection class static | Stress testing

iambmelt:
This change converts the `CustomTabsServiceConnection` to a `static class` and removes strong references to its enclosing `class`

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/AuthenticationActivity.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AuthenticationActivity.java
@@ -28,11 +28,12 @@ import android.content.ComponentName;
 import android.content.Intent;
 import android.net.Uri;
 import android.os.Bundle;
-import android.support.annotation.NonNull;
 import android.support.customtabs.CustomTabsClient;
 import android.support.customtabs.CustomTabsIntent;
 import android.support.customtabs.CustomTabsServiceConnection;
 import android.support.customtabs.CustomTabsSession;
+
+import java.lang.ref.WeakReference;
 
 /**
  * Custom tab requires the device to have a browser with custom tab support, chrome with version >= 45 comes with the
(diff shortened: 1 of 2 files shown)


msal/src/main/java/com/microsoft/identity/client/AuthenticationActivity.java
```diff
@@ -130,22 +131,38 @@ private void warmUpCustomTabs() {
...
-    @NonNull
-    private CustomTabsServiceConnection createCustomTabsServiceConnection() {
-        return new CustomTabsServiceConnection() {
-            @Override
-            public void onCustomTabsServiceConnected(ComponentName name, CustomTabsClient client) {
-                mCustomTabsServiceIsBound = true;
-                mCustomTabsClient = client;
-                mCustomTabsClient.warmup(0L);
-                mCustomTabsSession = mCustomTabsClient.newSession(null);
+    private static class MsalCustomTabsServiceConnection extends CustomTabsServiceConnection {
```
weijjia: I'm a little bit worried about the implementation 1) connection class having reference to enclosing class doesn't look good :) 2) in line 121, we call bindCustomTabsService, which is an asynchronous call, but we pass in the customtab session right after it to intent builder, is it possible that it's not connected yet? 3) I don't think we're using warm up at all. If you look at the onResume, it's creating a new customTabIntent and that is the intent used to launch URL.

iambmelt:
@weijjia f4289310936c9d840b90f19f0f553cf9ea8f71b6 shows what this `Activity` would look like if we could skip `warmup()`

weijjia:
so much cleaner :)

msal/src/main/java/com/microsoft/identity/client/AuthenticationActivity.java
```diff
@@ -125,29 +131,70 @@ private void warmUpCustomTabs() {
...
+
+        private final WeakReference<AuthenticationActivity> mAuthenticationActivityWeakReference;
```
weijjia: Like I mentioned above, having a reference to enclosing class is a bad sign. If it won't cause too much trouble, i would suggest we have the customsTabSession and customTabsClient declared inside of the connection, once it's connected you'll set those values. We can do connection.getTabSesson or connection.getTabsClient for it's enclosing class to use.

PR #151, opened 2017-05-01
Use OauthConstants for error code comparison

weijjia:
Update error code comparison for invalid_grant, use oauth2constants instead of the error code in exception class directly.


iambmelt:
Nit: Can we make `MsalUiRequiredException.INVALID_GRANT = OauthConstants.ErrorCode.INVALID_GRANT` so we're not maintaining duplicative copies of `"invalid_grant"`?

weijjia:
Nope, MsalUiRequiredException is public facing API. I think it would be better for us to expose the error code directly instead of having another layer for developer to find what is the value for the constant.

Issue #150, opened 2017-05-01
Crash in DevUtil MainActivity when Do Not Keep Activities is turned on


iambmelt:
```java
FATAL EXCEPTION: main
                                                                                    Process: com.microsoft.identity.client.sample, PID: 3332
                                                                                    java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
                                                                                        at android.support.v4.app.FragmentManagerImpl.checkStateLoss(FragmentManager.java:1842)
                                                                                        at android.support.v4.app.FragmentManagerImpl.enqueueAction(FragmentManager.java:1860)
                                                                                        at android.support.v4.app.BackStackRecord.commitInternal(BackStackRecord.java:649)
                                                                                        at android.support.v4.app.BackStackRecord.commit(BackStackRecord.java:609)
                                                                                        at com.microsoft.identity.client.sample.MainActivity.attachFragment(MainActivity.java:222)
                                                                                        at com.microsoft.identity.client.sample.MainActivity.onNavigationItemSelected(MainActivity.java:192)
                                                                                        at com.microsoft.identity.client.sample.MainActivity$2.onSuccess(MainActivity.java:317)
                                                                                        at com.microsoft.identity.client.PublicClientApplication$1.onSuccess(PublicClientApplication.java:668)
                                                                                        at com.microsoft.identity.client.BaseRequest$2.run(BaseRequest.java:278)
                                                                                        at android.os.Handler.handleCallback(Handler.java:751)
                                                                                        at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                                        at android.os.Looper.loop(Looper.java:154)
                                                                                        at android.app.ActivityThread.main(ActivityThread.java:6119)
                                                                                        at java.lang.reflect.Method.invoke(Native Method)
                                                                                        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
                                                                                        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
```

iambmelt:
Closing as a no-op. No plans to try and support DNKA

Issue #149, opened 2017-04-30
Crash in AuthenticationActivity when uiBuilder class is not initialized


addev-ashish:
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'com.microsoft.identity.client.UiEvent$Builder com.microsoft.identity.client.UiEvent$Builder.setUserDidCancel()' on a null object reference
                                                                                                   at com.microsoft.identity.client.AuthenticationActivity.cancelRequest(AuthenticationActivity.java:210)
                                                                                                   at com.microsoft.identity.client.AuthenticationActivity.onResume(AuthenticationActivity.java:173)
                                                                                                   at android.app.Instrumentation.callActivityOnResume(Instrumentation.java:1259)
                                                                                                   at android.app.Activity.performResume(Activity.java:6347)
                                                                                                   at android.app.ActivityThread.performResumeActivity(ActivityThread.java:3110)
                                                                                                   at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3152) 
                                                                                                   at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2499) 
                                                                                                   at android.app.ActivityThread.access$900(ActivityThread.java:157) 
                                                                                                   at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1356) 
                                                                                                   at android.os.Handler.dispatchMessage(Handler.java:102) 
                                                                                                   at android.os.Looper.loop(Looper.java:148) 
                                                                                                   at android.app.ActivityThread.main(ActivityThread.java:5551) 
                                                                                                   at java.lang.reflect.Method.invoke(Native Method) 
                                                                                                   at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:730) 
                                                                                                   at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:620)

weijjia:
closed with PR #152

PR #152, opened 2017-05-01
Save Telemetry requestId if AuthenticationActivity is torn down.

iambmelt:
Saves the Telemetry requestId to the `onSaveInstanceState()` `Bundle` if the `AuthenticationActivity` is destroyed (say, if `Don't Keep Activities` is turned on).

Testing with DevUtil is blocked due to #150 - @addev-ashish to test via alternative means

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/AuthenticationActivity.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AuthenticationActivity.java
@@ -65,6 +65,8 @@ public final class AuthenticationActivity extends Activity {
         if (savedInstanceState != null) {
             Logger.verbose(TAG, null, "AuthenticationActivity is re-created after killed by the os.");
             mRestarted = true;
+            mTelemetryRequestId = savedInstanceState.getString(Constants.TELEMETRY_REQUEST_ID);
+            mUiEventBuilder = new UiEvent.Builder();
             return;
         }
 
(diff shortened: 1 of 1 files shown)


msal/src/main/java/com/microsoft/identity/client/AuthenticationActivity.java
```diff
@@ -65,6 +65,8 @@ protected void onCreate(final Bundle savedInstanceState) {
         if (savedInstanceState != null) {
             Logger.verbose(TAG, null, "AuthenticationActivity is re-created after killed by the os.");
             mRestarted = true;
+            mTelemetryRequestId = savedInstanceState.getString(Constants.TELEMETRY_REQUEST_ID);
+            mUiEventBuilder = new UiEvent.Builder();
```
weijjia: why we're adding this?

iambmelt: See #149

weijjia: seems like you can move the mUiEventBuilder before the check for saveInstanceState != null

iambmelt: What's the advantage? Making that change requires inverting the order of [these two checks](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/AuthenticationActivity.java#L64-L75) and I _still_ need to conditionally choose whether I am going to hydrate `mRequestId` from `savedInstanceState` or from `getIntent()`. The consequences of inverting these checks are also not immediately apparent to me: they both control early `return/sendError()` conditions: I'm inclined to leave as-is

weijjia: ah sorry i overlooked the other check.

Issue #148, opened 2017-04-30
Logcat is getting spammed by telemetry code when authority is not passed in


addev-ashish:
05-03 17:37:55.628 31003-31003/com.microsoft.identity.client.developersample E/MsalUtils: MSAL 0.1.0 Android 23 [2017-05-04 00:37:55] Url is invalid java.net.MalformedURLException
                                                                                              at java.net.URL.<init>(URL.java:152)
                                                                                              at java.net.URL.<init>(URL.java:125)
                                                                                              at com.microsoft.identity.client.MsalUtils.getUrl(MsalUtils.java:409)
                                                                                              at com.microsoft.identity.client.HttpEvent.sanitizeUrlForTelemetry(HttpEvent.java:55)
                                                                                              at com.microsoft.identity.client.ApiEvent.<init>(ApiEvent.java:47)
                                                                                              at com.microsoft.identity.client.ApiEvent.<init>(ApiEvent.java:35)
                                                                                              at com.microsoft.identity.client.ApiEvent$Builder.build(ApiEvent.java:273)
                                                                                              at com.microsoft.identity.client.PublicClientApplication.getUsers(PublicClientApplication.java:234)
                                                                                              at com.microsoft.identity.client.developersample.MainActivity.onCreate(MainActivity.java:117)
                                                                                              at android.app.Activity.performCreate(Activity.java:6272)
                                                                                              at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1108)
                                                                                              at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2387)
                                                                                              at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2494)
                                                                                              at android.app.ActivityThread.access$900(ActivityThread.java:157)
                                                                                              at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1356)
                                                                                              at android.os.Handler.dispatchMessage(Handler.java:102)
                                                                                              at android.os.Looper.loop(Looper.java:148)
                                                                                              at android.app.ActivityThread.main(ActivityThread.java:5551)
                                                                                              at java.lang.reflect.Method.invoke(Native Method)
                                                                                              at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:730)
                                                                                              at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:620)

PR #153, opened 2017-05-01
Do not attempt to parse null or empty URLs (when processing Authorities)

iambmelt:
This change seeks to quiet the amount of `Logger` output by not allowing the `MalformedURLException` to be thrown in instances where an Authority `URL` is not passed-in

### DIFF
--- a/msal/src/telemetry/java/com/microsoft/identity/client/ApiEvent.java
+++ b/msal/src/telemetry/java/com/microsoft/identity/client/ApiEvent.java
@@ -44,7 +44,9 @@ final class ApiEvent extends Event {
         if (null != builder.mRequestId) {
             setProperty(EventProperty.REQUEST_ID, builder.mRequestId);
         }
-        setProperty(EventProperty.AUTHORITY_NAME, HttpEvent.sanitizeUrlForTelemetry(builder.mAuthority));
+        if (!MsalUtils.isEmpty(builder.mAuthority)) {
+            setProperty(EventProperty.AUTHORITY_NAME, HttpEvent.sanitizeUrlForTelemetry(builder.mAuthority));
+        }
         setAuthorityType(builder.mAuthorityType);
         setProperty(EventProperty.UI_BEHAVIOR, builder.mUiBehavior);
         setProperty(EventProperty.API_ID, builder.mApiId);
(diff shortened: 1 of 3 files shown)


msal/src/telemetry/java/com/microsoft/identity/client/ApiEvent.java
```diff
@@ -44,7 +44,9 @@ private ApiEvent(Builder builder) {
         if (null != builder.mRequestId) {
             setProperty(EventProperty.REQUEST_ID, builder.mRequestId);
         }
-        setProperty(EventProperty.AUTHORITY_NAME, HttpEvent.sanitizeUrlForTelemetry(builder.mAuthority));
+        if (!MsalUtils.isEmpty(builder.mAuthority)) {
```
weijjia: why builder.mAuthority could be empty? if this is the API event, you should always have the authority.

iambmelt: This isn't strictly speaking true: calls to `PublicClientApplication#getUsers()` are not associated with an authority yet still create `ApiEvents`. Yes, the `PublicClientApplication` will have `mAuthorityString` initialized, but the result of this call isn't guaranteed to return results related to that authority.

iambmelt: This issue was raised by @addev-ashish in #148

weijjia: hm, but publicClientApplciation is always created with an authority. If it's not provided, we always use the default one; developer can also provide it thru metadata or constructor.

weijjia: ok, i think i know why this could happen.I would add the authority url for the API event in getUsers, we look up users by authority host, that maybe the info we want. But talk to @addev-ashish, he may have some other ideas.

iambmelt:
Per conversation with @addev-ashish @weijjia and @nazukj -- going to update `PublicClientApplication#getUsers()` to include the `host` of `mAuthorityString` in the `ApiEvent`

nazukj:
:shipit:

msal/src/telemetry/java/com/microsoft/identity/client/ApiEvent.java
weijjia: Is it still possible that authority in API event will be null?

weijjia:
Make sure if it's still required to have the null check for authority on API event, my understanding is that we don't need it any more. Rest looks good to me.

iambmelt:
@weijjia Cool. Check is removed by 6ccb6a37a029ec91b5a3215283a7333e1b5f47b8

iambmelt:
Will run CI to pass. Then I'll merge.

PR #155, opened 2017-05-02
#141: update authority validation

weijjia:
#141: update authority validation. When using the resolved endpoints from cache, we should check if authority is validated. If authority validation is turned on and the cached authority has not been validated yet, we should do instance discovery and tenant discovery.


addev-ashish:
:shipit:

Issue #159, opened 2017-05-03
Expose API to allow query parameters to be sent on token endpoint


weijjia:
Features could be deployed on testslice, MSAL today doesn't allow any query parameters to be send on token endpoint, which makes it harder to test those feature without modifying the source.

weijjia:
Merged with PR #162

PR #162, opened 2017-05-03
expose API to add slice parameters to token endpoint

weijjia:
#159: expose API to add slice parameters to token endpoint which makes it convenient to test out features that have not been deployed to prod without doing any source code changes in the library.

### DIFF
--- a/msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
+++ b/msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
@@ -51,6 +51,8 @@ import java.util.GregorianCalendar;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -277,7 +279,7 @@ final class MsalUtils {
      * @return The Map of the items decoded with the given delimiter.
      */
     static Map<String, String> decodeUrlToMap(final String url, final String delimiter) {
-        final Map<String, String> decodedUrlMap = new HashMap<>();
+        final Map<String, String> decodedUrlMap = new LinkedHashMap<>();
 
         // delimiter can be " "
         if (MsalUtils.isEmpty(url) || delimiter == null) {
@@ -347,7 +349,7 @@ final class MsalUtils {
             return url;
         }
 
-        final Set<String> queryParamsSet = new HashSet<>();
+        final Set<String> queryParamsSet = new LinkedHashSet<>();
         for (Map.Entry<String, String> entry : requestParams.entrySet()) {
             queryParamsSet.add(entry.getKey() + "=" + urlFormEncode(entry.getValue()));
         }
--- a/msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
+++ b/msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
@@ -51,8 +51,6 @@ import java.util.GregorianCalendar;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
```diff
@@ -277,7 +279,7 @@ static String getChromePackage(final Context context) {
      * @return The Map of the items decoded with the given delimiter.
      */
     static Map<String, String> decodeUrlToMap(final String url, final String delimiter) {
-        final Map<String, String> decodedUrlMap = new HashMap<>();
+        final Map<String, String> decodedUrlMap = new LinkedHashMap<>();
```
kpanwar: Why LinkedHashMap and LinkedHashSet?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -219,6 +220,14 @@ public void setComponent(final String component) {
     }
 
     /**
+     * To send extra parameters to token endpoint, which provides the possibility if developer wants to test features that have not been enabled on prod yet.
```
kpanwar: this actually encourages external developers to use the property. Consider updating to "Custom query parameters which may be sent to the STS for dogfood testing. This parameter should not be set by the developers as it may have adverse effect on the application."

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -219,6 +220,14 @@ public void setComponent(final String component) {
...
+     * @param sliceParameters The query parameters sent to token endpoint.
```
kpanwar: these parameters should also be sent to authorize endpoint.

msal/src/request/java/com/microsoft/identity/client/BaseRequest.java
```diff
@@ -254,6 +255,15 @@ private void buildRequestParameters(final Oauth2Client oauth2Client) {
         oauth2Client.addBodyParameter(OauthConstants.Oauth2Parameters.SCOPE, scope);
         oauth2Client.addBodyParameter(OauthConstants.Oauth2Parameters.CLIENT_INFO, "1");
         setAdditionalOauthParameters(oauth2Client);
+
+        // add slice parameters
+        if (!MsalUtils.isEmpty(mAuthRequestParameters.getSliceParameters())) {
```
kpanwar: these parameters need to be sent to authorize endpoint as well.

msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
weijjia: I want to keep the order.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
weijjia: update.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
weijjia: done.

msal/src/request/java/com/microsoft/identity/client/BaseRequest.java
weijjia: done.

msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
iambmelt: If the order is to be preserved, should we declare a more explicit return type for the method so that callers know? Otherwise, this behavior might be relied upon and if we refactor this implementation later we might unexpectedly break something

`static LinkedHashMap<String, String> decodeUrlToMap(...)`

weijjia: actually order doesn't matter when the request is sent over the wire, let me remove it.

kpanwar:
:shipit:

PR #170, opened 2017-05-05
add change log file

weijjia:
Add change log file for release


changelog
```diff
@@ -0,0 +1,3 @@
+Version 0.0.1
```
iambmelt: We're going live with `0.1.0` aren't we?

weijjia: omg. Thanks for catching it.

PR #184, opened 2017-05-18
Update travis to set up emulator and run instrumentation tests

weijjia:
Set up the emulator with sdk 21 (it's not stable to start emulator 23+ on travis which will cause build to be error out more often), and enable test run for travis CI. Disable predex.


weijjia:
Close the PR for now.

Issue #186, opened 2017-05-22
Telemetry: Consider tracking in-progress/completed Events in a single Map


iambmelt:
Currently `Events` are tracked in two different dictionaries:
`Telemetry#mEventsInProgress` - used to track `Events` that are in-progress.
`Telemetry#mCompletedEvents` - used to track `Events` that are complete, but have not yet been dispatched via `flush()`

Per conversation with @addev-ashish, we should explore conglomerating these `Events` into a single `Map` and adding a `boolean mIsCompleted` status flag.

iambmelt:
Closed w/ #187

PR #187, opened 2017-05-23
Telemetry Events now tracked in a single Map + tests (#186)

iambmelt:
This change removes `Telemetry#mEventsInProgress` and `Telemetry#mCompletedEvents` in favor of tracking telemetry `Event` data in a single `Map` instance.

Events are now tracked via their `Event.Builder` subclasses; a field `mIsCompleted` has been added. The completed status is updated in `Telemetry#stopEvent()`.

`Event` `startTime` and `stopTime` are tracked in `Telemetry#startEvent()` and `Telemetry#stopEvent()` respectively.

`Event.Builder` instances are built when `Telemetry#flush(String)` is called. `OrphanedEvents` are determined to be `Events` whose `flush()` has been called without first calling `stopEvent()`.

### DIFF
--- a/msal/src/telemetry/java/com/microsoft/identity/client/Event.java
+++ b/msal/src/telemetry/java/com/microsoft/identity/client/Event.java
@@ -46,6 +46,15 @@ class Event extends ArrayList<Pair<String, String>> {
         if (!EventConstants.EventName.DEFAULT_EVENT.equals(builder.mEventName)) {
             setProperty(EventProperty.EVENT_NAME, builder.mEventName);
         }
+        if (null != builder.mEventStartTime) {
+            setProperty(EventProperty.START_TIME, String.valueOf(builder.mEventStartTime));
+        }
+        if (null != builder.mEventStartTime) {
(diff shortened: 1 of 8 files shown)


msal/src/telemetry/java/com/microsoft/identity/client/Event.java
```diff
@@ -74,13 +83,44 @@ String getEventName() {
...
+
+    /**
      * Builder object used for Events.
      *
      * @param <T> generic type parameter for Builder subtypes.
      */
     abstract static class Builder<T extends Builder> {
 
         private final String mEventName;
+        private Long mEventStartTime;
```
weijjia: Does this need to be Long object type?

iambmelt: I used a `Long` object type because this value can be `null` (in the case of an `OrphanedEvent` or if an `Event` is created but never started) - it would represent a bug in the flow of the application, but it's something that is 'planned for' with the `OrphanedEvent` code

PR #200, opened 2017-06-05
Make sure we have no args constructor to prevent usage of unsafe

### LINKED ISSUE #199: Create "no-args" constructor for GSON POJOs to prevent using "sun.misc.Unsafe"
Please read this post: https://stackoverflow.com/a/18645370/950427.

If you are not using a "no-args" constructor in Gson, it will use: `sun.misc.Unsafe`. See https://github.com/google/gson/blob/master/UserGuide.md#writing-an-instance-creator

Also, users will no longer have to keep "unsafe" for proguard: https://github.com/google/gson/blob/master/examples/android-proguard-example/proguard.cfg#L10

Same idea from https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/839

jaredsburrows:
@addev-ashish
Closes #199

### DIFF
--- a/msal/src/cache/java/com/microsoft/identity/client/BaseTokenCacheItem.java
+++ b/msal/src/cache/java/com/microsoft/identity/client/BaseTokenCacheItem.java
@@ -31,13 +31,13 @@ import com.google.gson.annotations.SerializedName;
 abstract class BaseTokenCacheItem {
 
     @SerializedName("client_id")
-    final String mClientId;
+    String mClientId;
 
     @SerializedName("client_info")
-    final String mRawClientInfo;
+    String mRawClientInfo;
 
     @SerializedName("ver")
-    private final String mVersion = "1";
+    private String mVersion = "1";
 
     transient User mUser;
     transient ClientInfo mClientInfo;
@@ -48,11 +48,10 @@ abstract class BaseTokenCacheItem {
     abstract TokenCacheKey extractTokenCacheKey();
 
     /**
-     * No args constructor for use ill serialization for Gson to prevent usage of sun.misc.Unsafe
+     * No args constructor for use in serialization for Gson to prevent usage of sun.misc.Unsafe.
      */
     @SuppressWarnings("unused")
-    BaseTokenCacheItem() throws MsalClientException {
-        this(null, null);
+    BaseTokenCacheItem() {
     }
 
     /**
--- a/msal/src/cache/java/com/microsoft/identity/client/BaseTokenCacheItem.java
+++ b/msal/src/cache/java/com/microsoft/identity/client/BaseTokenCacheItem.java
@@ -48,6 +48,14 @@ abstract class BaseTokenCacheItem {
     abstract TokenCacheKey extractTokenCacheKey();
 
     /**
+     * No args constructor for use ill serialization for Gson to prevent usage of sun.misc.Unsafe
+     */
+    @SuppressWarnings("unused")
+    BaseTokenCacheItem() throws MsalClientException {
+        this(null, null);
+    }
+
+    /**
      * Constructor for creating the token cache item.
      */
     BaseTokenCacheItem(final String clientId, final String rawClientInfo) throws MsalClientException {
--- a/msal/src/main/java/com/microsoft/identity/client/MsalServiceException.java
+++ b/msal/src/main/java/com/microsoft/identity/client/MsalServiceException.java
@@ -89,7 +89,7 @@ public final class MsalServiceException extends MsalException {
     public final static String INVALID_INSTANCE = "invalid_instance";
 
     /**
-     * Request to server failed, but no error and error_description is returned back from the service。
+     * Request to server failed, but no error and error_description is returned back from the service.
      */
     public final static String UNKNOWN_ERROR = "unknown_error";
 
--- a/msal/src/cache/java/com/microsoft/identity/client/AccessTokenCacheItem.java
+++ b/msal/src/cache/java/com/microsoft/identity/client/AccessTokenCacheItem.java
@@ -55,6 +55,14 @@ final class AccessTokenCacheItem extends BaseTokenCacheItem {
     final String mRawIdToken;
 
     /**
+     * No args constructor for use ill serialization for Gson to prevent usage of sun.misc.Unsafe
+     */
+    @SuppressWarnings("unused")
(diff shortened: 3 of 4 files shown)


msal/src/main/java/com/microsoft/identity/client/MsalServiceException.java
```diff
@@ -89,7 +89,7 @@
     public final static String INVALID_INSTANCE = "invalid_instance";
 
     /**
-     * Request to server failed, but no error and error_description is returned back from the service。
+     * Request to server failed, but no error and error_description is returned back from the service.
```
jaredsburrows: Bad windows character broke the build for linux/mac locally.

jaredsburrows:
@iambmelt No problem. Thank you!

msal/src/cache/java/com/microsoft/identity/client/BaseTokenCacheItem.java
```diff
@@ -31,13 +31,13 @@
 abstract class BaseTokenCacheItem {
 
     @SerializedName("client_id")
-    final String mClientId;
+    String mClientId;
```
weijjia: i'm kind of against removing final. Final guarantees that those variables are only initialized in the constructor.

jaredsburrows: @weijjia You have to for the empty constructor. Plus, those values are only initialized if the server returns those values else they will be null.

weijjia: why you cannot do this(null, null, null) which calls the other constructor?

jaredsburrows: @weijjia If they are not initialized and no data comes back(no json), then it will be `null` anyways. You want me to add more code to just "make" them `null`?

@iambmelt @nazukj Based on the documentation example, they are not using final: https://static.javadoc.io/com.google.code.gson/gson/2.8.1/com/google/gson/annotations/SerializedName.html.

Also, if you finalize the parameters, I believe GSON is forced to use reflection like this:  https://stackoverflow.com/a/11220912/950427. 

I am trying to keep this pull request similar to the one I made for ADAL: https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/842

weijjia: I'll find out how GSON will do it internally and let you know.

jaredsburrows: @iambmelt @weijjia This PR solves the unsafe allocator issue: https://github.com/google/gson/blob/09cbc8073bb2bc60aeed6b831be4473daaf915ae/gson/src/main/java/com/google/gson/internal/UnsafeAllocator.java#L42.

To prevent introducing a new one, I removed the finals: https://github.com/google/gson/blob/9e44d60b83d667d96150eeff4e3f04cd4f157e9a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L157.

The `ReflectiveTypeAdapterFactory` is used in the `Gson` class here: https://github.com/google/gson/blob/92b52d25d844b81620a04917e5fa164d44b9c8a5/gson/src/main/java/com/google/gson/Gson.java#L252.

Basically I would like this to use as little of reflection as we can.

weijjia: I'll need time to investigate into it. We should finalize everything if we can and of course reduce reflections. I'll get back to the PR on Monday.

jaredsburrows:
@weijjia 
@nazukj @addev-ashish Any update on moving forward with my optimizations?

weijjia:
If I'm reading the code correctly, unSafeAllocator will be used if no default constructor is provided. So your code change about providing will solve that issue(Thanks so much for sending the PR). https://github.com/google/gson/blob/0636635cbffa08157bdbd558b1212e4d806474eb/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java#L94. 

Then we come down why we have to remove the final. I'm not sure i fully understand this part, 
https://github.com/google/gson/blob/0636635cbffa08157bdbd558b1212e4d806474eb/gson/src/main/java/com/google/gson/internal/bind/TypeAdapterRuntimeTypeWrapper.java#L46
this is how the type adapter is chosen, how does it help remove the reflection?

jaredsburrows:
@weijjia Not entirely remove it but it forces Gson to check if the "member is accessible" and since the field is `final` it has to `setAccessible(true)` here: https://github.com/google/gson/blob/9e44d60b83d667d96150eeff4e3f04cd4f157e9a/gson/src/main/java/com/google/gson/internal/bind/ReflectiveTypeAdapterFactory.java#L157.

weijjia:
Sorry if i understood, but this is setting it anyways. getBoundFields is used in the reflective adapter, whenever this adapter is used, reflection is not avoid at all.

jaredsburrows:
@weijjia Honestly, this is taking much longer than it should be. I just wanted to make sure this library was using the "no-args" constructor.

I can add the `final`s back in, I just believe that the Gson API has to use reflection to make them accessible. I just want to use reflection as little as I can.

jaredsburrows:
@weijjia I have added in the `final` for the fields. The PR should be ready to go now!
@nazukj @addev-ashish

jaredsburrows:
@weijjia Adding `this(null)` to the constructors like you asked seems to break your build:
```
com.microsoft.identity.client.InteractiveRequestTest > testAcquireTokenExpiresInNotReturned[test(AVD) - 5.0.2] FAILED 
	java.lang.RuntimeException: Failed to invoke com.microsoft.identity.client.AccessTokenCacheItem() throws com.microsoft.identity.client.MsalClientException with no args
	at com.google.gson.internal.ConstructorConstructor$3.construct(ConstructorConstructor.java:116)
Tests on test(AVD) - 5.0.2 failed: Instrumentation run failed due to 'java.lang.NullPointerException'
:msal:connectedDebugAndroidTest FAILED
```

jaredsburrows:
@weijjia Adding back in my original changes. Everything passes! Thank you.

jaredsburrows:
@weijjia Thank you for your support!

Issue #206, opened 2017-06-29
Gradle Task 'jacocoTestReport' is always SKIPPED


iambmelt:
Executing:
`$ ./gradlew jacocoTestReport`

yields:

```sh
Incremental java compilation is an incubating feature.
:msal:preBuild UP-TO-DATE
:msal:preDebugBuild UP-TO-DATE
:msal:checkDebugManifest
:msal:preDebugAndroidTestBuild UP-TO-DATE
:msal:preDebugUnitTestBuild UP-TO-DATE
:msal:preReleaseBuild UP-TO-DATE
:msal:preReleaseUnitTestBuild UP-TO-DATE
:msal:prepareComAndroidSupportAnimatedVectorDrawable2531Library UP-TO-DATE
:msal:prepareComAndroidSupportAppcompatV72531Library UP-TO-DATE
:msal:prepareComAndroidSupportCustomtabs2531Library UP-TO-DATE
:msal:prepareComAndroidSupportSupportCompat2531Library UP-TO-DATE
:msal:prepareComAndroidSupportSupportCoreUi2531Library UP-TO-DATE
:msal:prepareComAndroidSupportSupportCoreUtils2531Library UP-TO-DATE
:msal:prepareComAndroidSupportSupportFragment2531Library UP-TO-DATE
:msal:prepareComAndroidSupportSupportMediaCompat2531Library UP-TO-DATE
:msal:prepareComAndroidSupportSupportV42531Library UP-TO-DATE
:msal:prepareComAndroidSupportSupportVectorDrawable2531Library UP-TO-DATE
:msal:prepareDebugDependencies
:msal:compileDebugAidl UP-TO-DATE
:msal:compileDebugRenderscript UP-TO-DATE
:msal:generateDebugBuildConfig UP-TO-DATE
:msal:generateDebugResValues UP-TO-DATE
:msal:generateDebugResources UP-TO-DATE
:msal:mergeDebugResources UP-TO-DATE
:msal:processDebugManifest UP-TO-DATE
:msal:processDebugResources UP-TO-DATE
:msal:generateDebugSources UP-TO-DATE
:msal:incrementalDebugJavaCompilationSafeguard UP-TO-DATE
:msal:compileDebugJavaWithJavac UP-TO-DATE
:msal:incrementalDebugUnitTestJavaCompilationSafeguard UP-TO-DATE
:msal:prepareDebugUnitTestDependencies
[...]

PR #207, opened 2017-06-29
Coverage file is now found via wildcard

iambmelt:
Gradle `2.2+` generates the `.ec` coverage file per scheme `<device> - <major_ver>-coverage.ec`

This PR updates `task jacocoTestReport` to accept a wildcard `*coverage.ec` file.


iambmelt:
This commit closes #206

Issue #214, opened 2017-08-22
[Workaround :: #209] Force Travis-CI to use Ubuntu Precise


iambmelt:
To temporarily work around #209 we should consider implementing the workaround applied in [ADAL/#924](https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/924) which forces the Travis-CI build environment to use `Ubuntu Precise`

PR #217, opened 2017-08-23
Closes #216 (Makes get/set property methods `final`)

### LINKED ISSUE #216: [VSTS-265177] Constructor Invokes Overridable Function
Abstract | <Content>   <Paragraph>A constructor of Event in Event.java at line 47 calls a function that can be overridden by an attacker.<AltParagraph>A constructor of the class calls a function that can be overridden.</AltParagraph></Paragraph> </Content>
-- | --
File | msal/src/telemetry/java/com/microsoft/identity/client/Event.java
Severity | 3.0
Line | 44
Snippet | throw new IllegalStateException("Event must have a name");         }         if (!EventConstants.EventName.DEFAULT_EVENT.equals(builder.mEventName)) {             setProperty(EventProperty.EVENT_NAME, builder.mEventName);         }         if (null != builder.mEventStartTime) {             setProperty(EventProperty.START_TIME, String.valueOf(builder.mEventStartTime));
Explanation | <Content>When a constructor calls an overridable function, it may allow an attacker to access the <code>this</code> reference prior to the object being fully initialized, which can in turn lead to a vulnerability. <b>Example 1:</b> The following calls a method that can be overridden. <pre>   ...   class User {     private String username;     private boolean valid;     public User(String username, String password){       this.username = username;       this.valid = validateUser(username, password);     }     public boolean validateUser(String username, String password){       //validate user is real and can authenticate       ...     }     public final boolean isValid(){       return valid;     }   } </pre> Since the function <code>validateUser</code> and the class are not <code>final</code>, it means that they can be overridden, and then initializing a variable to the subclass that overrides this function would allow bypassing of the <code>validateUser</code> functionality. For example:  <pre>   ...   class Attacker extends User{     public Attacker(String username, String password){       super(username, password);     }     public boolean validateUser(String username, String password){       return true;     }   }   ...   class MainClass{     public static void main(String[] args){       User hacker = new Attacker("Evil", "Hacker");       if (hacker.isValid()){         System.out.println("Attack successful!");       }else{         System.out.println("Attack failed");       }     }   } </pre> The above code prints "Attack successful!", since the <code>Attacker</code> class overrides the <code>validateUser()</code> function that is called from the constructor of the superclass <code>User</code>, and Java will first look in the subclass for functions called from the constructor.</Content>
Recommendation | <Content>Constructors should not call functions that can be overridden, either by specifying them as <code>final</code>, or specifying the class as <code>final</code>. Alternatively if this code is only ever needed in the constructor, the <code>private</code> access specifier can be used, or the logic could be placed directly into the constructor of the superclass. <b>Example 2:</b> The following makes the class <code>final</code> to prevent the function from being overridden elsewhere. <pre>   ...   final class User {     private String username;     private boolean valid;     public User(String username, String password){       this.username = username;       this.valid = validateUser(username, password);     }     private boolean validateUser(String username, String password){       //validate user is real and can authenticate       ...     }     public final boolean isValid(){       return valid;     }   } </pre> This example specifies the class as <code>final</code>, so that it cannot be subclassed, and changes the <code>validateUser()</code> function to <code>private</code>, since it is not needed elsewhere in this application. This is programming defensively, since at a later date it may be decided that the <code>User</code> class needs to be subclassed, ...

iambmelt:
See #216 - This changes resolves the security issue by making constructor-invoked methods `final`

### DIFF
--- a/msal/src/telemetry/java/com/microsoft/identity/client/Event.java
+++ b/msal/src/telemetry/java/com/microsoft/identity/client/Event.java
@@ -57,13 +57,13 @@ class Event extends ArrayList<Pair<String, String>> {
         }
     }
 
-    void setProperty(final String propertyName, final String propertyValue) {
+    final void setProperty(final String propertyName, final String propertyValue) {
         if (!MsalUtils.isEmpty(propertyName) && !MsalUtils.isEmpty(propertyValue)) {
             add(new Pair<>(propertyName, propertyValue));
         }
     }
 
-    String getProperty(final String propertyName) {
+    final String getProperty(final String propertyName) {
         String propertyValue = null;
         for (final Pair<String, String> property : this) {
             if (property.first.equals(propertyName)) {


iambmelt:
Blocked by #214 #215 & VSTS-CI (agent is down)

PR #220, opened 2017-08-23
Closes #219 - Adds Logging to catch Block

### LINKED ISSUE #219: [VSTS-265175] Defect : Poor Error Handling + //TODO
Abstract | <Content>   <Paragraph>The method decodeUrlToMap() in MsalUtils.java ignores an exception on line 304, which could cause the program to overlook unexpected states and conditions.<AltParagraph>Ignoring an exception can cause the program to overlook unexpected states and conditions.</AltParagraph></Paragraph> </Content>
-- | --
File | msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
Severity | 2.0
Line | 301
Snippet | decodedUrlMap.put(key, value);                 }                 //CHECKSTYLE:OFF: checkstyle:EmptyBlock             } catch (final UnsupportedEncodingException e) {                 //CHECKSTYLE:ON: checkstyle:EmptyBlock                 // TODO: log here.             }         }         return decodedUrlMap;
Explanation | <Content>Just about every serious attack on a software system begins with the violation of a programmer's assumptions. After the attack, the programmer's assumptions seem flimsy and poorly founded, but before an attack many programmers would defend their assumptions well past the end of their lunch break. Two dubious assumptions that are easy to spot in code are "this method call can never fail" and "it doesn't matter if this call fails". When a programmer ignores an exception, they implicitly state that they are operating under one of these assumptions. <b>Example 1:</b> The following code excerpt ignores a rarely-thrown exception from <code>doExchange()</code>. <pre> try {   doExchange(); } catch (RareException e) {   // this can never happen } </pre> If a <code>RareException</code> were to ever be thrown, the program would continue to execute as though nothing unusual had occurred. The program records no evidence indicating the special situation, potentially frustrating any later attempt to explain the program's behavior.</Content>
Recommendation | <Content>At a minimum, log the fact that the exception was thrown so that it will be possible to come back later and make sense of the resulting program behavior. Better yet, abort the current operation. If the exception is being ignored because the caller cannot properly handle it but the context makes it inconvenient or impossible for the caller to declare that it throws the exception itself, consider throwing a <code>RuntimeException</code> or an <code>Error</code>, both of which are unchecked exceptions. As of JDK 1.4, <code>RuntimeException</code> has a constructor that makes it easy to wrap another exception. <b>Example 2:</b> The code in Example 1 could be rewritten in the following way: <pre> try {   doExchange(); } catch (RareException e) {   throw new RuntimeException("This can never happen", e); } </pre></Content>
Tips | There are rare types of exceptions that can be discarded in some contexts. For instance, <code>Thread.sleep()</code> throws <code>InterruptedException</code>, and in many situations the program should behave the same way whether or not it was awoken prematurely. <pre>   try {     Thread.sleep(1000);   }   catch (InterruptedException e){     // The thread has been woken up prematurely, but its     // behavior should be the same either way.   } </pre>

iambmelt:
See #219 - This change adds logging to the `catch` block in the event of url encoding problems.

### DIFF
--- a/msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
+++ b/msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
@@ -57,8 +57,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.StringTokenizer;
 
-import static android.content.ContentValues.TAG;
-
 /**
  * Internal Util class for MSAL.
  */
@@ -72,6 +70,8 @@ final class MsalUtils {
      * Default access token expiration time in seconds.
      */
     public static final int DEFAULT_EXPIRATION_TIME_SEC = 3600;
+
+    private static final String TAG = MsalUtils.class.getSimpleName();
 
     private static final String CUSTOM_TABS_SERVICE_ACTION =
             "android.support.customtabs.action.CustomTabsService";
@@ -301,7 +301,7 @@ final class MsalUtils {
                     decodedUrlMap.put(key, value);
                 }
             } catch (final UnsupportedEncodingException e) {
-                Logger.error(MsalUtils.class.getSimpleName(), null, "URL form decode failed.", e);
+                Logger.error(TAG, null, "URL form decode failed.", e);
             }
         }
 
--- a/msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
+++ b/msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
@@ -300,10 +300,8 @@ final class MsalUtils {
                 if (!MsalUtils.isEmpty(key) && !MsalUtils.isEmpty(value)) {
                     decodedUrlMap.put(key, value);
                 }
-                //CHECKSTYLE:OFF: checkstyle:EmptyBlock
             } catch (final UnsupportedEncodingException e) {
-                //CHECKSTYLE:ON: checkstyle:EmptyBlock
-                // TODO: log here.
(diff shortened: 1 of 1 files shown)


iambmelt:
@weijjia Is there any reason we may not want to log the `Exception` thrown by `urlFormDecode(String)`? Could this present a possible PII problem?

iambmelt:
Blocked by #214 #215 and VSTS-CI (build agent is down)

weijjia:
Good catch, we should log.

msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
```diff
@@ -300,10 +300,8 @@ static String getChromePackage(final Context context) {
                 if (!MsalUtils.isEmpty(key) && !MsalUtils.isEmpty(value)) {
                     decodedUrlMap.put(key, value);
                 }
-                //CHECKSTYLE:OFF: checkstyle:EmptyBlock
             } catch (final UnsupportedEncodingException e) {
-                //CHECKSTYLE:ON: checkstyle:EmptyBlock
-                // TODO: log here.
+                Logger.error(MsalUtils.class.getSimpleName(), null, "URL form decode failed.", e);
```
weijjia: Add TAG in MsalUtils class and use it instead.

iambmelt: Fixed 👍

Issue #221, opened 2017-08-23
[VSTS-265172] Defect : Byte Array To String Conversion Uses Implicit Default Locale


iambmelt:
Abstract | <Content>   <Paragraph>The call to String() on line 316 of InteractiveRequest.java converts a byte array into a <code>String</code>, which may lead to data loss.<AltParagraph>Converting a byte array into a <code>String</code> may lead to data loss.</AltParagraph></Paragraph> </Content>
-- | --
File | msal/src/request/java/com/microsoft/identity/client/InteractiveRequest.java
Severity | 2.0
Line | 313
Snippet | }         final byte[] stateBytes = Base64.decode(encodedState, Base64.NO_PADDING \| Base64.URL_SAFE);         return new String(stateBytes);     }     private void addExtraQueryParameter(final String key, final String value, final Map<String, String> requestParams) {
Explanation | <Content>When data from a byte array is converted into a <code>String</code>, it is unspecified what will happen to any data that is outside of the applicable character set. This can lead to data being lost, or a decrease in the level of security when binary data is needed to ensure proper security measures are followed. <b>Example 1:</b> The following code converts data into a String in order to create a hash. <pre>   ...   FileInputStream fis = new FileInputStream(myFile);   byte[] byteArr = byte[BUFSIZE];   ...   int count = fis.read(byteArr);   ...   String fileString = new String(byteArr);   String fileSHA256Hex = DigestUtils.sha256Hex(fileString);   // use fileSHA256Hex to validate file    ... </pre> Assuming the size of the file is less than <code>BUFSIZE</code>, this works fine as long as the information in <code>myFile</code> is encoded the same as the default character set, however if it's using a different encoding, or is a binary file, it will lose information. This in turn will cause the resulting SHA hash to be less reliable, and could mean it's far easier to cause collisions, especially if any data outside of the default character set is represented by the same value, such as a question mark.</Content>
[...]

PR #222, opened 2017-08-23
Specifies the default Charset (UTF_8) to use when decoding byte[]

iambmelt:
See #221 - This change specifies the default locale to `UTF_8` to guarantee consistent behavior across all Android devices


iambmelt:
Blocked by #214 #215 & VSTS-CI build agent (down)

PR #224, opened 2017-09-13
Update Gradle to 4.1

jaredsburrows:
- Update Android Gradle Plugin to 2.3.3
- Update Gradle to 4.1


weijjia:
Nothing I'm aware of. We should update.

Issue #239, opened 2017-12-02
HTTPRequest does not have configurable timeouts


addev-ashish:
static final int CONNECT_TIME_OUT_MSEC = 30000;
    static final int READ_TIME_OUT_MSEC = 30000;

Connect timeout and read timeout both should have a user configurable setting.

heidijinxujia:
HttpRequest is a package-visible final class.

heidijinxujia:
As we need to sync the behavior with other platforms, so this issue is currently postponed.

heidijinxujia:
This one is cut. And the network timeout will be included in Shane's configuration feature.

PR #248, opened 2018-01-26
Add API for user to configure the connect timeout and read timeout

heidijinxujia:
RFC #239 

To set the the maximum time in milliseconds to wait while connecting. Connecting to a server will fail with a SocketTimeoutException if the timeout elapses before a connection is established. Default value is 30000 milliseconds.
`com.microsoft.identity.client.MsalUtils#setConnectTimeOut`

To get the connect timeout by calling
`com.microsoft.identity.client.MsalUtils#getConnectTimeOut`

To set the maximum time to wait for an input stream read to complete before giving up. Reading will fail with a SocketTimeoutException if the timeout elapses before data becomes available. The default value is 30000.
`com.microsoft.identity.client.MsalUtils#setReadTimeOut`
To get the read timeout by calling
`com.microsoft.identity.client.MsalUtils#getReadTimeOut`


msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
```diff
@@ -96,6 +110,54 @@ static boolean isEmpty(final String message) {
...
+     * @return read timeout
+     */
+    public int getReadTimeOut() {
+        return mReadTimeOut;
+    }
+
+    /**
+     * Sets the maximum time to wait for an input stream read to complete before
+     * giving up. Reading will fail with a SocketTimeoutException if the timeout
+     * elapses before data becomes available. The default value is 30000.
```
iambmelt: Regarding long-term ease of maintenance, I would consider making `DEFAULT_READ_TIMEOUT` `public final` and amending the last sentence of this javadoc to say "The default value is `DEFAULT_READ_TIMEOUT`".

I would consider the same approach for `DEFAULT_CONNECT_TIMEOUT`.

This way, if the value changes in the future, it needs only to be updated in a single spot and the documentation is always correct.

iambmelt: If you do plan to make `DEFAULT_READ_TIMEOUT` and `DEFAULT_CONNECT_TIMEOUT` `public`, I would also suggest adding inline javadoc for those fields.

msal/src/internal/java/com/microsoft/identity/client/MsalUtils.java
```diff
@@ -58,18 +58,21 @@
 import java.util.StringTokenizer;
 
 /**
- * Internal Util class for MSAL.
+ * Util class for MSAL.
  */
-final class MsalUtils {
+public final class MsalUtils {
```
iambmelt: Unless you're opposed to it, I think I'd prefer to see a single-purpose object for end-user tweaking of settings akin to ADAL's `AuthenticationSettings` (although, I'm admittedly not crazy about the naming of this class).

As `MsalUtils` is already a rather confusing mixture of odds 'n ends methods, I think we can reasonably create a new class which holds default values and accepts user overrides of them, possibly with inner-classes to delineate settings thematically.

For instance:
`MsalSettings.Network.CONNECT_TIMEOUT_MILLIS = 15000L;`

and maybe one day....
`MsalSettings.Logging.androidLoggingEnabled= true;`

or....
`MsalSettings.Telemetry.allowPii = true`

As a concept, I'd like to see us eventually corral our settings and flags into a unified place. I'll open an issue for discussion.

iambmelt: Issue opened [here](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/249).

heidijinxujia: Agree. I have the same thought.

heidijinxujia:
@iambmelt Thanks a lot for the review. Comments addressed. Please have a look.

msal/src/internal/java/com/microsoft/identity/client/MSALSettings.java
```diff
@@ -0,0 +1,95 @@
...
+public enum MsalSettings {
+    /**
+     * Singleton setting instance.
+     */
+    INSTANCE;
+
+    /**
+     * Default value of read timeout in milliseconds.
+     */
+    public static final int DEFAULT_READ_TIMEOUT = 30000;
```
addev-ashish: can be private

msal/src/internal/java/com/microsoft/identity/client/MSALSettings.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Default value of connect timeout in milliseconds.
+     */
+    public static final int DEFAULT_CONNECT_TIMEOUT = 30000;
```
addev-ashish: can be private

msal/src/internal/java/com/microsoft/identity/client/MSALSettings.java
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Sets the maximum time in milliseconds to wait while connecting.
+     * Connecting to a server will fail with a SocketTimeoutException if the
+     * timeout elapses before a connection is established. Default value is
+     * {@value DEFAULT_CONNECT_TIMEOUT} milliseconds.
+     *
+     * @param timeOutMillis the non-negative connect timeout in milliseconds.
+     */
+    public void setConnectTimeOut(int timeOutMillis) {
```
addev-ashish: final int

msal/src/internal/java/com/microsoft/identity/client/MSALSettings.java
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Sets the maximum time to wait for an input stream read to complete before
+     * giving up. Reading will fail with a SocketTimeoutException if the timeout
+     * elapses before data becomes available. The default value is {@value DEFAULT_READ_TIMEOUT}
+     * milliseconds.
+     *
+     * @param timeOutMillis the non-negative read timeout in milliseconds.
+     */
+    public void setReadTimeOut(int timeOutMillis) {
```
addev-ashish: final

Issue #251, opened 2018-01-31
Silent Auth is broken in the test app


heidijinxujia:
Error details
Unable to find the user when making the silent auth request.

nazukj:
@heidijinxujia is it ok to close this issue?

PR #252, opened 2018-01-31
Fix Silent Auth in Test App | issue#251

heidijinxujia:
RFC #251 
Fix the silent auth in the test app as well as other bugs encountered during the build.


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
```diff
@@ -233,26 +233,47 @@ public void onAcquireTokenClicked(final AcquireTokenFragment.RequestOptions requ
...
+            for (final User user : users) {
+                if (user.getDisplayableId().equals(loginHint.trim().toLowerCase())) {
+                    return user;
+                }
+            }
+        } catch (final MsalClientException e) {
+            Log.e(TAG, "Fail to retrieve users: " + e.getMessage(), e);
+        }
+
+        showMessage("Users are removed.");
```
iambmelt: Can we call `showMessage()` conditionally only if the `User` was _actually_ removed? Maybe consider showing an error message if the Exception is thrown or if the `login_hint` can't be found?

Issue #257, opened 2018-02-06
ActivityNotFoundException for devices with Chrome app disabled


imablanco:
For devices with Chrome app disabled but not uninstalled, PackageInfo still returns for Chrome package so library is throwing an unhandled ActivityNotFoundException when trying to start authentication flow.
Reproduced with a Nexus 5.

tipa:
When is this fix expected to show up as a part of an update to the `Microsoft.Identity.Client`-NuGet package to work in Xamarin.Android projects?

danieldobalian:
@imablanco thanks for calling this out.  In the case Chrome is disabled, do you expect the library to fall back on another browser or give the app the option fallback?  

@tipa are you using MSAL .NET and seeing the same exception? Or are you pulling in MSAL Android into your Xamarin project?

tipa:
@danieldobalian I am using this Nugetpackage in my Xamarin.Android project: https://www.nuget.org/packages/Microsoft.Identity.Client/ (version 1.1.2-preview0008).
Using this package results in the exception when Chrome is disabled/not installed. Does this answer your question?

imablanco:
@danieldobalian The ideal scenario is falling back on whatever browser the user has installed.
I made a PR for this issue ( #258 ), just preventing the runtime crash.

danieldobalian:
@tipa yes, thank you. That's MSAL .NET (a separate code base). I'll let that team know of this issue. 

@imablanco Agreed, that's the direction I was thinking as well, but wanted to check.  Thanks for the PR, we'll take a look and see if we can adopt it for the next release.

danieldobalian:
@jmprieur FYI on MSAL .NET having the same runtime crash when Chrome is disabled.

tipa:
This issue is getting old. Any news? Having tons of frustrated users because of this bug

tipa:
How is this issue still not fixed...? :(
The whole OneDrive login process worked flawflessly in the previous SDK, this is a big step back!

tipa:
Any updates?

danieldobalian:
This has been fixed in MSAL 0.2.0 with the introduction of support for all browsers. When Chrome is disabled, we'll fallback to the next browser available on the device.

PR #258, opened 2018-02-06
Check Chrome package enabled status when checking for Chrome support

imablanco:
Fixed issue #257


heidijinxujia:
@danieldobalian This PR seems fallback to whatever the device has when the Chrome is disable. But our code base haven't support other browser besides Chrome. We could fallback to embedded webview flow in this case once we finished the embedded webview flow.

Issue #278, opened 2018-04-30
Fix the broken unit tests due to Common-core Logger Migration


heidijinxujia:
![image](https://user-images.githubusercontent.com/16140683/39455764-1db90eac-4c97-11e8-9fc7-5910d11849b8.png)

PR #282, opened 2018-05-01
Fix the broken unit tests due to common-core logging migration

heidijinxujia:
1. Remove the Ignore annotation in the broker unit tests in LoggerTest
2. Fix the broker unit tests in #278 

Part of #277

### DIFF
--- a/msal/build.gradle
+++ b/msal/build.gradle
@@ -105,15 +105,24 @@ dependencies {
     implementation "com.google.code.gson:gson:$rootProject.ext.gsonVersion"
     // test dependencies
     testImplementation "junit:junit:$rootProject.ext.junitVersion"
-    testImplementation "org.mockito:mockito-core:$rootProject.ext.mockitoCoreVersion"
+    // TODO the below dependency should be mockito-core but the MockMaker isn't working...
+    testImplementation "org.mockito:mockito-android:$rootProject.ext.mockitoAndroidVersion"
     // instrumentation test dependencies
     androidTestImplementation "com.android.support.test:runner:$rootProject.ext.runnerVersion"
     // Set this dependency to use JUnit 4 rules
     androidTestImplementation "com.android.support.test:rules:$rootProject.ext.rulesVersion"
-    androidTestImplementation "org.mockito:mockito-core:$rootProject.ext.mockitoCoreVersion"
-    androidTestImplementation "com.google.dexmaker:dexmaker:$rootProject.ext.dexmakerMockitoVersion"
-    androidTestImplementation "com.google.dexmaker:dexmaker-mockito:$rootProject.ext.dexmakerMockitoVersion"
-    implementation project(':common')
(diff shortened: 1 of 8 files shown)


msal/src/androidTest/java/com/microsoft/identity/client/LoggerTest.java
```diff
@@ -285,11 +282,15 @@ private void verifyLogMessage(final RequestContext requestContext, final LogResp
 
     private void verifyLogMessageFormat(final LogResponse response, final String message, final RequestContext requestContext,
                                         final Throwable throwable) {
-        Assert.assertTrue(response.getMessage().contains("MSAL " + PublicClientApplication.getSdkVersion() + " Android " + Build.VERSION.SDK_INT + " ["));
+        Assert.assertTrue(response.getMessage().contains(" SDK ver:" + VERSION_NAME + " Android " + Build.VERSION.SDK_INT));
         if (requestContext != null && (requestContext.getCorrelationId() != null || !MsalUtils.isEmpty(requestContext.getComponent()))) {
             if (requestContext.getCorrelationId() != null && !MsalUtils.isEmpty(requestContext.getComponent())) {
-                Assert.assertTrue(response.getMessage().contains(" - " + requestContext.getCorrelationId().toString() + "] (" + requestContext.getComponent()
-                        + ") " + message + getStackTrace(throwable)));
+                final String expectedStr = " - " + requestContext.getCorrelationId().toString()
```
iambmelt: I'm expecting this test to break when I merge in my latest logging changes. I'll take care of modifying this test when the time comes

msal/build.gradle
```diff
@@ -105,15 +105,24 @@ dependencies {
     implementation "com.google.code.gson:gson:$rootProject.ext.gsonVersion"
     // test dependencies
     testImplementation "junit:junit:$rootProject.ext.junitVersion"
-    testImplementation "org.mockito:mockito-core:$rootProject.ext.mockitoCoreVersion"
+    // TODO the below dependency should be mockito-core but the MockMaker isn't working...
+    testImplementation "org.mockito:mockito-android:$rootProject.ext.mockitoAndroidVersion"
```
iambmelt: `mockito-core` is throwing weird exceptions on my test device. I've changed this dependency to use `mockito-android` though it returns its own strange warnings (tests pass, however). This is something we should revisit, but we're unblocked for now.

```
IMPORTANT NOTE FROM MOCKITO:

You included the 'mockito-android' dependency in a non-Android environment.
The Android mock maker was disabled. You should only include the latter in your 'androidTestCompile' configuration
If disabling was a mistake, you can set the 'org.mockito.mock.android' property to 'true' to override this detection.

Visit https://javadoc.io/page/org.mockito/mockito-core/latest/org/mockito/Mockito.html#0.1 for more information

Process finished with exit code 0
```

iambmelt: I'll open an issue for this and link back here.

iambmelt: See #284

PR #283, opened 2018-05-01
Fixed unit tests.

iambmelt:
Note: this commit uses mockito-android for plain-java tests. The Mockito.MockMaker in mockito-core is throwing odd Exceptions.


heidijinxujia:
P.S. I thought it would commit into the unified/dev. 😅

PR #285, opened 2018-05-02
Cleaner transitive dependency exclusion

iambmelt:
See description. Rather than use explicit `group` exclusion clauses, this now uses a bulk `transitive = false` flag


iambmelt:
Closing. Will make fixes on the other branch

PR #300, opened 2018-07-17
Add integration for common embedded webview

heidijinxujia:
* Fix the build and update the common submodule
* Cherry pick Krishna's refactor on customTab.
* Add resource files.
* Update the unit tests and removed ones no longer used.


msal/src/main/java/com/microsoft/identity/client/AuthenticationActivity.java
```diff
@@ -288,4 +244,42 @@ private void sendError(final String errorCode, final String errorDescription) {
...
+
+    private MicrosoftStsAuthorizationRequest getAuthorizationRequestFromIntent(final Intent callingIntent) {
+        MicrosoftStsAuthorizationRequest authRequest = null;
+        Serializable request = callingIntent
+                .getSerializableExtra(AuthenticationConstants.Browser.REQUEST_MESSAGE);
+
+        if (request instanceof MicrosoftStsAuthorizationRequest) {
+            Logger.verbose(TAG, null, "Finish generating the authorization request.");
+            authRequest = (MicrosoftStsAuthorizationRequest) request;
+
```
piqiums: unnecessary empty line

build.gradle
```diff
@@ -15,6 +15,7 @@ buildscript {
 allprojects {
     repositories {
         jcenter()
+        google()
```
addev-ashish: mavenLocal

msal/src/internal/java/com/microsoft/identity/client/AuthenticationRequestParameters.java
```diff
@@ -70,6 +71,12 @@ private AuthenticationRequestParameters(final Authority authority, final TokenCa
         mClientId = clientId;
         mSliceParameters = sliceParameters;
         mRequestContext = requestContext;
+
+        if (authority.getAuthorityType() == Authority.AuthorityType.B2C) {
+            mWebViewSelection = WebViewSelection.SYSTEM_BROWSER;
+        } else {
+            mWebViewSelection = WebViewSelection.EMBEDDED_WEBVIEW;
```
addev-ashish: nit: initialize the variable with EMBEDDED_WEBVIEW and then you don't need the else block

msal/src/main/java/com/microsoft/identity/client/AuthenticationActivity.java
```diff
@@ -100,99 +115,39 @@ protected void onCreate(final Bundle savedInstanceState) {
...
+        }
+    }
+
     @Override
     protected void onStart() {
         super.onStart();
-        if (mChromePackageWithCustomTabSupport != null) {
-            warmUpCustomTabs();
+        if (!mUseEmbeddedWebView) {
+            mChromeCustomTabManager.bindCustomTabsService();
```
addev-ashish: can we revisit this pattern of if (!mUseEmbeddedWebView), as the bindCustomTabsService has its own checks, so we can potentially get into it and then have no way to fallback or recover

heidijinxujia: Addressed. Please have a look. @addev-ashish

msal/src/main/java/com/microsoft/identity/client/AuthenticationActivity.java
```diff
@@ -288,4 +240,42 @@ private void sendError(final String errorCode, final String errorDescription) {
         errorIntent.putExtra(Constants.UIResponse.ERROR_DESCRIPTION, errorDescription);
         returnToCaller(Constants.UIResponse.AUTH_CODE_ERROR, errorIntent);
     }
-}
+
+    private MicrosoftStsAuthorizationRequest getAuthorizationRequestFromIntent(final Intent callingIntent) {
```
iambmelt: Do we have to use this specific return type? It looks like we _may_ be able to get away with casting-to and returning the supertype `AuthorizationRequest` -- you'll have to make the generic type params of `mEmbeddedWebViewAuthorizationStrategy` use the supertype too. Ideally, we keep this `Activity`'s implementation as non-Microsoft specific as possible

It looks like all we do is call `requestAuthorization()`, right?

msal/src/request/java/com/microsoft/identity/client/InteractiveRequest.java
```diff
@@ -51,10 +58,9 @@
 
     static final int BROWSER_FLOW = 1001;
     private static AuthorizationResult sAuthorizationResult;
+    private static MicrosoftStsAuthorizationRequest sAuthorizationRequest;
```
iambmelt: This assume all interactive requests will be made to the v2 endpoint? I suspect we don't want to couple these classes in this way

msal/src/request/java/com/microsoft/identity/client/InteractiveRequest.java
```diff
@@ -127,6 +140,73 @@ synchronized void preTokenRequest() throws MsalUserCancelException, MsalClientEx
         processAuthorizationResult(sAuthorizationResult);
     }
 
+    /**
+     * Create the {@Link MicrosoftStsAuthorizationRequest} from authentication parameters.
+     *
+     * @return MicrosoftStsAuthorizationRequest
+     * @throws MsalClientException
+     */
+    private MicrosoftStsAuthorizationRequest createAuthRequest() throws MsalClientException {
```
iambmelt: This is coupled awfully tight. Why not have `InteractiveRequest` accept an `AuthorizationRequest` (superclass) as a parameter?

Inside `InteractiveRequest`, `createAuthRequest()` is behaving like a Factory...

msal/src/request/java/com/microsoft/identity/client/InteractiveRequest.java
```diff
@@ -127,6 +140,73 @@ synchronized void preTokenRequest() throws MsalUserCancelException, MsalClientEx
...
+                PublicClientApplication.getSdkVersion(),
+                promptBehavior,
+                null,
+                null,
+                null,
+                getAuthRequestParameters().getSliceParameters(),
+                null);
+
+        if (null != getAuthRequestParameters().getUser()) {
+            authorizationRequest.setUid(getAuthRequestParameters().getUser().getUid());
```
iambmelt: I'd prefer if this method weren't here and if this class didn't know what type of `AuthorizationRequest` it is handling BUT I have a minor nit about this code block too: we derefence `getAuthRequestParameters().getUser()` 3 times.

For wherever this method ends-up, can we get the `User` once inside this scope and then call the getters on it?
```java
final User user = getAuthRequestParameters().getUser();
authorizationRequest.setUid(user.getUid());
authorizationRequest.setUtid(user.getUtid());
authorizationRequest.setDisplayableId(user.getDisplayableId());
```

msal/src/request/java/com/microsoft/identity/client/InteractiveRequest.java
heidijinxujia: Let's talk about this offline today.

msal/src/main/java/com/microsoft/identity/client/AuthenticationActivity.java
heidijinxujia: Thank you for the comments Brian! I got some confusions on this as well as the implementation of CoreAdapter. Please let me know when you have available to discuss about this offline.

heidijinxujia:
@addev-ashish @iambmelt  Please have a look, thanks! As to the refactor of EmbeddedWebViewAuthorizationStrategy, I am going to open a separate PR for it like we discussed last week. And I am going to start it after Shane pushes his work.

PR #304, opened 2018-07-18
Update MSAL Target Version to 27

heidijinxujia:
* Updated MSAL Target Version to 27
* Updated TravisCI config
* Fixed the Jacoco config

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,7 @@ buildscript {
 allprojects {
     repositories {
         jcenter()
+        google()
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -15,6 +15,7 @@ buildscript {
 allprojects {
     repositories {
         jcenter()
+        google()
```
addev-ashish: can we add mavenLocal as the first entry?

PR #322, opened 2018-08-13
[WIP] Implements getAccounts + object model

iambmelt:


msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -0,0 +1,80 @@
...
+     * Sets the home account id.
+     *
+     * @param homeAccountId The IAccountId to set.
+     */
+    void setHomeAccountId(final IAccountId homeAccountId) {
+        mHomeAccountId = homeAccountId;
+    }
+
+    @Override
+    public IAccountId getHomeAccountId() {
```
shoatman: Let's change Id to Identifier

PR #326, opened 2018-08-21
Integrate the AuthorizationStrategy 

heidijinxujia:


msal/src/main/java/com/microsoft/identity/client/AuthenticationActivity.java
```diff
@@ -24,22 +24,18 @@
 package com.microsoft.identity.client;
 
 import android.app.Activity;
+import android.content.ComponentName;
```
heidijinxujia: AuthenticationActivity is deprecated. Here, I just revert the change according to the master branch.

heidijinxujia:
TODO: fix the unit tests.

heidijinxujia:
Had an offline discussion with Shane. 
1. Ignore the InteractiveRequestTest and move the interactive request unit tests into MSALAPIControllerTest
2. Deprecate AuthenticationActivity (need to follow up with Jean-Marc)
3. Ignore AuthenticationActivityTest
4. Add AuthorizationActivityTest into common repo

PR #332, opened 2018-08-27
Add public client configuration object, authorities and audiences

shoatman:


msal/src/androidTest/java/com/microsoft/identity/client/DiscoveryTests.java
```diff
@@ -0,0 +1,27 @@
...
+import org.junit.runner.RunWith;
+
+import java.io.IOException;
+
+@RunWith(AndroidJUnit4.class)
+public class DiscoveryTests {
+
+
+    @Test
+    public void testInstanceDiscovery() throws IOException {
```
iambmelt: Additional tests we need to add here? Seems to only check that an Exception was not thrown

msal/src/androidTest/java/com/microsoft/identity/client/PublicClientConfigurationTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(AndroidJUnit4.class)
+public class PublicClientConfigurationTest {
+
+    /**
+     * Verifies that a minimum configuration containing only:
+     * client_id
+     * redirect_uri
```
iambmelt: Are we checking only that these values were not null? Do we need to check a specific schema? (GUID, prefix, anything more than != null?)

msal/src/main/java/com/microsoft/identity/client/KnownAuthorities.java
```diff
@@ -0,0 +1,34 @@
+package com.microsoft.identity.client;
```
iambmelt: Need to license info (can follow in subsequent PR)

msal/src/main/res/raw/msal_default_config.json
```diff
@@ -0,0 +1,16 @@
...
+  "authorities" : [
+    {
+      "type" : "AAD",
+      "audience": {
+        "type" : "AzureADandPersonalMicrosoftAccount"
+      },
+      "default": true
+    }
+  ],
+  "authorization_user_agent": "DEFAULT", // DEVICE_BROWSER, EMBEDDED_USERAGENT, DEVICECODEFLOW, DEFAULT
```
iambmelt: We'll want to remove the comment in this JSON file prior to merging -- while not an ultra scientific investigation, the consensus on SO seems to be that [JSON does not support comments](https://stackoverflow.com/a/244858/741827)

msal/src/main/res/values/public.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
iambmelt: This looks like it can be removed

msal/src/request/java/com/microsoft/identity/client/InteractiveRequest.java
```diff
@@ -152,6 +152,27 @@ private MicrosoftStsAuthorizationRequest createAuthRequest() throws MsalClientEx
                 break;
         }
 
+        /*
```
iambmelt: Can we delete this?

msal/src/androidTest/java/com/microsoft/identity/client/PublicClientConfigurationTest.java
shoatman: Not null is sufficient.  Just verifying that we have a value is sufficient for now.  Their may be more specific validation required for redirect_uri at some point.

msal/src/androidTest/java/com/microsoft/identity/client/DiscoveryTests.java
shoatman: Yes.  Let's discuss.

msal/src/main/java/com/microsoft/identity/client/KnownAuthorities.java
shoatman: Missed this one I guess.  Sounds good.

msal/src/main/res/raw/msal_default_config.json
shoatman: Sure.

msal/src/main/res/values/public.xml
shoatman: This actually has purpose.  By declaring something public... you make everything not declared public private.  With that said... I have not been able to verify the documented behavior yet... since i've been loading the library locally and not as an package.

msal/src/request/java/com/microsoft/identity/client/InteractiveRequest.java
shoatman: Done.

msal/src/main/java/com/microsoft/identity/client/KnownAuthorities.java
shoatman: Done.

iambmelt:
Build is green but required update from target branch. Pulling in those commits now...

PR #335, opened 2018-08-29
Tests for PublicClientApplicationConfig

iambmelt:
Tests:
- Test that the default config loads successfully
- Tests that Http settings can be configured for timeouts.
- Test that the min configuration loads (`client_id` and `redirect_uri`)
- Verifies that a minimum configuration containing only `client_id` and `redirect_uri` Is merged with the default configuration correctly
- Verify a `B2CAuthority` can be set via config
- Verify an `ADFSAuthority` can be set via config
- Verify an `AudienceType` of `AccountsInOneOrganization` can be set
- Verify an `AudienceType` of `AnyOrganizationAccount` can be set
- Verify an `AudienceType` of `AllAccounts` can be set
- Verify an `AudienceType` of `AnyPersonalAccount` can be set
- Test that `slice` parameters can be set
- Tests that `flight` parameters can be set
- Verify that `UnknownAuthority` results in `IllegalArgumentException` when validated
- Verify that `UnknownAudience` results in `IllegalArgumentException` when validated
- Verify that a `null` `client_id` throws `IllegalArgumentException` when validated
- Verify that a `null` `redirect_uri` throws `IllegalArgumentException` when validated


msal/src/androidTest/res/raw/test_pcaconfig_with_slice.json
```diff
@@ -0,0 +1,21 @@
...
+  "client_id": "[clientid]",
+  "redirect_uri": "[redirectUri]",
+  "authorities": [
+    {
+      "type": "AAD",
+      "audience": {
+        "type": "AzureADMyOrg",
+        "tenant_id": "[tenantid]"
+      },
+      "slice": {
```
iambmelt: @shoatman This is my proposed spec change. `slice` is now an object (map) instead of array of objects

shoatman: Looks good.

PR #350, opened 2018-09-07
AcquireTokenSilent implemented w/ Command, Strategy Pattern

iambmelt:
Work in prog


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -781,12 +768,71 @@ public void acquireTokenSilentAsync(@NonNull final String[] scopes, @NonNull fin
...
+            // Do I use the authority of the Account?
+            requestAuthority = "";
+        }
+
+        final MSALAcquireTokenSilentOperationParameters params = getSilentOperationParameters(
+                scopes,
+                requestAuthority
+        );
+
+        final MSALSilentTokenCommand silentTokenCommand = new MSALSilentTokenCommand(
```
iambmelt: I suspect that this subclass is not needed, as `MSALTokenCommand` looks to basically cover the silent case

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -781,12 +768,71 @@ public void acquireTokenSilentAsync(@NonNull final String[] scopes, @NonNull fin
...
+     */
+    public void acquireTokenSilentAsync(@NonNull final String[] scopes,
+                                        @NonNull final IAccount account,
+                                        @Nullable final String authority,
+                                        final boolean forceRefresh,
+                                        @NonNull final AuthenticationCallback callback) {
+        // TODO add support for forceRefresh
+        String requestAuthority = authority;
+
+        if (StringUtil.isEmpty(requestAuthority)) {
```
iambmelt: Not sure which values to fall back on here since the `PublicClientApplicationConfiguration` can return multiple entries

shoatman: We'll need to look for the default one.... Not sure we've implemented that yet.  I'll take a look.

msal/src/main/java/com/microsoft/identity/client/authorities/Authority.java
```diff
@@ -123,4 +126,9 @@ protected boolean getKnownToDeveloper() {
         return mKnownToDeveloper;
     }
 
+    public static String getAuthorityFromAccount(final IAccount account) {
+        return "https://login.microsoftonline.com/"
```
shoatman: This looks like a design problem that we need to resolve.  Don't we need to know which cloud the account is associated with?  Here we're using world wide as default?

msal/src/main/java/com/microsoft/identity/client/controllers/LocalMSALController.java
```diff
@@ -134,26 +140,138 @@ private TokenResult performTokenRequest(OAuth2Strategy strategy, AuthorizationRe
...
+                        tokenResult.getTokenResponse()
+                );
+
+                // Create a new AuthenticationResult to hold the saved record
+                final AuthenticationResult authenticationResult = new AuthenticationResult(savedRecord);
+
+                // Set the AuthenticationResult on the final result object
+                acquireTokenSilentResult.setAuthenticationResult(authenticationResult);
+            } else {
+                // TODO Is this the correct Exception?
```
shoatman: I believe that's the correct exception.

msal/src/main/java/com/microsoft/identity/client/controllers/LocalMSALController.java
```diff
@@ -134,26 +140,138 @@ private TokenResult performTokenRequest(OAuth2Strategy strategy, AuthorizationRe
...
+        final String[] scopes = scopesStr.split("\\s");
+        final List<String> atScopes = Arrays.asList(scopes);
+        return atScopes.containsAll(requestParameters);
+    }
+
+    private TokenResult performSilentTokenRequest(final OAuth2Strategy strategy,
+                                                  final MSALAcquireTokenSilentOperationParameters parameters) throws MsalClientException, IOException {
+        throwIfNetworkNotAvailable(parameters.getAppContext());
+
+        final TokenRequest request = new MicrosoftStsTokenRequest();
```
shoatman: Concrete type.  We likely should add another createTokenRequest method on the strategy.  There's a design problem here to work on .... relative to the type of token request... but I think we can defer that for now, by adding a createRefreshTokenRequest abstract method to oauth2strategy

msal/src/main/java/com/microsoft/identity/client/controllers/LocalMSALController.java
```diff
@@ -134,26 +140,138 @@ private TokenResult performTokenRequest(OAuth2Strategy strategy, AuthorizationRe
...
+                        .getIdentifier();
+
+        final Account targetAccount = tokenCache.getAccount(
+                environment,
+                clientId,
+                homeAccountId
+        );
+
+        final OAuth2Strategy strategy = parameters.getAuthority().createOAuth2Strategy();
+        final ICacheRecord cacheRecord = tokenCache.load(clientId, targetAccount);
```
shoatman: I think i'm missing something here.  Where did you say which access token that you're looking for?  The account could have multiple access tokens?

PR #352, opened 2018-09-07
Updates to utilize all public api parameters in authorization request

shoatman:


msal/src/main/java/com/microsoft/identity/client/UiBehavior.java
```diff
@@ -36,11 +36,27 @@
...
+    public String toString(){
+        switch(this){
+            case SELECT_ACCOUNT:
+                return SELECT_ACCOUNT.name().toLowerCase();
+            case FORCE_LOGIN:
+                return "login";
+            case CONSENT:
+                return CONSENT.name().toLowerCase();
+            default:
+                throw new IllegalArgumentException();
```
iambmelt: `IllegalStateException` may be a closer representation of what went wrong here, but the outcome is the same (aka `RuntimeException`)

PR #366, opened 2018-09-20
Remove the dependency on AuthorizationConfiguration

heidijinxujia:
Add support for AuthorizationAgent in LocalMSALController
Remove the dependency on AuthorizationConfiguration
Update the test app config file.


heidijinxujia:
Thanks @iambmelt resolved.

iambmelt:
LGTM

PR #373, opened 2018-09-24
Removes Telemetry from public API, corrects test package declaration

iambmelt:


msal/versioning/version.properties
```diff
@@ -1,3 +1,3 @@
 #Wed Aug 01 15:24:11 PDT 2018
-versionName=0.1.3
+versionName=0.2.0-RC1
```
addev-ashish: why is the version update a part of this change?

PR #375, opened 2018-09-24
Fixed bug and added unit tests

shoatman:


msal/src/main/java/com/microsoft/identity/client/internal/authorities/AnyOrganizationalAccount.java
```diff
@@ -28,12 +28,13 @@
 
     public AnyOrganizationalAccount() {
         this.setCloudUrl(AzureActiveDirectory.DEFAULT_CLOUD_URL);
-        this.setTenantId("organizations");
+        this.setTenantId(AzureActiveDirectoryAudience.ORGANIZATIONS);
```
kreedula: Nit : can be `this(AzureActiveDirectory.DEFAULT_CLOUD_URL)` to avoid code duplication

shoatman: The tenant id and the cloudUrl are separate values.  Where the cloud url is the authority part and the tenant id is the first path segment.  So these are separate values.

PR #384, opened 2018-09-27
Fix the bug for Huawei browser and add webview fallback

heidijinxujia:
1. Fix the bug for Huawei browser.
2. Add validation for browser and fallback to embedded webview if no browser found.


msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
```diff
@@ -116,6 +125,17 @@ private AuthorizationResult performAuthorizationRequest(final OAuth2Strategy str
         return result;
     }
 
+    private AuthorizationAgent validAuthorizationAgent(final AuthorizationAgent agent, final Context context) {
```
shoatman: Looks like it's possible to move this logic inside the factory.  Feels like it's a responsibility of the factory to return an authorization strategy that will work on the current device.

heidijinxujia:
@shoatman Thank you for the comment. I've moved it into authorization factory in common.

PR #387, opened 2018-09-28
Adds support for MRRTs

iambmelt:
See corollary pr here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/250

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/ExceptionAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/ExceptionAdapter.java
@@ -40,21 +40,24 @@ public class ExceptionAdapter {
 
     public static MsalException exceptionFromAcquireTokenResult(final AcquireTokenResult result) {
         final AuthorizationResult authorizationResult = result.getAuthorizationResult();
-        final AuthorizationErrorResponse authorizationErrorResponse = authorizationResult.getAuthorizationErrorResponse();
-        final TokenResult tokenResult = result.getTokenResult();
-        final TokenErrorResponse tokenErrorResponse;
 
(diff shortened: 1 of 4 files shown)


iambmelt:
@shoatman - Some additional error handling is needed inside of `LocalMSALController` to catch errors returned by the STS on refresh requests. I'll introduce this tomorrow

msal/src/main/java/com/microsoft/identity/client/internal/controllers/ExceptionAdapter.java
```diff
@@ -40,22 +40,25 @@
...
-                case USER_CANCEL:
-                    return new MsalUserCancelException();
 
+        if (null != authorizationResult) {
+            final AuthorizationErrorResponse authorizationErrorResponse = authorizationResult.getAuthorizationErrorResponse();
+            if (!authorizationResult.getSuccess()) {
+                //THERE ARE CURRENTLY NO USAGES of INVALID_REQUEST
+                switch (result.getAuthorizationResult().getAuthorizationStatus()) {
+                    case FAIL:
+                        return new MsalServiceException(authorizationErrorResponse.getError(), authorizationErrorResponse.getError() + ";"
```
kreedula: Should we do a null check here? 
Can `authorizationErrorResponse` be null in case of `FAIL` ?

iambmelt: I looked around (there's a few usages) and nothing stuck-out to me.... if there's any issues, hopefully we can suss them out in our bug bash

capture

kreedula: Sounds good. Thanks @iambmelt

PR #388, opened 2018-09-30
Delete the RT when invalid_grant is received and the requested scopes match the cached RT target

iambmelt:
If we attempt to renew an AT using the RT and `invalid_grant` is returned, then we inspect the contents of the cached RT and, if the scopes match the request, it is deleted (we can assume it's expired).

In the case where `invalid_grant` is returned and the cached RT scopes do NOT match the request, the RT is preserved -- this allows us to receive `invalid_grant` for admin-consent scenarios and not delete what is otherwise a valid RT until a new one is attained in a subsequent acquireToken call.

Other changes in this PR:
- Make `IAccount` and `IAccountIdentifier` interfaces implement `Serializable`
    * This supports apps who need to save these values between restarts, whatever
- `LocalMsalController` will now throw an `Exception` if the `IAccount` supplied to `acquireTokenSilent` doesn't exist in the cache
    * If an app saves an `IAccount` and calls `removeAccount()` and then passes that `IAccount` to `acquireTokenSilent()` this may happen

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/IAccount.java
+++ b/msal/src/main/java/com/microsoft/identity/client/IAccount.java
@@ -22,10 +22,12 @@
 //  THE SOFTWARE.
 package com.microsoft.identity.client;
 
+import java.io.Serializable;
+
 /**
  * Interface describing MSAL's externally-exposed Account representation.
  */
-public interface IAccount {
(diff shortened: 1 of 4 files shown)


iambmelt:
@kreedula Putting this PR on your radar to go out in next RC build

PR #392, opened 2018-10-02
Plumbed correlation id for authorization and token requests

shoatman:


iambmelt:
@shoatman I'll file an issue to move the `correlation_id` out of `AuthenticationConstants` and into `DiagnosticContext`

PR #406, opened 2018-10-04
[B2C] Fix for B2C Silent Auth 

heidijinxujia:


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -784,7 +784,16 @@ public void acquireTokenSilentAsync(@NonNull final String[] scopes,
         String requestAuthority = authority;
 
         if (StringUtil.isEmpty(requestAuthority)) {
-            requestAuthority = Authority.getAuthorityFromAccount(account);
+            if (!StringUtil.isEmpty(mAuthorityString)
```
iambmelt: This looks to be copy/pasted out of `Authority` -- why can't we modify the behavior of that method to return `null` if the `tid` is blank? Line `797` would then set this value correctly

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
```diff
@@ -226,6 +228,8 @@ IAccount getAccount(final String loginHint) {
         for (final IAccount account : mApplication.getAccounts()) {
             if (account.getUsername().equals(loginHint.trim().toLowerCase())) {
                 return account;
+            } else if (account.getAccountIdentifier().getIdentifier().equals(loginHint.trim().toLowerCase())) {
```
iambmelt: I don't think this _should_ work: this value is either going to be a GUID or an pair-wise ID: I don't know under what circumstances this could reliably match a login_hint but this seems flawed 🤔

heidijinxujia: The oid is returned as string here account.getAccountIdentifier().getIdentifier().

heidijinxujia: In IAccount.java

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
heidijinxujia: Addressed.

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
heidijinxujia: The oid is returned as a string in the code `account.getAccountIdentifier().getIdentifier()`

gradle/versions.gradle
```diff
@@ -1,7 +1,7 @@
 // Variables for entire project
 ext {
     // SDK
-    minSdkVersion = 16
+    minSdkVersion = 21
```
iambmelt: 👍

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
heidijinxujia: I will remove this part for now. 'cause I found some other bug for testapp. Let's just leave it as it is for now.

PR #409, opened 2018-10-05
Fix the the B2C silent request

heidijinxujia:
Fix the segment for b2c authority. Fix the construction for the b2c authority part of the B2C token endpoint.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -785,7 +785,8 @@ public final class PublicClientApplication {
         String requestAuthority = authority;
 
         if (StringUtil.isEmpty(requestAuthority)) {
-            if (Authority.getAuthorityFromAuthorityUrl(mAuthorityString) instanceof AzureActiveDirectoryB2CAuthority) {
+            if (!StringUtil.isEmpty(mAuthorityString)
+                    && Authority.getAuthorityFromAuthorityUrl(mAuthorityString) instanceof AzureActiveDirectoryB2CAuthority) {
                 requestAuthority = AzureActiveDirectoryB2CAuthority.getAuthorityFromAccount(account);
             } else {
                 requestAuthority = Authority.getAuthorityFromAccount(account);
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -787,7 +787,7 @@ public final class PublicClientApplication {
         if (StringUtil.isEmpty(requestAuthority)) {
             if (!StringUtil.isEmpty(mAuthorityString)
                     && Authority.getAuthorityFromAuthorityUrl(mAuthorityString) instanceof AzureActiveDirectoryB2CAuthority) {
-                requestAuthority = AzureActiveDirectoryB2CAuthority.getAuthorityFromAccount(account);
+                requestAuthority = AzureActiveDirectoryB2CAuthority.getAuthorityFromAccount(account, mAuthorityString);
             } else {
                 requestAuthority = Authority.getAuthorityFromAccount(account);
             }
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -39,6 +39,7 @@ import com.microsoft.identity.client.exception.MsalException;
 import com.microsoft.identity.client.internal.MsalUtils;
 import com.microsoft.identity.client.internal.authorities.Authority;
 import com.microsoft.identity.client.internal.authorities.AzureActiveDirectoryAudience;
+import com.microsoft.identity.client.internal.authorities.AzureActiveDirectoryB2CAuthority;
 import com.microsoft.identity.client.internal.configuration.AuthorityDeserializer;
 import com.microsoft.identity.client.internal.configuration.AzureActiveDirectoryAudienceDeserializer;
 import com.microsoft.identity.client.internal.configuration.LogLevelDeserializer;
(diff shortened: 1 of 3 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -784,7 +785,12 @@ public void acquireTokenSilentAsync(@NonNull final String[] scopes,
         String requestAuthority = authority;
 
         if (StringUtil.isEmpty(requestAuthority)) {
-            requestAuthority = Authority.getAuthorityFromAccount(account);
+            if (Authority.getAuthorityFromAuthorityUrl(mAuthorityString) instanceof AzureActiveDirectoryB2CAuthority) {
```
iambmelt: Check if `null == mAuthorityString`

iambmelt: Please review this logic :-)

msal/src/main/java/com/microsoft/identity/client/internal/authorities/AzureActiveDirectoryB2CAuthority.java
```diff
@@ -58,4 +63,32 @@ public OAuth2Strategy createOAuth2Strategy() {
         config.setAuthorityUrl(this.getAuthorityURL());
         return new MicrosoftStsOAuth2Strategy(config);
     }
+
+    public static String getAuthorityFromAccount(final IAccount account) {
```
iambmelt: This method LGTM

msal/src/main/java/com/microsoft/identity/client/internal/authorities/Authority.java
```diff
@@ -44,7 +44,7 @@
     private static final String TAG = Authority.class.getSimpleName();
 
     private static final String ADFS_PATH_SEGMENT = "adfs";
-    private static final String B2C_PATH_SEGMENT = "b2c";
+    private static final String B2C_PATH_SEGMENT = "tfp";
```
iambmelt: This part LGTM

heidijinxujia:
@iambmelt Pushed the fix. Turns out we should just use the passed-in authorityUrl `https://[authority]/tfp/[tenant]/[policy]`. Did the E2E test, it sticks this time. Sent you the video link on Teams.

msal/src/main/java/com/microsoft/identity/client/internal/authorities/Authority.java
addev-ashish: mAuthorityTypeString in the AzureActiveDirectoryB2CAuthority (https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/a88cf0e6b7a4f1b6fa52363fb1044c1d4b422301/msal/src/main/java/com/microsoft/identity/client/internal/authorities/AzureActiveDirectoryB2CAuthority.java#L37) does this need to change too?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -784,7 +785,13 @@ public void acquireTokenSilentAsync(@NonNull final String[] scopes,
         String requestAuthority = authority;
 
         if (StringUtil.isEmpty(requestAuthority)) {
-            requestAuthority = Authority.getAuthorityFromAccount(account);
+            if (!StringUtil.isEmpty(mAuthorityString)
+                    && Authority.getAuthorityFromAuthorityUrl(mAuthorityString) instanceof AzureActiveDirectoryB2CAuthority) {
```
addev-ashish: Authority.getAuthorityFromAuthorityUrl can throw an illegalargumentexception, since we are doing this last moment, see if you can make this more defensive

addev-ashish: also can you add a comment on why this logic is this way

msal/src/main/java/com/microsoft/identity/client/internal/authorities/Authority.java
heidijinxujia: No, it needn't. This should be the same with the authority type in config file. @shoatman Please confirm.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -807,6 +809,28 @@ public void acquireTokenSilentAsync(@NonNull final String[] scopes,
         MSALApiDispatcher.submitSilent(silentTokenCommand);
     }
 
+    private String getSilentRequestAuthority(final IAccount account) {
```
shoatman: Why wouldn't we move this method to the authority class?

msal/src/main/java/com/microsoft/identity/client/internal/authorities/Authority.java
shoatman: Yes this should be tfp.  Looks like b2c was set in error... i'm assuming by me ;-)

heidijinxujia: Hi @shoatman sorry, I'm bit confused. In the config.json, for the B2C authority, should we lable the authority type as "B2C" or "tfp"?

I think "B2C" is a more seasonable type. Because there are two kinds of authority schema for B2C scenario. "tfp" is just one of them. I've asked @danieldobalian for the B2C wiki link.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: Which public API are you referring to?  I meant why wouldn't we move getSilentRequestAuthority to the Authority class?  The Authority class is in the internal package... hence not part of the public API

heidijinxujia: Ah, I got what you mean here. I will make the fix after release.

PR #423, opened 2018-10-12
Prepares release/0.2.1

### LINKED ISSUE #420: if ext_expires_in becomes > expires_in, MSAL will return an expired access token
See the call here
https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java#L316

and implementation here
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/dto/AccessTokenRecord.java#L262

Not a burning issue yet as AAD v2 is still returning same value for both expires_in and ext_expires_in
{"token_type":"Bearer","scope":"https://graph.windows.net/User.Read","expires_in":3599,"ext_expires_in":3599," ....

iambmelt:
Closes #420 

Revs version, changelog

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,5 +1,11 @@
 MSAL Wiki : https://github.com/AzureAD/microsoft-authentication-library-for-android/wiki
 
+Version 0.2.1
+----------------------------
+- Picks up common@0.0.5-hf1 (Hotfix for MSAL/#420)
+    * AccessTokenRecord#isExpired no longer falls back on ext_expires_in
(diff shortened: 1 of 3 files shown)


iambmelt:
Closing, will reopen with a different base branch (`dev`).

Release will still be `0.2.1`, but will be cut from `dev` instead of `release/0.2.0`

PR #439, opened 2018-10-25
first MSAL broker PR

shoatman:
- Not tested yet, but checking in work in progress....

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerActivity.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerActivity.java
@@ -1,0 +1,72 @@
+package com.microsoft.identity.client.internal.controllers;
+
+import android.app.Activity;
+import android.content.Intent;
+import android.os.Bundle;
+
+public final class BrokerActivity extends Activity {
+
+    public static final String BROKER_INTENT = "broker_intent";
+    static final String BROKER_INTENT_STARTED = "broker_intent_started";
+    static final int BROKER_INTENT_REQUEST_CODE = 1001;
+
+    private static final String TAG = BrokerActivity.class.getSimpleName();
+
+    private Intent mBrokerInteractiveRequestIntent;
+    private Boolean mBrokerIntentStarted = false;
+
+
+    @Override
+    protected void onCreate(final Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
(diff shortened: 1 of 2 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerActivity.java
```diff
@@ -0,0 +1,95 @@
...
+        if (!mBrokerIntentStarted) {
+            mBrokerIntentStarted = true;
+            startActivityForResult(mBrokerInteractiveRequestIntent, BROKER_INTENT_REQUEST_CODE);
+        }
+
+    }
+
+    @Override
+    protected void onStop() {
+        super.onStop();
```
iambmelt: Can this go or are we planning on placing additional logic inside this method?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerActivity.java
```diff
@@ -0,0 +1,95 @@
...
+
+import android.app.Activity;
+import android.content.Intent;
+import android.os.Bundle;
+
+public final class BrokerActivity extends Activity {
+
+    public static final String BROKER_INTENT = "broker_intent";
+    static final String BROKER_INTENT_STARTED = "broker_intent_started";
+    static final int BROKER_INTENT_REQUEST_CODE = 1001;
```
iambmelt: I recall @danieldobalian asking whether or not this value could change between MSAL/ADAL. Possibly we address that?

heidijinxujia: request code `1001` is for msal/adal ui flow as well.

shoatman: this was actually not for broker....

shoatman: I'm following Ashish's spec.  @addev-ashish - Do you want to select a different request_code?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerActivity.java
shoatman: I'll remove if I end up not logging something from here.  Leaving for now... in case I need to investigate anything.

PR #444, opened 2018-10-31
Implemented MSAL Controller Factory

shoatman:
- Implemented MSAL Controller Factory.  Returning controller or controllers based on the scenario.
- Updated MsalTokenCommand (Silent) to execute acquireTokenSilent on both controllers when necessary
- Updated PublicClientApplication to use the controller factory


msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALTokenCommand.java
```diff
@@ -60,9 +64,51 @@ public MSALTokenCommand(@NonNull final Context context,
...
+                            @NonNull final MSALOperationParameters parameters,
+                            @NonNull final List<MSALController> controllers,
+                            @NonNull final AuthenticationCallback callback) {
+        mContext = context;
+        mParameters = parameters;
+        mController = null;
+        mControllers = controllers;
+        mCallback = callback;
+
+        if (!(mParameters instanceof MSALAcquireTokenSilentOperationParameters)) {
```
iambmelt: Is this check needed? Or could we change the constructor args to explicitly require this type?

```java
public MSALTokenCommand(@NonNull final Context context,
                            @NonNull final MSALAcquireTokenSilentOperationParameters parameters,
                            @NonNull final List<MSALController> controllers,
    // ...
}
```

msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.java
```diff
@@ -22,25 +22,143 @@
...
+     * Audience != AnyPersonalAccounts
+     * Broker Installed & Verified
+     *
+     *
+     * @param applicationContext
+     * @param authority
+     * @param applicationConfiguration
+     * @return
+     */
+    public static boolean brokerEligible(Context applicationContext, Authority authority, PublicClientApplicationConfiguration applicationConfiguration){
```
iambmelt: Similar question here regarding types - possible to avoid `instanceof` via explicit param types?

iambmelt: Not an issue per se, but might consider naming this method `isBrokerEligible()`

shoatman: No.  Why would i restrict the types ahead of time?

shoatman: Ok.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALTokenCommand.java
shoatman: This part isn't actually new in this PR.

PR #456, opened 2018-11-12
Adding acquireTokenParameters and acquireTokenSilentParameters

shoatman:
- cleaned up all member variables on PublicClientApplication and migrated to PublicClientApplicationConfiguration
- Added TokenParameters, AquireTokenParameters and AcquireTokenSilentParameters
- Added acquireToken and acquireTokenSilent overloads accepting these methods
- Moved construction of MsalOperatioParameters to their respective classes from PublicClientApplication
- Refactored existing overloads mapping their parameters into the new parameter objects
- Refactored existing overloads to call the 2 new overloads accepting the parameters objects

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/AcquireTokenParameters.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AcquireTokenParameters.java
@@ -1,9 +1,7 @@
-package com.microsoft.identity.client.parameters;
+package com.microsoft.identity.client;
 
 import android.app.Activity;
 import android.util.Pair;
-
-import com.microsoft.identity.client.UiBehavior;
 
 import java.util.List;
 
--- a/msal/src/main/java/com/microsoft/identity/client/AcquireTokenParameters.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AcquireTokenParameters.java
@@ -1,3 +1,26 @@
+//  Copyright (c) Microsoft Corporation.
+//  All rights reserved.
+//
+//  This code is licensed under the MIT License.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files(the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions :
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
 package com.microsoft.identity.client;
 
 import android.app.Activity;
@@ -5,6 +28,9 @@ import android.util.Pair;
 
 import java.util.List;
 
+/**
+ * Encapsulates the parameters passed to the acquireToken methods of PublicClientApplication
+ */
 public class AcquireTokenParameters extends TokenParameters {
 
     private Activity mActivity;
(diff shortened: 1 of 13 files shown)


msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
```diff
@@ -0,0 +1,151 @@
...
+     * {@link AuthenticationCallback#onSuccess(AuthenticationResult)}
+     * 3) All the other errors will be sent back via
+     * {@link AuthenticationCallback#onError(com.microsoft.identity.client.exception.MsalException)}.
+     * @param callback
+     */
+    public void setCallback(AuthenticationCallback callback) {
+        this.mCallback = callback;
+    }
+
+    void setAccountRecord(AccountRecord record){
```
kreedula: Nit : java doc missing. Also should setAccoutRecord be public?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -779,34 +675,54 @@ public void acquireToken(@NonNull final Activity activity,
                              @Nullable final String[] extraScopesToConsent,
                              @Nullable final String authority,
                              @NonNull final AuthenticationCallback callback) {
-        String loginHint = null;
 
-        if (null != account) {
-            loginHint = account.getUsername();
-        }
+        AcquireTokenParameters acquireTokenParameters = new AcquireTokenParameters();
```
kreedula: What do you think creating a private `acquireToken method` which accepts all the parameters and all the overloads just call this method and `AcquireTokenParameters` are created inside it to avoid code duplication like below

`    private void acquireToken(@NonNull final Activity activity,
                             @Nullable final IAccount account,
                             @Nullable final String loginHint,
                             @NonNull final UiBehavior uiBehavior,
                             @Nullable final List<Pair<String, String>> extraQueryParameters,
                             @Nullable final String[] extraScopesToConsent,
                             @Nullable final String authority,

        AcquireTokenParameters acquireTokenParameters = new AcquireTokenParameters();
        acquireTokenParameters.setLoginHint(loginHint);
        acquireTokenParameters.setAccount(account);
        acquireTokenParameters.setActivity(activity);
        acquireTokenParameters.setScopes(Arrays.asList(scopes));
        acquireTokenParameters.setUIBehavior(uiBehavior);
        acquireTokenParameters.setExtraQueryStringParameters(extraQueryParameters);
        acquireTokenParameters.setAccountRecord(getAccountRecord(acquireTokenParameters.getAccount()));
        acquireTokenParameters.setCallback(callback);

        acquireToken(acquireTokenParameters);


    }`

msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
```diff
@@ -0,0 +1,151 @@
...
+        mAccountRecord = record;
+    }
+
+    public AccountRecord getAccountRecord(){
```
kreedula: Nit : java doc missing.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: @kreedula I'd suggest this trades code duplication for the method-with-too-many-parameters problem.

If we're looking to clean this up, I'd prefer a builder object. @shoatman can make the call, I'd say drop it on the backlog

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -779,34 +675,54 @@ public void acquireToken(@NonNull final Activity activity,
...
-                        scopes,
-                        loginHint,
-                        uiBehavior,
-                        extraQueryParameters,
-                        extraScopesToConsent,
-                        authority,
-                        account
-                );
+
+        acquireTokenParameters.setAccountRecord(getAccountRecord(acquireTokenParameters.getAccount()));
```
iambmelt: Ooph some gymnastics going on here... Not immediately apparent how we might clean this up - perhaps we consider passing a homeAccountId and hitting the cache later?

msal/src/main/java/com/microsoft/identity/client/AcquireTokenParameters.java
```diff
@@ -0,0 +1,127 @@
...
+    private UiBehavior mUIBehavior;
+    private List<String> mExtraScopesToConsent;
+    private List<Pair<String, String>> mExtraQueryStringParameters;
+
+
+    /**
+     * Non-null {@link Activity} that will be used as the parent activity for launching the {@link AuthenticationActivity}
+     * @return
+     */
+    public Activity getActivity() {
```
iambmelt: Something we might want to consider: if we declare this to be a `Context` we could add support for [`Services`](https://developer.android.com/reference/android/app/Service) to support silent background scenarios

msal/src/main/java/com/microsoft/identity/client/AcquireTokenParameters.java
```diff
@@ -0,0 +1,127 @@
...
+     */
+    public UiBehavior getUIBehavior() {
+        return mUIBehavior;
+    }
+
+    /**
+     * Controls the value of the prompt parameter sent along with the authorization request.
+     * @param uiBehavior
+     */
+    public void setUIBehavior(UiBehavior uiBehavior) {
```
iambmelt: Nit: Can we call this `setUiBehavior` to match the capitalization of the param?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -779,34 +675,54 @@ public void acquireToken(@NonNull final Activity activity,
...
+
+    /**
+     * Acquire token interactively, will pop-up webUI. Interactive flow will skip the cache lookup.
+     * Default value for {@link UiBehavior} is {@link UiBehavior#SELECT_ACCOUNT}.
+     *
+     * Convey parameters via the AquireTokenParameters object
+     *
+     * @param acquireTokenParameters
+     */
+    public void acquireToken(@NonNull AcquireTokenParameters acquireTokenParameters) {
```
iambmelt: 🔥 🔥 🔥

msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
iambmelt: AccountRecord is internal - the external representation is the IAccount

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: I didn't follow you comment.  The method with too many parameters problem could be resolved by removing all existing overloads save the acquireToken(acquireTokenParameters) one.  Is that what you're suggesting?  I was thinking that we don't want to take that scale of breaking change at this time, but I"m open to it.  Thoughts?

I believe the second part of the suggestion is to support the builder patter on the token parameters classes and i'm open to that as well.

msal/src/main/java/com/microsoft/identity/client/AcquireTokenParameters.java
shoatman: Discussed offline

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: Discussed offline.

msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
shoatman: Leaving off for now.  I plan to move these out of here shortly.

PR #457, opened 2018-11-15
 Changes to create AcquireTokenResult from BrokerResult

kreedula:
common PR : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/305

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
@@ -23,6 +23,7 @@
 package com.microsoft.identity.client;
 
 import com.microsoft.identity.common.internal.dto.AccountRecord;
+import com.microsoft.identity.common.internal.dto.IAccountRecord;
 import com.microsoft.identity.common.internal.logging.Logger;
 import com.microsoft.identity.common.internal.providers.microsoft.MicrosoftAccount;
 
@@ -39,7 +40,7 @@ class AccountAdapter {
      * @param accountIn The Account to transform.
      * @return A representation of the supplied Account, as an IAccount.
      */
-    static IAccount adapt(final AccountRecord accountIn) {
+    static IAccount adapt(final IAccountRecord accountIn) {
         final String methodName = ":adapt";
         final com.microsoft.identity.client.Account accountOut
                 = new com.microsoft.identity.client.Account();
--- a/msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
@@ -53,13 +53,11 @@ public final class AuthenticationResult {
     }
 
     public AuthenticationResult(AccessTokenRecord accessToken,
-                                String tenantId,
-                                String uniqueId,
                                 String rawIdToken,
                                 IAccountRecord accountRecord){
         mAccessToken = accessToken;
-        mTenantId = tenantId;
-        mUniqueId = uniqueId;
+        mTenantId = accessToken.getRealm();
+        mUniqueId = accessToken.getHomeAccountId();
         mRawIdToken = rawIdToken;
         mAccount = AccountAdapter.adapt(accountRecord);
 
--- a/msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
@@ -25,6 +25,7 @@ package com.microsoft.identity.client;
 
 import com.microsoft.identity.common.internal.cache.ICacheRecord;
 import com.microsoft.identity.common.internal.dto.AccessTokenRecord;
+import com.microsoft.identity.common.internal.dto.IAccountRecord;
 
 import java.util.Date;
 import java.util.concurrent.TimeUnit;
@@ -49,6 +50,19 @@ public final class AuthenticationResult {
         mUniqueId = cacheRecord.getAccount().getHomeAccountId();
         mRawIdToken = cacheRecord.getIdToken().getSecret();
         mAccount = AccountAdapter.adapt(cacheRecord.getAccount());
+    }
+
+    public AuthenticationResult(AccessTokenRecord accessToken,
+                                String tenantId,
+                                String uniqueId,
+                                String rawIdToken,
+                                IAccountRecord accountRecord){
+        mAccessToken = accessToken;
+        mTenantId = tenantId;
+        mUniqueId = uniqueId;
+        mRawIdToken = rawIdToken;
+        mAccount = AccountAdapter.adapt(accountRecord);
+
     }
 
     /**
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
@@ -183,7 +183,6 @@ public class BrokerMSALController extends MSALController {
         request.setRedirect(parameters.getRedirectUri());
         //NOTE: Assumption: Broker will handle removing empty string scopes and appending stanard OIDC scopes
         request.setScope(StringUtil.join(' ', parameters.getScopes()));
-        request.setApplicationName(parameters.getAppContext().getPackageName());
         //request.setVersion();
 
         return request;
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
@@ -40,6 +40,7 @@ import com.microsoft.identity.common.internal.cache.SchemaUtil;
 import com.microsoft.identity.common.internal.dto.AccessTokenRecord;
 import com.microsoft.identity.common.internal.dto.CredentialType;
 import com.microsoft.identity.common.internal.logging.DiagnosticContext;
+import com.microsoft.identity.common.internal.logging.Logger;
 import com.microsoft.identity.common.internal.providers.microsoft.azureactivedirectory.ClientInfo;
 import com.microsoft.identity.common.internal.providers.microsoft.microsoftsts.MicrosoftStsAccount;
 import com.microsoft.identity.common.internal.providers.oauth2.IDToken;
@@ -52,6 +53,8 @@ import java.util.concurrent.ExecutionException;
  */
 public class BrokerMSALController extends MSALController {
 
+    private static final String TAG = BrokerMSALController.class.getSimpleName();
+
     private BrokerResultFuture mBrokerResultFuture;
 
     @Override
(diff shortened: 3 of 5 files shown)


msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
```diff
@@ -41,7 +42,7 @@
      * @param accountIn The Account to transform.
      * @return A representation of the supplied Account, as an IAccount.
      */
-    static IAccount adapt(final AccountRecord accountIn) {
+    static IAccount adapt(final IAccountRecord accountIn) {
```
iambmelt: 👍

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
```diff
@@ -51,6 +52,19 @@ public AuthenticationResult(final ICacheRecord cacheRecord) {
         mAccount = AccountAdapter.adapt(cacheRecord.getAccount());
     }
 
+    public AuthenticationResult(AccessTokenRecord accessToken,
+                                String tenantId,
```
iambmelt: If this `AccessTokenRecord` is fully-formed, the `realm` property of this object should match the `tenantId` - can we lose this parameter? Unclear what it would mean if these values were different

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
```diff
@@ -51,6 +52,19 @@ public AuthenticationResult(final ICacheRecord cacheRecord) {
...
+                                String uniqueId,
```
iambmelt: Careful.... what is this value?

If it's not the `<uid>.<utid>` it's not going to be the same as the `home_account_id` coming from the `AccountRecord`.

I suspect this value should be sourced from the `AccessTokenRecord` and can be removed as a param

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
```diff
@@ -79,7 +88,7 @@ public AcquireTokenResult acquireToken(MSALAcquireTokenOperationParameters reque
      * @param request
      * @return
      */
-    private Intent getBrokerAuthorizationItent(MSALAcquireTokenOperationParameters request) throws ClientException {
+    private Intent getBrokerAuthorizationIntent(MSALAcquireTokenOperationParameters request) throws ClientException {
```
iambmelt: Hah. Good catch

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
```diff
@@ -136,31 +143,32 @@ public AcquireTokenResult acquireTokenSilent(MSALAcquireTokenSilentOperationPara
...
 
         try {
-            //result = service.acquireTokenSilently(getSilentParameters(request));
-
-            // Path is broken by common/#303 - @kreedula & @shoatman to resolve
-            BrokerResult brokerResult = service.acquireTokenSilently(new BrokerRequest());
+            BrokerResult brokerResult = service.acquireTokenSilently(getBrokerRequest(parameters));
+            AcquireTokenResult acquireTokenResult = new AcquireTokenResult();
+            acquireTokenResult.setTokenResult(brokerResult);
+            if (brokerResult.isSuccessful() && brokerResult.getTokenResponse() != null) {
```
iambmelt: Are we being defensive here, or is there some case where `true == brokerResult#isSuccessful()` _and_ the response is `null`?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
```diff
@@ -169,16 +177,51 @@ private Map getSilentParameters(MSALAcquireTokenSilentOperationParameters parame
         request.setForceRefresh(parameters.getForceRefresh());
         request.setLoginHint(parameters.getAccount().getUsername());
         request.setName(parameters.getAccount().getUsername());
-        request.setUserId(parameters.getAccount().getHomeAccountId());
+        request.setUserId(parameters.getAccount().getLocalAccountId());
```
iambmelt: The goal here is to have the tenant-specific value, right? To support guest scenarios?

If **yes**, this is good.

If **no**, what's going on?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
```diff
@@ -169,16 +177,51 @@ private Map getSilentParameters(MSALAcquireTokenSilentOperationParameters parame
...
         //request.setPrompt(parameters.get);
         //TODO: This should be the broker redirect URI and not the non-broker redirect URI
         request.setRedirect(parameters.getRedirectUri());
         //NOTE: Assumption: Broker will handle removing empty string scopes and appending stanard OIDC scopes
         request.setScope(StringUtil.join(' ', parameters.getScopes()));
+        request.setApplicationName(parameters.getAppContext().getPackageName());
```
iambmelt: Setting this value 2x?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
```diff
@@ -169,16 +177,51 @@ private Map getSilentParameters(MSALAcquireTokenSilentOperationParameters parame
...
+    }
+
+    private AuthenticationResult getAuthenticationResult(BrokerTokenResponse brokerTokenResponse){
+        try {
+            ClientInfo clientInfo = new ClientInfo(brokerTokenResponse.getClientInfo());
+            String homeAccountId = SchemaUtil.getHomeAccountId(clientInfo);
+            String idToken = brokerTokenResponse.getIdToken();
+            String tenantId = clientInfo.getUtid();
+
+            AccessTokenRecord accessTokenRecord = new AccessTokenRecord();
```
iambmelt: Do we need to set the `local_account_id` here? I suspect **YES**: this value is surfaced to users through `IAccount#getAccountIdentifier` in `AuthenticationResult#getAccount`

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
kreedula: Yes. just being defensive, there should not be any case as such.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
kreedula: We are mapping Account in Broker with upn currently. Let's discuss about this tommorow

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
kreedula: Thanks for catching this. Changed as needed

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
kreedula: `AccessTokenRecord` doesn't have a local_account_id right?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
iambmelt: I wouldn't expect either of these values to be what you want then... consider adding a `username` property?

I'd suggest not calling this parameter `upn`, as that's a v1 claim which has been replaced by `preferred_username` in v2.

Per our schema doc

| Field name  |  Description |
|---|---|
| `home_account_id`  | Home account identifier for that authentication scheme. <br/><br/>For AAD/MSA:<br/>`<uid>.<utid>`<br/><br/>STS returns the clientInfo on both v1 and v2 for AAD. This value is combined from two fields in the client Info. It should be a unique account identifier across tenants (e.g. the value will be the same for user’s home tenant and a guest tenant). <br/><br/><br/>(There are other cases where this value might be different, but this affects `NTLM`/`Basic` scenarios and we can ignore it for our purposes)  |
| `local_account_id`  |  Original tenant specific account identifier. <br/><br/>Can be needed for legacy purposes. OID for AAD (in some unique cases subject instead of `OID`) and `CID` for MSA. Comes from the `id_token.oid` and falls back to `id_token.sub`.  |
| `username`  | The primary username that represents the user (corresponds to the preferred_username claim in in the `/v2.0` endpoint). It could be an email address, phone number, or a generic username without a specified format. Its value is mutable and might change over time. For MSA it's email. For `NTLM`, `NTLM` username.  |

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
iambmelt: Just kidding ignore me :-)

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
kreedula: Thanks @iambmelt . This is should be `homeAccountId` in case of MSAL, changed it back. Currently in ad-accounts we are writing user id (oid) to AccountManager and using that to look up to get the account. So when making changes for joined account cache, if it's written with `homeAccountId`, we should be good.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
```diff
@@ -136,31 +143,32 @@ public AcquireTokenResult acquireTokenSilent(MSALAcquireTokenSilentOperationPara
         }
 
         try {
-            //result = service.acquireTokenSilently(getSilentParameters(request));
-
-            // Path is broken by common/#303 - @kreedula & @shoatman to resolve
-            BrokerResult brokerResult = service.acquireTokenSilently(new BrokerRequest());
+            BrokerResult brokerResult = service.acquireTokenSilently(getBrokerRequest(parameters));
+            AcquireTokenResult acquireTokenResult = new AcquireTokenResult();
+            acquireTokenResult.setTokenResult(brokerResult);
```
iambmelt: This line is throwing a build error - it looks like the submodule pointer needs to be updated for this PR

```
:msal:compileLocalDebugJavaWithJavac/home/travis/build/AzureAD/microsoft-authentication-library-for-android/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java:148: error: incompatible types: BrokerResult cannot be converted to TokenResult
            acquireTokenResult.setTokenResult(brokerResult);
                                              ^
Note: /home/travis/build/AzureAD/microsoft-authentication-library-for-android/msal/src/main/java/com/microsoft/identity/client/internal/telemetry/PlatformIdHelper.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
Note: Some messages have been simplified; recompile with -Xdiags:verbose to get full output
1 error
 FAILED
```

kreedula:
Thanks @iambmelt . Updated submodule, builds fine now.

PR #464, opened 2018-11-27
Fixes for guest scenarios, extends Common/#318

iambmelt:
Extends https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/311

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -511,14 +511,14 @@ public final class PublicClientApplication {
         acquireToken(
                 activity,
                 scopes,
-                null,
-                null,
-                null,
-                null,
-                null,
+                null, // account
+                null, // uiBehavior
+                null, // extraQueryParams
+                null, // extraScopes
+                null, // authority
                 callback,
-                null,
-                null
+                null, // loginHint
+                null // claimsRequest
         );
     }
 
@@ -549,14 +549,14 @@ public final class PublicClientApplication {
         acquireToken(
                 activity,
                 scopes,
-                null,
-                null,
-                null,
-                null,
-                null,
+                null, // account
+                null, // uiBehavior
+                null, // extraQueryParams
+                null, // extraScopes
+                null, // authority
                 callback,
                 loginHint,
-                null
+                null // claimsRequest
         );
     }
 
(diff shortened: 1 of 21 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -406,9 +406,15 @@ public void run() {
      * @param homeAccountIdentifier The home_account_id of the sought IAccount.
      * @return The IAccount stored in the cache or null, if no such matching entry exists.
      */
-    public IAccount getAccount(final String homeAccountIdentifier) {
+    public IAccount getAccount(@NonNull final String homeAccountIdentifier,
```
iambmelt: If we were >v1.0.0, this would be a API major breaking change. Since we still <1.0.0, I'm feeling empowered to make this change.

iambmelt: I've left the `realm` as `@Nullable` since this is IdP specific

iambmelt:
New method annotations relate to #471

iambmelt:
Consider the following changes if merges

```java
index 16d43c3..f4f5abb 160000
--- a/common
+++ b/common
@@ -1 +1 @@
-Subproject commit 16d43c321d332b4c6014af78b2a330f6ec7f0f26
+Subproject commit f4f5abba26ee60aff467029a64dd03df1ad2793d
diff --git a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
index 37a106a..05d5868 100644
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -450,14 +450,19 @@ public final class PublicClientApplication {
             return false;
         }

-        final String realm = getRealm(account);
-
-        return mPublicClientConfiguration.getOAuth2TokenCache().removeAccount(
-                account.getEnvironment(),
-                mPublicClientConfiguration.getClientId(),
-                account.getHomeAccountIdentifier().getIdentifier(),
-                realm
-        );
+        // FEATURE SWITCH: Set to false to allow deleting Accounts in a tenant-specific way.
+        final boolean deleteAccountsInAllTenants = true;
+
+        final String realm = deleteAccountsInAllTenants ? null : getRealm(account);
+
+        return !mPublicClientConfiguration
+                .getOAuth2TokenCache()
+                .removeAccount(
+                        account.getEnvironment(),
+                        mPublicClientConfiguration.getClientId(),
+                        account.getHomeAccountIdentifier().getIdentifier(),
+                        realm
+                ).isEmpty();
     }

     @Nullable

```

PR #467, opened 2018-11-29
Changes to add extra qp and claims

kreedula:


msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
```diff
@@ -188,6 +190,9 @@ private static BrokerRequest getBrokerRequestForInteractive(MSALAcquireTokenOper
         request.setName(parameters.getLoginHint());
         request.setRedirect(parameters.getRedirectUri());
         request.setScope(StringUtil.join(' ', parameters.getScopes()));
+        String extraQP = BrokerRequest.getJsonStringForExtraQueryParams(parameters.getExtraQueryStringParameters());
```
iambmelt: See comment [here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/313#discussion_r237582637) regarding `BrokerRequest` -- might be better to let the `QueryParamsAdapter` handle this

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMSALController.java
```diff
@@ -161,18 +165,16 @@ private static BrokerRequest getBrokerRequestForSilent(MSALAcquireTokenSilentOpe
...
-        //request.setExtraQueryStringParameter();
         request.setForceRefresh(parameters.getForceRefresh());
         request.setLoginHint(parameters.getAccount().getUsername());
         request.setName(parameters.getAccount().getUsername());
         request.setUserId(parameters.getAccount().getHomeAccountId());
-        //request.setPrompt(parameters.get);
         //TODO: This should be the broker redirect URI and not the non-broker redirect URI
         request.setRedirect(parameters.getRedirectUri());
         request.setScope(StringUtil.join(' ', parameters.getScopes()));
+        request.setClaims(ClaimsRequest.getJsonStringFromClaimsRequest(parameters.getClaimsRequest()));
```
iambmelt: `ClaimsRequests` could delegate de/serialization to `ClaimsRequestSerializer` and `ClaimsRequestDeserializer`.

I can also buy that this should stay as-is since this is by nature _slightly_ different than the `QueryParamsAdapter` since, unlike the `List` we use to hold params, we have an actual `class` which can handle its own serialization.

kreedula: `ClaimsRequestSerializer` and `ClaimsRequestDeserializer` are package protected and cannot be accessed in ad-accounts. So leaving it as it to get it from `ClaimsRequests`

kreedula:
Thanks @iambmelt

PR #475, opened 2018-12-11
Refactor to make acquire token calls from common directly

kreedula:


msal/src/main/java/com/microsoft/identity/client/IAuthenticationResult.java
```diff
@@ -25,14 +25,26 @@
...
+public interface IAuthenticationResult extends IBaseAuthenticationResult {
+
+
+    /**
+     * Gets the Account.
+     *
+     * @return The Account to get.
+     */
+    @NonNull
+    IAccount getAccount();
```
iambmelt: If this interface inheritance structure stays, we can remove these `@Override` annotations and the re-declaration.

I'm still pondering this interface design....

iambmelt: At a minimum, we should remove the javadoc so that it doesn't need to be updated in two places. It will be automatically surfaced by any decent editor from the interface superclass anyway

iambmelt: Feedback I left in https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/320/

We should separate these interfaces before deployment. Backlog item.

msal/src/main/java/com/microsoft/identity/client/UiBehavior.java
```diff
@@ -59,4 +61,17 @@ public String toString() {
                 throw new IllegalArgumentException();
         }
     }
+
+    public static OpenIdConnectPromptParameter mapUIBehaviorToOpenIdConnect(UiBehavior uiBehavior){
```
iambmelt: This utility method doesn't feel at-home to me here. This behavior feels very adapter-like (but this _also_ doesn't seem worth making a whole new object for...)

How about an instance method on this enum called `toOpenIdConnectPromptParameter()`?

iambmelt: So you could go....

`UiBehavior.SELECT_ACCOUNT.toOpenIdConnectPromptParameter()`

msal/src/main/java/com/microsoft/identity/client/claims/ClaimsRequest.java
```diff
@@ -1,156 +0,0 @@
...
-import com.google.gson.GsonBuilder;
-
-import java.util.ArrayList;
-import java.util.List;
-
-/**
- * Represents the claims request parameter as an object
- * @see <a href="https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter">https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter</a>
- */
-public class ClaimsRequest {
```
iambmelt: This is deleting public API -- this class is exposed through `PublicClientApplication`.

❌

msal/src/main/java/com/microsoft/identity/client/claims/RequestedClaim.java
```diff
@@ -1,67 +0,0 @@
...
-//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-//  THE SOFTWARE.
-package com.microsoft.identity.client.claims;
-
-/**
- * Represents an individual requested claims that's part of a complete claims request parameter
- * @see <a href="https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter">https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter</a>
- */
-public class RequestedClaim {
```
iambmelt: Also a public class, exposed through `ClaimsRequest`

❌

msal/src/main/java/com/microsoft/identity/client/claims/RequestedClaimAdditionalInformation.java
```diff
@@ -1,70 +0,0 @@
...
-import com.google.gson.annotations.SerializedName;
-
-import java.util.ArrayList;
-import java.util.List;
-
-/**
- * Represents the additional information that can be sent to an authorization server for a request claim in the claim request parameter
- * @see <a href="https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter">https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter</a>
- */
-public class RequestedClaimAdditionalInformation {
```
iambmelt: Also a public class, exposed through `ClaimsRequest`

❌

msal/src/main/java/com/microsoft/identity/client/claims/WWWAuthenticateHeader.java
```diff
@@ -1,100 +0,0 @@
...
-//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-//  THE SOFTWARE.
-package com.microsoft.identity.client.claims;
-
-/**
- * Helper class for pulling any available claims directive out of the WWW-Authenticate header returned
- * by resource servers
- */
-public class WWWAuthenticateHeader {
```
iambmelt: This class is a public utility to support `ClaimsRequest` -- more public API

❌

msal/src/main/java/com/microsoft/identity/client/IAuthenticationResult.java
kreedula: Addressed the issue in common here https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/320 and via this commit here : https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/475/commits/2cb693a9eff22eeaefd9f8a2b119e0e691ef3960

msal/src/main/java/com/microsoft/identity/client/UiBehavior.java
kreedula: I like the idea. Thanks for the suggestion Brian, adopted here

msal/src/main/java/com/microsoft/identity/client/claims/ClaimsRequest.java
kreedula: Thanks a lot for catching this, an oversight on my part to miss this. Fixed it

kreedula:
Thanks @iambmelt for the review and feedback, made changes as appropriate. Also addressed the issue : AzureAD/microsoft-authentication-library-for-android#478

iambmelt:
LGTM, thanks @kreedula!

PR #483, opened 2018-12-19
Do not fail with exit code if javadoc cannot document all files

iambmelt:
This change is to fix the` MSAL Nightly VSTS Release`

The `javadoc` task is failing on project dependency files.

Implications of this change:
- If `javadoc` cannot find the proper file (as it will not for project dependencies), the javadocs will still be generated, but no source link will be provided to the absent file (only the name will be present)
- If the dependency can be resolved (such as from maven or a snapshot) the file should be generated normally


iambmelt:
Testing removing `afterEvaluate` -- trying to diagnose if this is the source of unexpected resolution errors in Travis

PR #493, opened 2019-01-09
Update submodule to latest, fixes test imports

iambmelt:
Picking up changes in https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/330


kreedula:
Thanks @iambmelt

PR #501, opened 2019-01-18
Allow to read any configuration file (not just a resource)

guperrot:
* Add a new constructor that takes File as parameter.
* Modified a few unit tests to test the same configuration content with both resource and generic file.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -1038,18 +1038,19 @@ public final class PublicClientApplication {
         return loadConfiguration(configStream, useDefaultConfigResourceId);
     }
 
-    private void setupConfiguration(Context context) {
-        mPublicClientConfiguration = loadDefaultConfiguration(context);
-        mPublicClientConfiguration.setAppContext(context);
-        mPublicClientConfiguration.setOAuth2TokenCache(getOAuth2TokenCache());
-    }
-
-    private static PublicClientApplicationConfiguration loadConfiguration(File configFile) {
+    @VisibleForTesting
+    static PublicClientApplicationConfiguration loadConfiguration(@NonNull File configFile) {
         try {
             return loadConfiguration(new FileInputStream(configFile), false);
         } catch (FileNotFoundException e) {
             throw new IllegalArgumentException("Provided config file path=" + configFile.getPath() + " not found");
         }
+    }
+
+    private void setupConfiguration(Context context) {
+        mPublicClientConfiguration = loadDefaultConfiguration(context);
+        mPublicClientConfiguration.setAppContext(context);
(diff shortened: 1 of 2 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1014,17 +1053,14 @@ private void setupConfiguration(Context context) {
         mPublicClientConfiguration.setOAuth2TokenCache(getOAuth2TokenCache());
     }
 
-    @VisibleForTesting
-    static PublicClientApplicationConfiguration loadConfiguration(@NonNull final Context context,
-                                                                  final int configResourceId) {
-        InputStream configStream = context.getResources().openRawResource(configResourceId);
+    private static PublicClientApplicationConfiguration loadConfiguration(InputStream configStream, boolean useDefaultConfigResourceId) {
```
shoatman: I'd suggest changing the naming of the boolean to isDefaultConfiguration.  In this case you're communicating to the method that's it being used to load the default configuration or not.  You're not instructing the method to use the default configuration.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -999,13 +1030,21 @@ private void loadMetaDataFromManifest() {
...
 
-    private void setupConfiguration(final int configResourceId, @NonNull final Context context) {
-        final PublicClientApplicationConfiguration developerConfig = loadConfiguration(context, configResourceId);
-        final PublicClientApplicationConfiguration defaultConfig = loadDefaultConfiguration(context);
-        defaultConfig.mergeConfiguration(developerConfig);
-        mPublicClientConfiguration = defaultConfig;
-        mPublicClientConfiguration.setAppContext(context);
-        mPublicClientConfiguration.setOAuth2TokenCache(getOAuth2TokenCache());
+    @VisibleForTesting
+    static PublicClientApplicationConfiguration loadConfiguration(@NonNull final Context context,
```
shoatman: Looks like this method could be refactored out.... it's only purpose now seems to be whether to determine the resource id is the resource id for the default configuration or not.  Since you've already gone most of the way... why not go all the way?

guperrot: No the method handle alternate resource configuration files that are bundled with the application. The boolean is just to keep the existing error message that is different whether it was using default or not.

shoatman: Ack.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
guperrot: I renamed the variable as suggested in aafd2fb.

guperrot:
Thanks a lot for the merge 👍

Issue #510, opened 2019-01-29
Configuration file is not closed after reading, causing strict mode issue


guperrot:
Init a client with either a resource file or a storage file:

```
2019-01-28 16:52:17.077 20658-20673/com.microsoft.appcenter.sasquatch.project D/StrictMode: StrictMode policy violation: android.os.strictmode.LeakedClosableViolation: A resource was acquired at attached stack trace but never released. See java.io.Closeable for information on avoiding resource leaks.
        at android.os.StrictMode$AndroidCloseGuardReporter.report(StrictMode.java:1786)
        at dalvik.system.CloseGuard.warnIfOpen(CloseGuard.java:264)
        at java.io.FileInputStream.finalize(FileInputStream.java:487)
        at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:250)
        at java.lang.Daemons$FinalizerDaemon.runInternal(Daemons.java:237)
        at java.lang.Daemons$Daemon.run(Daemons.java:103)
        at java.lang.Thread.run(Thread.java:764)
     Caused by: java.lang.Throwable: Explicit termination method 'close' not called
        at dalvik.system.CloseGuard.open(CloseGuard.java:221)
        at java.io.FileInputStream.<init>(FileInputStream.java:168)
        at com.microsoft.identity.client.PublicClientApplication.loadConfiguration(PublicClientApplication.java:1044)
        at com.microsoft.identity.client.PublicClientApplication.<init>(PublicClientApplication.java:244)
```

Indeed the code does not close the InputStream after reading.

PR #520, opened 2019-02-04
Close resources after reading the config file.

annakocheshkova:
Fixes the #510 issue.
Need to close the resources after reading the config file to prevent memory leaks.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -1066,6 +1066,23 @@ public final class PublicClientApplication {
                 throw new IllegalArgumentException("Unable to open provided configuration file.", e);
             }
         }
+        finally {
+            try {
+                configStream.close();
+            } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1066,6 +1066,23 @@ private static PublicClientApplicationConfiguration loadConfiguration(InputStrea
                 throw new IllegalArgumentException("Unable to open provided configuration file.", e);
             }
         }
+        finally {
+            try {
+                configStream.close();
+            } catch (IOException e) {
+                if (isDefaultConfiguration) {
+                    com.microsoft.identity.common.internal.logging.Logger.verbose(
```
iambmelt: Can you log this at `Logger.warn` instead of `Logger.verbose`?

annakocheshkova: 👍

iambmelt:
Thanks @annakocheshkova!

PR #533, opened 2019-02-20
Changes to make scopes type as set in operation parameters

kreedula:
Also fixed an compile issue with a moved constant in common

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
@@ -79,6 +79,9 @@ public class LocalMSALController extends BaseController {
         //00) Validate MSAL Parameters
         parameters.validate();
 
+        // Add default scopes
+        addDefaultScopes(parameters);
+
         //0) Get known authority result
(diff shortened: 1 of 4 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
```diff
@@ -172,22 +175,14 @@ public AcquireTokenResult acquireTokenSilent(
...
-            parameters.getScopes().removeAll(Arrays.asList("", null));
-        }
-
         final AcquireTokenResult acquireTokenSilentResult = new AcquireTokenResult();
 
         //Validate MSAL Parameters
         parameters.validate();
 
+        // Add default scopes
+        addDefaultScopes(parameters);
```
heidijinxujia: If the default scopes string is required, would it be better to change the order here and add the scope validation inside `parameters.validate();`?

kreedula: Not really right, that's the validation for input request parameters. Also if we did this we would need to make changes for scope validation in base class as well where we check if scopes are not empty (User provided atleast one) but adding these would break that behavior

PR #538, opened 2019-03-13
Fix the compile error and update submodule

heidijinxujia:


iambmelt:
Thanks for keeping this up-to-date, @heidijinxujia!

PR #548, opened 2019-03-25
Add method to silently refresh token synchronously

### LINKED ISSUE #544: Add method to synchronously acquire a token
This would be similar to `acquireTokenSilentAsync`, but would instead return the `AuthenticationResult` _synchronously_ and throw the `MsalException` if there was an error.

This would be useful in scenarios where the work is already happening in the background, so an async callback just adds complexity to the code. For example, OkHttp request interceptors work synchronously, so if I want to obtain an auth token and add it to each request that is intercepted, I have to somehow figure out how to block on the `acquireTokenSilentAsync` method.

My hacky temporary solution uses rx to provide a blocking mechanism:

```kotlin
private fun PublicClientApplication.acquireTokenSilent(
  scopes: Array<String>,
  account: IAccount
): AuthenticationResult {
  return Single.defer {
    Single.create<AuthenticationResult> { observer ->
      acquireTokenSilentAsync(scopes, account, object : AuthenticationCallback {
        override fun onSuccess(authenticationResult: AuthenticationResult) {
          observer.onSuccess(authenticationResult)
        }

        override fun onCancel() {
          observer.onError(OperationCanceledException("acquireTokenSilentAsync() was canceled"))
        }

        override fun onError(exception: MsalException) {
          observer.onError(exception)
        }
      })
    }
  }.blockingGet()
}
```

joshfriend:
closes #544 

I based this mostly on [what ADAL does currently](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/800c89690fd8a1b5daf27ffcf04954cf3efce010/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java#L568-L714)

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -30,14 +30,17 @@ import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageManager;
 import android.os.Handler;
 import android.os.Looper;
+import android.os.NetworkOnMainThreadException;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.annotation.VisibleForTesting;
+import android.support.annotation.WorkerThread;
 import android.util.Pair;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.microsoft.identity.client.claims.ClaimsRequest;
+import com.microsoft.identity.client.exception.MsalClientException;
 import com.microsoft.identity.client.exception.MsalException;
 import com.microsoft.identity.client.internal.MsalUtils;
 import com.microsoft.identity.client.internal.configuration.LogLevelDeserializer;
@@ -95,6 +98,8 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static com.microsoft.identity.common.internal.cache.SharedPreferencesAccountCredentialCache.DEFAULT_ACCOUNT_CREDENTIAL_SHARED_PREFERENCES;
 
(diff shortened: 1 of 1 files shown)


shoatman:
Thanks for this contribution Josh.  I want to take a look at naming here and whether it's reasonable to rely on a developer reading the Java Doc to know that they shouldn't call this method from the main thread.  We are introducing other methods, in the next release where something similar would potentially make sense, so want to be comfortable with whatever pattern we elect to adopt before proceeding.  What's your sense of how much of problem this is or could be?

joshfriend:
I just pushed an update that enforces the no main-thread calls mentioned in the docs.

If it detects that it is being called on the main thread looper, it will throw a `NetworkOnMainThreadException` and log a helpful error message. This is basically [what `penaltyDeathOnNetwork()` from StrictMode does.](https://developer.android.com/reference/android/os/StrictMode.ThreadPolicy.Builder.html#penaltyDeathOnNetwork())

The `@WorkerThread` annotations from android support lib allow some linting to be done ahead of time to preempt main thread use. For example:



Regarding the name of the method, to me the "Sync" part of the name implies blocking which should not be done in the UI. Those three factors make me feel as if I've done due diligence to prevent incorrect usage.

One thing I think might be improved upon would be to have `Logger` connected to logcat by default, or at least when `BuildConfig.DEBUG` is `true` so as to not hide important error logs by default, but that seems outside the scope of this PR.

You are of course welcome to implement this a different way. I'm not very familiar with the library yet and just based this implementation off of ADAL. I figured at minimum it would be helpful for those who stumble on #544 and dont want to use RxJava to have an alternate implementation.

shoatman:
Thanks @joshfriend.  Very helpful.  I'll review with @danieldobalian and we'll get back to you.

shoatman:
@joshfriend - Here's the proposed implementation for sync methods in MSAL.  

Interface: https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/c904b4251b594ccd911e0fa93cc09f58223bb058/msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java#L107-L117

Implementation: https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/c904b4251b594ccd911e0fa93cc09f58223bb058/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java#L1374-L1415

Feedback welcome!

shoatman:
The draft PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/668/files

PR #550, opened 2019-03-26
Msal-Broker interaction changes

kreedula:


msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -0,0 +1,184 @@
...
+        IMicrosoftAuthService service;
+
+        final MicrosoftAuthClient client = new MicrosoftAuthClient(request.getAppContext());
+        final MicrosoftAuthServiceFuture authServiceFuture = client.connect();
+
+        try {
+            service = authServiceFuture.get();
+            interactiveRequestIntent = service.getIntentForInteractiveRequest();
+
+            // TODO : See what's the right exception here
```
iambmelt: My assumption would be that these errors roll-up to `ClientException` unless a good case can be made for any other error

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerActivity.java
```diff
@@ -89,8 +89,18 @@ protected void onSaveInstanceState(final Bundle outState) {
...
+        final String methodName = ":onActivityResult";
+        Logger.info(TAG + methodName,
+                "Result received from Broker "
+                        + "Request code: " + requestCode
+                        +  " Result code: " + requestCode
+        );
+
+        if (resultCode == AuthenticationConstants.UIResponse.TOKEN_BROKER_RESPONSE ||
+                resultCode == AuthenticationConstants.UIResponse.BROWSER_CODE_CANCEL
+                || requestCode == AuthenticationConstants.UIResponse.BROWSER_CODE_ERROR) {
```
heidijinxujia: requestCode  ->resultCode ?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -0,0 +1,184 @@
...
+    @Override
+    public AcquireTokenResult acquireTokenSilent(AcquireTokenSilentOperationParameters parameters) throws BaseException {
+        IMicrosoftAuthService service;
+
+        MicrosoftAuthClient client = new MicrosoftAuthClient(parameters.getAppContext());
+        MicrosoftAuthServiceFuture future = client.connect();
+
+        try {
+            //Do we want a time out here?
+            service = future.get();
```
heidijinxujia: Yeap, and we could call MicrosoftAuthServiceFuture.get(long l, TimeUnit timeUnit) instead to set the timeout.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.java
```diff
@@ -88,10 +88,11 @@ public static BaseController getAcquireTokenController(@NonNull final Context ap
                                                                         @NonNull final Authority authority,
                                                                         @NonNull final PublicClientApplicationConfiguration applicationConfiguration) {
         List<BaseController> controllers = new ArrayList<>();
-        controllers.add(new LocalMSALController());
         if (brokerEligible(applicationContext, authority, applicationConfiguration)) {
-            controllers.add(new BrokerMSALController());
+            controllers.add(new BrokerMsalController());
         }
+        controllers.add(new LocalMSALController());
```
heidijinxujia: Why does it need to add both lcoal msal controller as well when the broker msal controller is already added?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
kreedula: Agreed, missed to resolve todo's , fixed them

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
kreedula: Yes the question is what's the time limit though, I will create another work item to discuss and add this we needed

msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.java
kreedula: Reverted the change back. The idea behind the design is we look in local msal first and then broker

PR #565, opened 2019-04-08
[WIP] TenantProfile Discovery Spike

iambmelt:
Expands:
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/461

Outstanding:
- Resolve in-code `// TODOs`
    * How should `isHomeTenant()` work for B2C?
        - Could return `true` if user is AAD?
            * This is weird, so probably not
        - Could return `Boolean` object wrapper-type which can be `null`
        - Could remove this method entirely?
    * How should the `id` be determined for the `TenantProfiles`
        - `oid` claim will work for V2, but not for B2C
        - `sub` claim will work for B2C
        - How can I tell which I should use?
- Wire-up to `getAccounts()`
    * How should `acquireToken` work once the `IAccount` goes away?
- Figure out how this is going to work with Broker since I need to hit the cache more than once...
- Straighten-out `IAccount` / `Account` insanity
    * There are multiple classes with these names/interfaces
    * Determine what should stay and what should go


msal/src/main/java/com/microsoft/identity/client/profile/AccountAdapter.java
```diff
@@ -0,0 +1,93 @@
...
+        // First, bucket the AccountRecords on the home_account_id to remove duplicates
+        final Map<String, AccountRecord> uniquePrinciples = new HashMap<>();
+
+        for (final AccountRecord record : allAccountRecords) {
+            uniquePrinciples.put(record.getHomeAccountId(), record);
+        }
+
+        for (final AccountRecord record : uniquePrinciples.values()) {
+            final Account account = new Account();
+            account.setAccountType(AccountType.AzureAD); // TODO How should this be determined?
```
iambmelt: I have an [internal StackOverflow question](https://stackoverflow.microsoft.com/questions/142584/how-can-i-determine-if-an-idtoken-was-issued-by-b2c) on this topic.

msal/src/main/java/com/microsoft/identity/client/profile/TenantProfile.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * The authority - could be a String? Do we need any of these methods?
+     */
+    private String mAuthority;
+
+    @NonNull
+    @Override
+    public String getId() {
+        return (String) getClaims().get(MicrosoftIdToken.OBJECT_ID);
```
iambmelt: This claim may not work for B2C accounts.... it could be the `sub`?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -462,9 +463,21 @@ public void onMigrationFinished(int numberOfAccountsMigrated) {
         }
     }
 
+    public List<com.microsoft.identity.client.profile.IAccount> getAccountsNew() {
```
iambmelt: This is just testing surface. Not going to be the final API

msal/src/main/java/com/microsoft/identity/client/profile/TenantProfile.java
shoatman: In the case of B2C i'd consider falling back to home object id

msal/src/main/java/com/microsoft/identity/client/profile/IAccount.java
```diff
@@ -0,0 +1,46 @@
...
+
+public interface IAccount {
+
+    /**
+     * Gets this IAccount's home tenant OID.
+     *
+     * @return The oid of this IAccount.
+     */
+    @NonNull
+    String getId();
```
shoatman: I'm thinking this should still be homeAccountId

iambmelt:
Closing, superseded by #629

PR #573, opened 2019-04-13
IMicrosoftAuthService.getAccounts() and removeAccount()

heidijinxujia:
- Add the fix for acquireToken silent with broker
- Add the implementation for getAccounts on MSAL side
- Add the implementation for removeAccount on MSAL side
- Update the msal test app

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
@@ -46,7 +46,7 @@ public class AccountAdapter {
      * @return A representation of the supplied Account, as an IAccount.
      */
     @NonNull
-    static IAccount adapt(@NonNull final IAccountRecord accountIn) {
+    public static IAccount adapt(@NonNull final IAccountRecord accountIn) {
         final String methodName = ":adapt";
         final com.microsoft.identity.client.Account accountOut
                 = new com.microsoft.identity.client.Account();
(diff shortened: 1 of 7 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -525,22 +618,47 @@ public boolean removeAccount(@Nullable final IAccount account) {
...
                         account.getHomeAccountIdentifier().getIdentifier(),
                         realm
                 ).isEmpty();
+
+        if (MSALControllerFactory
+                .brokerEligible(mPublicClientConfiguration.getAppContext(),
+                mPublicClientConfiguration.getDefaultAuthority(),
+                mPublicClientConfiguration)) {
+            //Remove the account from Broker
+            sBackgroundExecutor.submit(new Runnable() {
```
kreedula: move the Background operation logic to `BrokerMsalController().removeBrokerAccount` as discussed

heidijinxujia:
@kreedula @iambmelt  Please let me know if there is any other comment. Thanks.

heidijinxujia:
Thank you, Krishna. :)

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
```diff
@@ -46,7 +46,7 @@
      * @return A representation of the supplied Account, as an IAccount.
      */
     @NonNull
-    static IAccount adapt(@NonNull final IAccountRecord accountIn) {
+    public static IAccount adapt(@NonNull final IAccountRecord accountIn) {
```
iambmelt: If we're going to make this public, we need to move this class to an `internal` namespace

PR #580, opened 2019-04-19
Msa changes

kreedula:
Saving/Setting single sign on state for  MSA account locally using refresh token and account

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
@@ -239,6 +239,7 @@ public class BrokerMsalController extends BaseController {
                 Logger.errorPII(TAG + methodName, "Exception while creating Idtoken or ClientInfo," +
                         " cannot save MSA account tokens", e
                 );
+                throw new ClientException(ErrorStrings.INVALID_JWT, e.getMessage(), e);
             }
         }
 
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
@@ -204,7 +204,9 @@ public class BrokerMsalController extends BaseController {
      * @param msalOAuth2TokenCache
      */
     private void saveMsaAccountToCache(@NonNull final Bundle resultBundle,
-                                       @NonNull final MsalOAuth2TokenCache msalOAuth2TokenCache) {
+                                       @NonNull final MsalOAuth2TokenCache msalOAuth2TokenCache) throws ClientException {
+        final String methodName = ":saveMsaAccountToCache";
+
         final BrokerResult brokerResult = (BrokerResult) resultBundle.getSerializable(
                 AuthenticationConstants.Broker.BROKER_RESULT_V2
         );
@@ -212,7 +214,7 @@ public class BrokerMsalController extends BaseController {
         if (resultBundle.getBoolean(AuthenticationConstants.Broker.BROKER_REQUEST_V2_SUCCESS)
                 && brokerResult != null &&
                 AzureActiveDirectoryAudience.MSA_MEGA_TENANT_ID.equalsIgnoreCase(brokerResult.getTenantId())) {
-            Logger.info(TAG, "Result returned for MSA Account, saving to cache");
+            Logger.info(TAG + methodName, "Result returned for MSA Account, saving to cache");
 
             try {
                 final ClientInfo clientInfo = new ClientInfo(brokerResult.getClientInfo());
@@ -234,7 +236,7 @@ public class BrokerMsalController extends BaseController {
                 msalOAuth2TokenCache.setSingleSignOnState(microsoftStsAccount, microsoftRefreshToken);
 
             } catch (ServiceException e) {
-                Logger.errorPII(TAG, "Exception while creating Idtoken or ClientInfo," +
+                Logger.errorPII(TAG + methodName, "Exception while creating Idtoken or ClientInfo," +
                         " cannot save MSA account tokens", e
                 );
             }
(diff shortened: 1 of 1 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -188,4 +197,49 @@ private AcquireTokenResult getAcquireTokenResult(final Bundle resultBundle) thro
...
+                        clientInfo
+                );
+                microsoftStsAccount.setEnvironment(brokerResult.getEnvironment());
+
+                final MicrosoftRefreshToken microsoftRefreshToken = new MicrosoftRefreshToken(
+                        brokerResult.getRefreshToken(),
+                        clientInfo,
+                        brokerResult.getScope(),
+                        brokerResult.getClientId(),
+                        false, // TODO : is family refresh token valid here ? Should we get this info from Broker
```
heidijinxujia: You can call `!StringExtensions.isNullOrBlank(brokerResult.getRefreshToken().getFamilyId())` here.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -188,4 +197,49 @@ private AcquireTokenResult getAcquireTokenResult(final Bundle resultBundle) thro
...
+                        brokerResult.getEnvironment(),
+                        null
+                );
+
+                msalOAuth2TokenCache.setSingleSignOnState(microsoftStsAccount, microsoftRefreshToken);
+
+            } catch (Exception e) {
```
heidijinxujia: Could it be more specific on the kinds of exceptions here?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
iambmelt: FRT _does_ apply to MSA accounts. We should expose this information such that TSL can integrate in the future

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -188,4 +197,49 @@ private AcquireTokenResult getAcquireTokenResult(final Bundle resultBundle) thro
...
+                Logger.error(TAG, "Exception while creating Idtoken or ClientInfo," +
```
iambmelt: Let's also do a `Logger.errorPII()` on the `Exception` (or subtype), if it's thrown- the likelihood of the `ServiceException` being thrown by the `IDToken` constructor is low, given that the broker would have had to parse it to make a MSA/non-MSA determination in the first place

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -188,4 +198,48 @@ private AcquireTokenResult getAcquireTokenResult(final Bundle resultBundle) thro
...
+                final MicrosoftRefreshToken microsoftRefreshToken = new MicrosoftRefreshToken(
+                        brokerResult.getRefreshToken(),
+                        clientInfo,
+                        brokerResult.getScope(),
+                        brokerResult.getClientId(),
+                        brokerResult.getEnvironment(),
+                        brokerResult.getFamilyId()
+                );
+
+                msalOAuth2TokenCache.setSingleSignOnState(microsoftStsAccount, microsoftRefreshToken);
```
iambmelt: May be a good idea to log the result value here - should be a true/false on success. Could log a warning on non-success.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -81,21 +89,23 @@ public AcquireTokenResult acquireToken(AcquireTokenOperationParameters parameter
         //Wait to be notified of the result being returned... we could add a timeout here if we want to
         final Bundle resultBundle = mBrokerResultFuture.get();
 
-        return getAcquireTokenResult(resultBundle);
+        // For MSA Accounts Broker doesn't save the accounts, instead it just passes the result along,
+        // MSAL needs to save this account locally for future token calls.
+        saveMsaAccountToCache(resultBundle, (MsalOAuth2TokenCache)parameters.getTokenCache());
 
+        return getAcquireTokenResult(resultBundle);
```
iambmelt: If the call to `save()` returns `false` for any reason, we should think what smart thing we can do here to signal there was an issue

iambmelt: Might be worth adding a return value to `saveMsaToCache()`

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
kreedula: Throwing back the exception to user

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -188,4 +198,50 @@ private AcquireTokenResult getAcquireTokenResult(final Bundle resultBundle) thro
...
+                        clientInfo,
+                        brokerResult.getScope(),
+                        brokerResult.getClientId(),
+                        brokerResult.getEnvironment(),
+                        brokerResult.getFamilyId()
+                );
+
+                msalOAuth2TokenCache.setSingleSignOnState(microsoftStsAccount, microsoftRefreshToken);
+
+            } catch (ServiceException e) {
```
iambmelt: Does it make sense to `throw` the `ClientException` but not the `ServiceException`? Since this shouldn't happen I think we can make this judgment call based on how much adding a checked exception here complicates the code

iambmelt: If you opt not to throw the `ServiceException` to the caller, let's drop an inline comment to clarify why

kreedula: Throwing `ClientException` as discussed

PR #586, opened 2019-04-23
Merge dev into release/0.3.1

heidijinxujia:


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
```diff
@@ -169,7 +169,7 @@ protected void onCreate(Bundle savedInstanceState) {
         }
 
         if (mApplication == null) {
-            mApplication = new PublicClientApplication(this.getApplicationContext(), R.raw.msal_config);
+            mApplication = new PublicClientApplication(this.getApplicationContext(), R.raw.msal_arlington_config);
```
iambmelt: For the purposes of this PR, this doesn't matter. May want to set back to the default config so we're not surprised during testing

heidijinxujia: @shoatman Could you have a look? Thanks.

iambmelt: We can make this change directly on either branch - my call-out was an FYI

PR #589, opened 2019-04-26
Refactor for removeLocalAccount()

heidijinxujia:
**Problem**
When configuring the app with ```"authorization_user_agent": "WEBVIEW"```
Although the user calling -
```public void removeAccount(@Nullable final IAccount account, final AccountsRemovedCallback callback)```

Ref: #590 

and getting onAccountsRemoved = true, when the user tries to resign in, MSAL automatically signs me in, without prompting for password.

**Diagnose**
This is because the app embedded webview's cookie is still valid and not cleaned even the local MSAL cache is removed.

**Repro steps**
1. acquireToken with user agent as WEBVIEW
2. entering the username and credentials
3. get the token back 
4. type in the login hint used in Step 2 in the test app
5. Click "remove account"
6. Toast msg shows "the account is deleted successfully"
7. acquireToken with user agent as WEBVIEW with loginhint
8. [not expected] get the token back without entering credentials.

**Work items**
1. Add the cache clean up for embedded webview when calling removeLocalAccount()
2. Refactor the removeAccounts()

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
@@ -140,7 +140,7 @@ public class AccountAdapter {
     }
 
     @Nullable
-    public static String getRealm(@NonNull IAccount account) {
+    static String getRealm(@NonNull IAccount account) {
         String realm = null;
 
         if (null != account.getAccountIdentifier() // This is an AAD account w/ tenant info
--- a/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
@@ -138,4 +138,17 @@ public class AccountAdapter {
 
         return accountToReturn;
     }
+
+    @Nullable
+    public static String getRealm(@NonNull IAccount account) {
+        String realm = null;
+
+        if (null != account.getAccountIdentifier() // This is an AAD account w/ tenant info
+                && account.getAccountIdentifier() instanceof AzureActiveDirectoryAccountIdentifier) {
+            final AzureActiveDirectoryAccountIdentifier identifier = (AzureActiveDirectoryAccountIdentifier) account.getAccountIdentifier();
+            realm = identifier.getTenantIdentifier();
+        }
+
+        return realm;
+    }
 }
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -43,6 +43,7 @@ import com.microsoft.identity.client.exception.MsalException;
 import com.microsoft.identity.client.internal.MsalUtils;
 import com.microsoft.identity.client.internal.configuration.LogLevelDeserializer;
 import com.microsoft.identity.client.internal.controllers.BrokerMsalController;
+import com.microsoft.identity.client.internal.controllers.LocalMSALController;
 import com.microsoft.identity.client.internal.controllers.MSALControllerFactory;
 import com.microsoft.identity.client.internal.controllers.MsalExceptionAdapter;
 import com.microsoft.identity.client.internal.controllers.OperationParametersAdapter;
@@ -620,20 +621,6 @@ public final class PublicClientApplication {
             callback.onAccountsRemoved(false);
         }
 
-        // FEATURE SWITCH: Set to false to allow deleting Accounts in a tenant-specific way.
-        final boolean deleteAccountsInAllTenants = true;
-
-        final String realm = deleteAccountsInAllTenants ? null : getRealm(account);
-
-        final boolean localRemoveAccountSuccess = !mPublicClientConfiguration
-                .getOAuth2TokenCache()
-                .removeAccount(
-                        account.getEnvironment(),
-                        mPublicClientConfiguration.getClientId(),
-                        account.getHomeAccountIdentifier().getIdentifier(),
-                        realm
-                ).isEmpty();
-
         if (MSALControllerFactory.brokerEligible(
                 mPublicClientConfiguration.getAppContext(),
                 mPublicClientConfiguration.getDefaultAuthority(),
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java [hunk 1]
```diff
@@ -265,4 +273,39 @@ public AcquireTokenResult acquireTokenSilent(
...
+                .removeAccount(
+                        account.getEnvironment(),
+                        configuration.getClientId(),
+                        account.getHomeAccountIdentifier().getIdentifier(),
+                        realm
+                ).isEmpty();
+
+        // Step 2.
+        // Clear cookies from embedded webView.
+        final CookieManager cookieManager = CookieManager.getInstance();
```
rpdome: Given that I also have this code change in Broker (FLW global sign out), should we extract this function to common?

heidijinxujia: FLW is a different scenario. FLW is restricted with one account with a device level scenario. Whereas, the normal local MSAL follow is mutli-accounts with a application level supported. I am not inclined to combine the removeAccounts() together.

rpdome: oh no. I'm referring to this 'clear cookie' functionality specifically.

heidijinxujia: Could you kindly link the code here?

iambmelt: @heidijinxujia @rpdome's work is here https://github.com/AzureAD/ad-accounts-for-android/pull/951/files

As Dome calls out, looks like we need a shared function here

heidijinxujia:
Synced with Shane offline this morning. Actually, in the multi-accounts scenarios, "removeAllCookies()" is overkill and may cause the lose of other account's SSO. We do not clean all webview cookies by default. "removeAllCookies()" can be applied for FLW only. Instead, the application can optional choose to handle the webview cleaning cookie by them own.  

For future enhancement, there are two ways to sign-out the user from webview or browser in general.
1. We can call `CookieManager.getCookie(url)` and  `CookieManager.getCookie(url)` to modify the cookie. However, url string is required and currently we are not clear about the EVO sub-domains.
Quoted from Shane for future visit
`CookieManager will let us enumerate the cookies based on a url (I haven't tried, but i'm guessing this means we have to be aware of EVO sub-domains... since i see that evo emits cookies prefixed with a ".").  We could then update each of those to expire them.    B2C has documented their cookies, not sure if this list is definitive...https://docs.microsoft.com/en-us/azure/active-directory-b2c/cookie-definitions `

2. Change the removeAccount() call from background to foreground by calling the logout endpoint. Here is the [ref](https://docs.microsoft.com/en-us/azure/active-directory/develop/v1-protocols-openid-connect-code#send-a-sign-out-request). The user needs to click the upn on the UI to logout. However, the logout redirect seems only works for web requests. Server change (support custom scheme redirect url with client_id) is required to apply this solution. Otherwise, the UI won't be redirect to the calling app.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
rpdome: I moved the code to common https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/491

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
```diff
@@ -138,4 +138,17 @@ static AccountRecord getAccountInternal(@NonNull final String clientId,
 
         return accountToReturn;
     }
+
+    @Nullable
+    public static String getRealm(@NonNull IAccount account) {
```
iambmelt: I don't think we should make this public API - can we move this class down into the `internal` pkg?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -43,6 +43,7 @@
 import com.microsoft.identity.client.internal.MsalUtils;
 import com.microsoft.identity.client.internal.configuration.LogLevelDeserializer;
 import com.microsoft.identity.client.internal.controllers.BrokerMsalController;
+import com.microsoft.identity.client.internal.controllers.LocalMSALController;
 import com.microsoft.identity.client.internal.controllers.MSALControllerFactory;
```
Om83: Not part of your change but since MSAL is an acronym, would be good to change all MsalController to MSALController.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
```diff
@@ -247,4 +256,24 @@ public AcquireTokenResult acquireTokenSilent(
 
         return acquireTokenSilentResult;
     }
+
+    public void removeLocalAccount(@Nullable final IAccount account,
+                                   @NonNull final PublicClientApplicationConfiguration configuration,
+                                   @NonNull final PublicClientApplication.AccountsRemovedCallback callback) {
+        // FEATURE SWITCH: Set to false to allow deleting Accounts in a tenant-specific way.
+        final boolean deleteAccountsInAllTenants = true;
```
Om83: This looks little confusing. We are trying to remove one account and making use of variable with deleteAccountsInAllTenants.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: As best we can, we try to adhere to the [Android Style Guide](https://source.android.com/setup/contribute/code-style#treat-acronyms-as-words) which advocates treating acronyms as words. As we have a lot of contributors there's been some inconsistency on this point; however, we ought to stylize as `MsalClassName` rather than `MSALClassName`

msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
iambmelt: In this case, we wish to remove both home and any guest accounts so we _are_ removing multiple accounts.

Because a user principal maps to multiple `AccountRecords` in guest scenarios, we (internally) support deleting `IAccounts` in multiple ways:
- Tenant-specific
   * Ex: a user is logged in Contoso and Fabrikam, but only wishes to remove the `AccountRecord` for Contoso
- Tenant-agnostic
    * Ex: a user is logged in Contoso and Fabrikam and wishes to remove both records

To make this more clear, let's rename this variable to something like `deleteHomeAndGuestAccounts` @heidijinxujia

heidijinxujia:
@iambmelt  Could have have a look at this one? Just refactor.

heidijinxujia:
@iambmelt  Thank you for reviews. Comments addressed. :)

msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
```diff
@@ -23,9 +23,18 @@
 package com.microsoft.identity.client.internal.controllers;
 
 import android.content.Intent;
+import android.os.Build;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.text.TextUtils;
-
+import android.webkit.CookieManager;
```
iambmelt: Handful of unused imports in this file

PR #591, opened 2019-04-29
Introduce PublicClientApplication.Create and related interfaces

rpdome:
This is part of the End-my-Shift functionality. 

- Replace PublicClientApplication's public constructor with PublicClientApplication.Create()

- Introduce IPublicClientApplication, ISingleAccountPublicClientApplication and IMultipleAccountPublicClientApplication
    - Also move all of the API descriptions to these interfaces.

- Update test app to accompany the new interface.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -605,7 +605,7 @@ public class PublicClientApplication implements IPublicClientApplication, IMulti
     }
 
     @Override
-    public void removeAccount(@Nullable final IAccount account, final AccountsRemovedCallback callback) {
+    public void removeAccount(@Nullable final IAccount account, final AccountRemovedListener callback) {
         ApiDispatcher.initializeDiagnosticContext();
         if (null == account
                 || null == account.getHomeAccountIdentifier()
@@ -615,7 +615,7 @@ public class PublicClientApplication implements IPublicClientApplication, IMulti
                     "Requisite IAccount or IAccount fields were null. Insufficient criteria to remove IAccount."
             );
 
-            callback.onAccountsRemoved(false);
+            callback.onAccountRemoved(false);
         }
 
         // FEATURE SWITCH: Set to false to allow deleting Accounts in a tenant-specific way.
@@ -644,7 +644,7 @@ public class PublicClientApplication implements IPublicClientApplication, IMulti
                     callback
             );
         } else {
-            callback.onAccountsRemoved(localRemoveAccountSuccess);
+            callback.onAccountRemoved(localRemoveAccountSuccess);
         }
     }
 
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -232,11 +232,9 @@ public class PublicClientApplication implements IPublicClientApplication, IMulti
         if (MsalUtils.isEmpty(clientId)) {
             throw new IllegalArgumentException("client id is empty or null");
         }
-        new BrokerMsalController().getBrokerAccountMode(context, new BrokerAccountModeCallback()
-        {
+        new BrokerMsalController().getBrokerAccountMode(context, new BrokerAccountModeCallback() {
             @Override
-            public void onGetMode(String mode)
-            {
+            public void onGetMode(String mode) {
                 if (AuthenticationConstants.Broker.BROKER_ACCOUNT_MODE_SINGLE_ACCOUNT.equalsIgnoreCase(mode)) {
                     // TODO: return SingleAccountPublicClientApplication
                     listener.onCreated(new PublicClientApplication(context, clientId));
@@ -279,11 +277,9 @@ public class PublicClientApplication implements IPublicClientApplication, IMulti
             throw new IllegalArgumentException("authority is null");
         }
 
-        new BrokerMsalController().getBrokerAccountMode(context, new BrokerAccountModeCallback()
-        {
+        new BrokerMsalController().getBrokerAccountMode(context, new BrokerAccountModeCallback() {
             @Override
-            public void onGetMode(String mode)
-            {
+            public void onGetMode(String mode) {
                 if (AuthenticationConstants.Broker.BROKER_ACCOUNT_MODE_SINGLE_ACCOUNT.equalsIgnoreCase(mode)) {
                     // TODO: return SingleAccountPublicClientApplication
                     listener.onCreated(new PublicClientApplication(context, clientId, authority));
@@ -298,11 +294,9 @@ public class PublicClientApplication implements IPublicClientApplication, IMulti
     private static void create(@NonNull final Context context,
                                final PublicClientApplicationConfiguration developerConfig,
                                @NonNull final ApplicationCreatedListener listener){
-        new BrokerMsalController().getBrokerAccountMode(context, new BrokerAccountModeCallback()
-        {
+        new BrokerMsalController().getBrokerAccountMode(context, new BrokerAccountModeCallback() {
             @Override
-            public void onGetMode(String mode)
-            {
+            public void onGetMode(String mode) {
                 if (AuthenticationConstants.Broker.BROKER_ACCOUNT_MODE_SINGLE_ACCOUNT.equalsIgnoreCase(mode)) {
                     // TODO: return SingleAccountPublicClientApplication
                     listener.onCreated(new PublicClientApplication(context, developerConfig));
@@ -403,6 +397,7 @@ public class PublicClientApplication implements IPublicClientApplication, IMulti
         return mPublicClientConfiguration;
     }
 
+    @Override
     public void getAccounts(@NonNull final AccountsLoadedCallback callback) {
         ApiDispatcher.initializeDiagnosticContext();
         final String methodName = ":getAccounts";
@@ -567,6 +562,7 @@ public class PublicClientApplication implements IPublicClientApplication, IMulti
         return accountsInCache;
     }
 
+    @Override
     @Nullable
     public IAccount getAccount(@NonNull final String homeAccountIdentifier,
                                @Nullable final String authority) {
@@ -608,6 +604,7 @@ public class PublicClientApplication implements IPublicClientApplication, IMulti
         return null == accountToReturn ? null : AccountAdapter.adapt(accountToReturn);
     }
 
+    @Override
     public void removeAccount(@Nullable final IAccount account, final AccountsRemovedCallback callback) {
         ApiDispatcher.initializeDiagnosticContext();
         if (null == account
@@ -664,12 +661,14 @@ public class PublicClientApplication implements IPublicClientApplication, IMulti
         return realm;
     }
 
+    @Override
     public void handleInteractiveRequestRedirect(final int requestCode,
                                                  final int resultCode,
                                                  @NonNull final Intent data) {
         ApiDispatcher.completeInteractive(requestCode, resultCode, data);
     }
 
+    @Override
     public void acquireToken(@NonNull final Activity activity,
                              @NonNull final String[] scopes,
                              @NonNull final AuthenticationCallback callback) {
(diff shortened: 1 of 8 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -257,24 +220,36 @@ public PublicClientApplication(@NonNull final Context context, final File config
...
         if (MsalUtils.isEmpty(clientId)) {
             throw new IllegalArgumentException("client id is empty or null");
         }
-
-        setupConfiguration(context);
-        mPublicClientConfiguration.mClientId = clientId;
-        initializeApplication();
-        Authority.addKnownAuthorities(mPublicClientConfiguration.getAuthorities());
+        new BrokerMsalController().getBrokerAccountMode(context, new BrokerAccountModeCallback()
+        {
```
iambmelt: Minor nit on brace formatting

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -257,24 +220,36 @@ public PublicClientApplication(@NonNull final Context context, final File config
...
+            @Override
+            public void onGetMode(String mode)
+            {
```
iambmelt: Minor nit on brace formatting

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -285,17 +260,83 @@ public PublicClientApplication(@NonNull final Context context, @NonNull final St
...
+        if (MsalUtils.isEmpty(clientId)) {
+            throw new IllegalArgumentException("client id is empty or null");
         }
 
+        if (authority == null) {
+            throw new IllegalArgumentException("authority is null");
+        }
+
+        new BrokerMsalController().getBrokerAccountMode(context, new BrokerAccountModeCallback()
+        {
```
iambmelt: Minor nit on brace formatting

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -285,17 +260,83 @@ public PublicClientApplication(@NonNull final Context context, @NonNull final St
...
+                }
+            }
+        });
+    }
+
+    private static void create(@NonNull final Context context,
+                               final PublicClientApplicationConfiguration developerConfig,
+                               @NonNull final ApplicationCreatedListener listener){
+        new BrokerMsalController().getBrokerAccountMode(context, new BrokerAccountModeCallback()
+        {
```
iambmelt: Minor nit on brace formatting

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -145,53 +145,16 @@
  * </p>
  * </p>
  */
-public final class PublicClientApplication {
+public class PublicClientApplication implements IPublicClientApplication, IMultipleAccountPublicClientApplication {
```
iambmelt: Long-term I think we would benefit from splitting this into 4 classes - this way implementations are thematically grouped and we minimize the risk of developers internally calling the wrong functions. We also then prohibit the developer from casting the result and doing weird things

- 1 Factory class to produce subclasses
    * Factory should have an `interface` and a default implementation
- 1 abstract superclass with common methods for the various PCA subclasses
- 1 interface + 1 subclass for single-user
- 1 interface + 1 subclass for multi-user

rpdome: Based on the design doc, we were originally planning to split them into 3 classes.

- IMultipleAccountPublicClientApplication will be splitted into another class in the next CR. (Keep it here temporarily so that the test app doesn't break).
- ISingleAccountPublicClientApplication will also be another class.
- PublicClientApplication & the Factory is still in the same class

I agree we could/should split them into separate cards. Could do that in the upcoming PRs.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -285,17 +260,83 @@ public PublicClientApplication(@NonNull final Context context, @NonNull final St
...
+    }
+
+    private static void create(@NonNull final Context context,
+                               final PublicClientApplicationConfiguration developerConfig,
+                               @NonNull final ApplicationCreatedListener listener){
+        new BrokerMsalController().getBrokerAccountMode(context, new BrokerAccountModeCallback()
+        {
+            @Override
+            public void onGetMode(String mode)
+            {
```
iambmelt: Nit: bracing

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -547,13 +567,6 @@ public void run() {
...
 
-    /**
-     * Returns the IAccount object matching the supplied home_account_id.
-     *
-     * @param homeAccountIdentifier The home_account_id of the sought IAccount.
-     * @param authority             The authority of the sought IAccount.
-     * @return The IAccount stored in the cache or null, if no such matching entry exists.
-     */
     @Nullable
     public IAccount getAccount(@NonNull final String homeAccountIdentifier,
```
iambmelt: Can we add the `@Override` annotations to these methods?

iambmelt: Will just call out once, but should apply to a few methods

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -199,6 +202,57 @@ public AcquireTokenResult acquireTokenSilent(AcquireTokenSilentOperationParamete
...
+    /**
+     * Get MSAL PublicClientApplication mode from Broker.
+     * */
+    public void getBrokerAccountMode(final Context appContext,
+                                     final PublicClientApplication.BrokerAccountModeCallback callback) {
+
+        final String methodName = ":getBrokerAccountMode";
+        final Handler handler = new Handler(Looper.getMainLooper());
+
+        new Thread(new Runnable() {
```
iambmelt: Rather than make a one-off thread, can we add an `Executor` to this class and submit the job?

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
```diff
@@ -238,12 +249,15 @@ public void onAcquireTokenClicked(final AcquireTokenFragment.RequestOptions requ
...
 
     public void onRemoveUserClicked(final String username) {
-        mApplication.getAccounts(new PublicClientApplication.AccountsLoadedCallback() {
-            @Override
-            public void onAccountsLoaded(List<IAccount> accountsToRemove) {
-                for (final IAccount accountToRemove : accountsToRemove) {
-                    if (TextUtils.isEmpty(username) || accountToRemove.getUsername().equalsIgnoreCase(username.trim())) {
-                        mApplication.removeAccount(
+        if (mApplication instanceof IMultipleAccountPublicClientApplication){
+            final IMultipleAccountPublicClientApplication application = (IMultipleAccountPublicClientApplication)(mApplication);
```
iambmelt: This feels clunky

iambmelt:
I don't think that the single Factory-method API works... the return types here, save for a few methods, are pretty different. IMO having to explicitly test the dynamic return type of the Factory creates _more work_ than calling a separate constructor.

I think there's a couple of things we can do to mitigate...

One option: add a return-type _parameter_ to the factory method where `onCreated` pre-casts the result to the generic type
```java
class Foo {
    ISingleAccountPublicClientApplication mPca;

    PublicClientApplication.<ISingleAccountPublicClientApplication> create(this.getApplicationContext(),
                R.raw.msal_config,
                new PublicClientApplication.ApplicationCreatedListener<ISingleAccountPublicClientApplication>() {
                    @Override
                    public void onCreated(ISingleAccountPublicClientApplication application) {
                        mApplication = application;
                    }
                });
}
```

Another option: separate method for construction

rpdome:
The main reason the factory methods were introduced is because the type of the returned object is based on internal state of the Broker. This means that the calling application **will not** know what type it will be getting - therefore Type cannot be set in advance.

Once FLW functionality is shipped, I'd expect the consuming applications to write a wrapper around MSAL and implements something along the line of the following.

- If the application doesn't support FLW/Single account mode, then it should adjust the app state/UI accordingly when it fails to cast the resulted application object to IMultipleAccountPublicClientApplication. 

- If the application supports FLW/Single account mode, then the wrapper will have to support two types of interfaces, and it will have to do the casting and invoke the right behavior accordingly.

Thanks for the feedback. Let's sync up on this.

iambmelt:
> The main reason the factory methods were introduced is because the type of the returned object is based on internal state of the Broker. This means that the calling application **will not** know what type it will be getting - therefore Type cannot be set in advance.
> 
> Once FLW functionality is shipped, I'd expect the consuming applications to write a wrapper around MSAL and implements something along the line of the following.
> 
> * If the application doesn't support FLW/Single account mode, then it should adjust the app state/UI accordingly when it fails to cast the resulted application object to IMultipleAccountPublicClientApplication.
> * If the application supports FLW/Single account mode, then the wrapper will have to support two types of interfaces, and it will have to do the casting and invoke the right behavior accordingly.
> 
> Thanks for the feedback. Let's sync up on this.

Thanks Dome. Makes sense to me, appreciate the clarification - consider us synced :)

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
rpdome: I refactored the test app (basically make a wrapper around MSAL. Will send out in a separate PR. Hopefully it helps.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -324,65 +359,45 @@ private void initializeApplication() {
...
     /**
-     * Asynchronously returns a List of {@link IAccount} objects for which this application has RefreshTokens.
+     * Returns the PublicClientConfiguration for this instance of PublicClientApplication
+     * Configuration is based on the defaults established for MSAl and can be overridden by creating the
+     * PublicClientApplication using {@link PublicClientApplication#create(Context, int, ApplicationCreatedListener)}
      *
-     * @param callback The callback to notify once this action has finished.
+     * @return
      */
+    public PublicClientApplicationConfiguration getConfiguration() {
```
iambmelt: Should this move up to the interface?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -199,6 +202,57 @@ public AcquireTokenResult acquireTokenSilent(AcquireTokenSilentOperationParamete
         }
     }
 
+    /**
+     * Get MSAL PublicClientApplication mode from Broker.
+     * */
+    public void getBrokerAccountMode(final Context appContext,
```
iambmelt: Same question here - should this be on the interface?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
rpdome: I'm slightly inclined to the 'no' side. The reason is that I want the interface to only contains API that are meant to be used by app developers. Not strong opinion tho.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
rpdome: Same answer as above.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: I was thrown off by the `public` access modifier here - can we remove move this down to default? I think this is only called by tests which should be in the same pkg

rpdome: will do.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
rpdome: This one is in another class.

iambmelt:
Taking another look now, thanks @rpdome

PR #592, opened 2019-04-29
[Blocked] Add PCA.getAccount(oid) API to retrieve account by object id

heidijinxujia:
**Problem**
Currently MSAL getAccount expects a homeAccountIdentifier.  I think we should just an identifier.  This will allow for a simpler migration for customers who've already persisted the object id of the user in their system.

The oid is the obejct id, which should be the same value of local account id. For the accounts with same home accout id, the local account ids differ with tenants. 

**Solution**
Add the API to retrieve the account with oid
Remove the API to getAccount() with homeAccountId, because the aim here is to restrict the retrieving with oid only.

Also fixed the redirect uri bug founding during the E2E testing
- Update the redirectUri generation with the new format

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -567,10 +567,20 @@ public final class PublicClientApplication {
         return accountsInCache;
     }
 
-    public void getAccount(@NonNull final String oid, @NonNull final AccountLoadedCallback callback) {
+    /**
+     * Returns the IAccount object matching the supplied object id.
+     *
+     * @param oid The object id of the sought IAccount.
+     * @param callback AccountLoadedCallback
+     */
+    public void getAccount(@NonNull final String oid,
+                           @NonNull final AccountLoadedCallback callback) {
         final String methodName = ":getAccount";
+
+        ApiDispatcher.initializeDiagnosticContext();
+
         if (StringUtil.isEmpty(oid)) {
-            throw new IllegalArgumentException("Illegal oid to get account.");
+            throw new IllegalArgumentException("Illegal oid to get the sought account.");
         }
 
         if (MSALControllerFactory.brokerEligible(
@@ -592,7 +602,8 @@ public final class PublicClientApplication {
         }
     }
 
-    private void getBrokerAndLocalAccount(@NonNull final String oid, @NonNull final AccountLoadedCallback callback) {
+    private void getBrokerAndLocalAccount(@NonNull final String oid,
+                                          @NonNull final AccountLoadedCallback callback) {
         final String methodName = ":getBrokerAndLocalAccount";
 
         final Handler handler;
@@ -620,6 +631,7 @@ public final class PublicClientApplication {
                             handler.post(new Runnable() {
                                 @Override
                                 public void run() {
+                                    //Return null if no account found.
                                     callback.onAccountLoaded(null);
                                 }
                             });
@@ -646,7 +658,8 @@ public final class PublicClientApplication {
     }
 
 
-    private void getLocalAccount(@NonNull final String oid, @NonNull final AccountLoadedCallback callback) {
+    private void getLocalAccount(@NonNull final String oid,
+                                 @NonNull final AccountLoadedCallback callback) {
         final String methodName = ":getLocalAccount";
         List<AccountRecord> localAccountList = getLocalAccounts();
         if (localAccountList == null || localAccountList.isEmpty()) {
@@ -675,54 +688,6 @@ public final class PublicClientApplication {
                 "No account found in the local cache."
         );
         callback.onAccountLoaded(null);
-    }
-
-    /**
-     * Returns the IAccount object matching the supplied home_account_id.
-     *
-     * @param homeAccountIdentifier The home_account_id of the sought IAccount.
-     * @param authority             The authority of the sought IAccount.
-     * @return The IAccount stored in the cache or null, if no such matching entry exists.
-     */
-    @Nullable
-    public IAccount getAccount(@NonNull final String homeAccountIdentifier,
-                               @Nullable final String authority) {
-        final String methodName = ":getAccount";
-
-        ApiDispatcher.initializeDiagnosticContext();
-
-        String realm = StringUtil.getTenantInfo(homeAccountIdentifier).second;
-
-        Authority authorityObj = Authority.getAuthorityFromAuthorityUrl(authority);
-
-        if (authorityObj instanceof AzureActiveDirectoryAuthority) {
-            final AzureActiveDirectoryAuthority aadAuthority = (AzureActiveDirectoryAuthority) authorityObj;
-            final AzureActiveDirectoryAudience audience = aadAuthority.getAudience();
-            realm = audience.getTenantId();
-        } else {
-            com.microsoft.identity.common.internal.logging.Logger.warn(
-                    TAG + methodName,
-                    "Provided authority was not AAD - defaulting to parsed home_account_id"
-            );
-        }
-
-        AccountRecord accountToReturn = null;
-
-        if (null != realm) {
-            accountToReturn = AccountAdapter.getAccountInternal(
-                    mPublicClientConfiguration.getClientId(),
-                    mPublicClientConfiguration.getOAuth2TokenCache(),
-                    homeAccountIdentifier,
-                    realm
-            );
-        } else {
-            com.microsoft.identity.common.internal.logging.Logger.warn(
-                    TAG + methodName,
-                    "Realm could not be resolved. Returning null."
-            );
-        }
-
-        return null == accountToReturn ? null : AccountAdapter.adapt(accountToReturn);
     }
 
     /**
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -56,6 +56,7 @@ import com.microsoft.identity.common.internal.authorities.AuthorityDeserializer;
 import com.microsoft.identity.common.internal.authorities.AzureActiveDirectoryAudience;
 import com.microsoft.identity.common.internal.authorities.AzureActiveDirectoryAudienceDeserializer;
 import com.microsoft.identity.common.internal.authorities.AzureActiveDirectoryAuthority;
+import com.microsoft.identity.common.internal.broker.BrokerValidator;
 import com.microsoft.identity.common.internal.cache.CacheKeyValueDelegate;
 import com.microsoft.identity.common.internal.cache.IAccountCredentialCache;
 import com.microsoft.identity.common.internal.cache.ICacheKeyValueDelegate;
@@ -361,6 +362,19 @@ public final class PublicClientApplication {
     }
 
     /**
+     * Listener callback for asynchronous finding of IAccount matching the given oid.
+     */
+    public interface AccountLoadedCallback {
+
+        /**
+         * Called once Account have been found from the cache.
+         *
+         * @param account The account in the cache.
+         */
+        void onAccountLoaded(IAccount account);
+    }
+
+    /**
      * Listener callback for asynchronous loading of msal IAccount accounts.
      */
     public interface AccountsRemovedCallback {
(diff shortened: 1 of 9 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -360,6 +361,19 @@ public PublicClientApplicationConfiguration getConfiguration() {
         void onAccountsLoaded(List<IAccount> accounts);
     }
 
+    /**
+     * Listener callback for asynchronous finding of IAccount matching the given oid.
+     */
+    public interface AccountLoadedCallback {
```
iambmelt: Can we re-tool the existing interface to accommodate returning 1 _or_ many accounts? I think we can keep the method name plural without confusing ppl

Something like
```java
public interface AccountsLoadedCallback<T> {
    void onAccountsLoaded(T account);
}
```

And then our various implementations will pass-in the right arg depending on what's needed
```java
new AccountsLoadedCallback<Account>() { // Just need one
    @Override
    public void onAccountsLoaded(Account account) {
        // Do something useful            
    }
}

new AccountsLoadedCallback<List<Account>>() { // Need multiple
    @Override
    public void onAccountsLoaded(List<Account> account) {
        // Something else! 
    }
}
```

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -554,51 +568,126 @@ public void run() {
...
+                                    && accountRecord.getLocalAccountId().equalsIgnoreCase(oid)) {
+                                com.microsoft.identity.common.internal.logging.Logger.info(
+                                        TAG + methodName,
+                                        "Found the matched account from broker cache."
+                                );
+
+                                handler.post(new Runnable() {
+                                    @Override
+                                    public void run() {
+                                        callback.onAccountLoaded(AccountAdapter.adapt(accountRecord));
```
iambmelt: Certainly we can't be meaning to call this multiple times, right?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -554,51 +568,126 @@ public void run() {
...
+
+        for (AccountRecord accountRecord : getLocalAccounts()) {
+            if (!StringUtil.isEmpty(accountRecord.getLocalAccountId())
+                    && accountRecord.getLocalAccountId().equalsIgnoreCase(oid)) {
+                com.microsoft.identity.common.internal.logging.Logger.info(
+                        TAG + methodName,
+                        "Found the matched account."
+                );
+
+                callback.onAccountLoaded(AccountAdapter.adapt(accountRecord));
```
iambmelt: Same here - invocation multiple times?

testapps/testapp/src/main/AndroidManifest.xml
```diff
@@ -55,14 +55,15 @@
...
+                <data android:scheme="msauth"
+                    android:host= "com.microsoft.identity.client.sample.local"
+                    android:path= "/QK0hWtPIQviyU3IX8AhunaS0IY4=" />
             </intent-filter>
         </activity>
 
         <meta-data
             android:name="com.microsoft.identity.client.ClientId"
-            android:value="9851987a-55e5-46e2-8d70-75f8dc060f21"/>
+            android:value="b92e0ba5-f86e-4411-8e18-6b5f928d968a"/>
```
iambmelt: Intentional?

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
```diff
@@ -37,6 +37,9 @@
 import android.widget.Switch;
 import android.widget.TextView;
 
+import com.google.gson.Gson;
```
iambmelt: Unused imports being added to this file

testapps/testapp/src/main/res/raw/msal_config_webview.json
```diff
@@ -1,7 +1,7 @@
 {
-  "client_id" : "9851987a-55e5-46e2-8d70-75f8dc060f21",
+  "client_id" : "2523b71d-3ac3-4aea-833d-7096286ec643",
```
iambmelt: Intentional?

heidijinxujia: Could change back after Gladwin updates the redirectUri

testapps/testapp/src/main/res/raw/msal_config.json
```diff
@@ -1,7 +1,7 @@
 {
-  "client_id" : "9851987a-55e5-46e2-8d70-75f8dc060f21",
+  "client_id" : "b92e0ba5-f86e-4411-8e18-6b5f928d968a",
```
heidijinxujia: 9851987a-55e5-46e2-8d70-75f8dc060f21  is not lab app client id. Going to use 4b0db8c2-9f26-4417-8bde-3f0e3656f8e0 instead.

heidijinxujia:
Thank you, Brian! I've addressed all your comments.
1. `9851987a-55e5-46e2-8d70-75f8dc060f21` is not a lab app. Could not find it anywhere under my subscriptions. So I update the client id to lab's as well as the redirect uri.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -603,7 +590,7 @@ public void getAccount(@NonNull final String oid,
     }
 
     private void getBrokerAndLocalAccount(@NonNull final String oid,
-                                          @NonNull final AccountLoadedCallback callback) {
+                                          @NonNull final AccountsLoadedCallback callback) {
```
iambmelt: This should generate a warning for using a raw type - will this only ever return 1?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -554,51 +555,128 @@ public void run() {
...
+        final String methodName = ":getBrokerAndLocalAccount";
 
-        if (null != realm) {
-            accountToReturn = AccountAdapter.getAccountInternal(
-                    mPublicClientConfiguration.getClientId(),
-                    mPublicClientConfiguration.getOAuth2TokenCache(),
-                    homeAccountIdentifier,
-                    realm
-            );
+        final Handler handler;
```
iambmelt: `getAccounts()` and `getBrokerLocalAccounts()` have the same piece of logic - let's split this into a convenience function like `getActiveLooper()` or `getPreferredLooper()`

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -554,51 +555,128 @@ public void run() {
...
+                                    && accountRecord.getLocalAccountId().equalsIgnoreCase(oid)) {
+                                com.microsoft.identity.common.internal.logging.Logger.info(
+                                        TAG + methodName,
+                                        "Found the matched account from broker cache."
+                                );
+
+                                handler.post(new Runnable() {
+                                    @Override
+                                    public void run() {
+                                        callback.onAccountsLoaded(AccountAdapter.adapt(accountRecord));
```
iambmelt: This callback is being invoked inside a loop, surely you mean to call this only once, right?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -554,51 +555,128 @@ public void run() {
...
+
+        for (AccountRecord accountRecord : getLocalAccounts()) {
+            if (!StringUtil.isEmpty(accountRecord.getLocalAccountId())
+                    && accountRecord.getLocalAccountId().equalsIgnoreCase(oid)) {
+                com.microsoft.identity.common.internal.logging.Logger.info(
+                        TAG + methodName,
+                        "Found the matched account."
+                );
+
+                callback.onAccountsLoaded(AccountAdapter.adapt(accountRecord));
```
iambmelt: Same here - callback invocation inside of a loop

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -554,51 +557,127 @@ public void run() {
...
+                        if (cachedAccounts.isEmpty()) {
+                            com.microsoft.identity.common.internal.logging.Logger.info(
+                                    TAG + methodName,
+                                    "No account stored in the broker cache."
+                            );
+                            handler.post(new Runnable() {
+                                @Override
+                                public void run() {
+                                    //Return null if no account found.
+                                    callback.onAccountsLoaded(null);
```
iambmelt: Let's return an empty `List` instead - otherwise we force callers to check for both `null` and `isEmpty()`

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -384,23 +385,25 @@ public PublicClientApplicationConfiguration getConfiguration() {
         void onAccountsLoaded(List<AccountRecord> accountRecords);
     }
 
+    private static Handler getPreferredLooper() {
```
iambmelt: We can decorate this method's return type as `@NonNull`

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -473,7 +476,7 @@ public void onMigrationFinished(int numberOfAccountsMigrated) {
      * @param handler : handler to post
      * @param callback: AccountsLoadedCallback
      */
-    private void postLocalAccountsResult(final Handler handler, final AccountsLoadedCallback callback) {
+    private void postLocalAccountsResult(final Handler handler, final AccountsLoadedCallback<List<IAccount>> callback) {
```
iambmelt: Both params here `@NonNull`

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -493,7 +496,7 @@ public void run() {
      * @param handler : handler to post
      * @param callback: AccountsLoadedCallback
      */
-    private void postBrokerAndLocalAccountsResult(final Handler handler, final AccountsLoadedCallback callback) {
+    private void postBrokerAndLocalAccountsResult(final Handler handler, final AccountsLoadedCallback<List<IAccount>> callback) {
```
iambmelt: Both params `@NonNull`

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -554,51 +557,127 @@ public void run() {
...
-            realm = audience.getTenantId();
+        if (MSALControllerFactory.brokerEligible(
+                mPublicClientConfiguration.getAppContext(),
+                mPublicClientConfiguration.getDefaultAuthority(),
+                mPublicClientConfiguration)) {
+            com.microsoft.identity.common.internal.logging.Logger.info(
+                    TAG + methodName,
+                    "Eligible to use broker. Try to get the account from local and broker cache."
+            );
+            getBrokerAndLocalAccount(oid, callback);
```
iambmelt: `getBrokerAndLocalAccounts()` (plural)

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -554,51 +557,127 @@ public void run() {
...
-     * @param homeAccountIdentifier The home_account_id of the sought IAccount.
-     * @param authority             The authority of the sought IAccount.
-     * @return The IAccount stored in the cache or null, if no such matching entry exists.
+     * @param oid The object id of the sought IAccount.
+     * @param callback AccountLoadedCallback
      */
-    @Nullable
-    public IAccount getAccount(@NonNull final String homeAccountIdentifier,
-                               @Nullable final String authority) {
+    public void getAccount(@NonNull final String oid,
```
iambmelt: I think that the factoring of this method is more complicated than it needs to be....

```java
void getAccount(@NonNull final String oid, @NonNull final String AccountsLoadedCallback<IAccount> callback) {
    final AccountsLoadedCallback<IAccount> brokerCallback = new AccountsLoadedCallback<>() {
        @Override
        public void onAccountsLoaded(@Nullable final IAccount account) {
            callback.onAccountsLoaded(account);
        }
    }

    getLocalAccountByOid(oid, new AccountsLoadedCallback<IAccount>() {
        @Override
        public void onAccountsLoaded(@Nullable final IAccount account) {
            if (null == account) {
                getBrokerAccountByOid(oid, brokerCallback); // Method yields null if not broker elligible
            } else {
                callback.onAccountsLoaded(account);
            }
        }
    })
}
```

Then.... we define two methods...

```java
// One method to load accounts locally and return a List
private void getLocalAccountByOid(@NonNull String oid, @NonNull final AccountsLoadedCallback<IAccount> callback) {
    // On a worker thread, if necessary....
    final List<IAccount> accounts = getLocalAccounts();
    callback.onAccountsLoaded(getAccountByOid(acounts));
}
````

````java
// Another method to load Accounts from Broker and return a List
private void getBrokerAccountByOid(@NonNull String oid, @NonNull final AccountsLoadedCallback<IAccount> callback) {
    // On a worker thread, if necessary....
    final List<IAccount> accounts = getBrokerAccounts(); // null or empty, as appropriate (null for no-broker, empty for no-accounts)
    callback.onAccountsLoaded(getAccountByOid(accounts));
}
````

And getAccountByOid() does all the filtering once....

```java
public IAccount getAccountByOid(@NonNull final String oid, @NonNull List<IAccount> accounts) {
    // fill in the rest here....
}
````

File: msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java [hunk 18]
```diff
@@ -554,51 +557,127 @@ public void run() {
...
+                                    //Return null if no account found.
+                                    callback.onAccountsLoaded(null);
+                                }
+                            });
+
+                            return;
+                        }
+
+                        for (final AccountRecord accountRecord : cachedAccounts) {
+                            if (!StringUtil.isEmpty(accountRecord.getLocalAccountId())
```
iambmelt: This loop and its conditions are duplicated across methods [here ](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/8709d0d6214f9e67a778096fcf102801752bc8ff/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java#L626-L643)and [here](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/8709d0d6214f9e67a778096fcf102801752bc8ff/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java#L663-L673) -- we should split this into a few different functions that we can call again and again, rather than duplicate --[ I have specific suggestions here](https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/592/files#r281266640).

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
heidijinxujia: This function is trying to find the target account, the one with matching oid, from local msal and broker's cache. Not the entire account list.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
heidijinxujia: I'm afraid the getBrokerAccountByOid cannot be refactored in this way. Because the call to broker cannot happen on main thread like the getLocalAccount, so it has to call the looper to post the result on a different thread. Thus, it is better to get both broker and local accounts on the same thread, fetch the target account and return it back.

heidijinxujia:
Comments addressed. Please have a look. Thanks.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -555,51 +561,104 @@ public void run() {
...
+        com.microsoft.identity.common.internal.logging.Logger.info(
+                TAG + methodName,
+                "Start finding account from local cache."
+        );
+        List<AccountRecord> localAccountList = getLocalAccounts();
+        callback.onAccountsLoaded(getAccountByOid(oid, localAccountList));
+    }
+
+    private IAccount getAccountByOid(@NonNull final String oid,
+                                     List<AccountRecord> accounts) {
```
iambmelt: Minor nit here on `final` and `@NonNull`

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
```diff
@@ -269,21 +269,11 @@ public PublicClientApplication getPublicClientApplication() {
     public void onAcquireTokenSilentClicked(final AcquireTokenFragment.RequestOptions requestOptions) {
         prepareRequestParameters(requestOptions);
 
-        //final IAccount requestAccount = getAccount();
-        mApplication.getAccounts(new PublicClientApplication.AccountsLoadedCallback() {
+        mApplication.getAccount(requestOptions.getLoginHint().trim(), new PublicClientApplication.AccountsLoadedCallback<IAccount>() {
```
iambmelt: Why is the `loginHint` being passed here?

heidijinxujia: This is the loginhint text box in our test app. We can pass in the oid for testing here. I will add a comment there.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -555,51 +561,104 @@ public void run() {
...
-     * @param homeAccountIdentifier The home_account_id of the sought IAccount.
-     * @param authority             The authority of the sought IAccount.
-     * @return The IAccount stored in the cache or null, if no such matching entry exists.
+     * @param oid The object id of the sought IAccount.
+     * @param callback AccountLoadedCallback
      */
-    @Nullable
-    public IAccount getAccount(@NonNull final String homeAccountIdentifier,
-                               @Nullable final String authority) {
+    public void getAccount(@NonNull final String oid,
```
shoatman: I believe that we'll need to support getting an account by OID, HomeAccountId, etc... so let's make that possible by accepting a parameter object.

heidijinxujia:
This PR is out of sync after AccountManager change. Close this one and open another PR for it.

PR #594, opened 2019-04-30
Add APIs for acquireToken with resource id

heidijinxujia:
Add the builder method for `withResource(String resource)` into TokenParameters.java, which adapts the resource string into scope string array.

The developer can only use either resource string or scope string array when creating the acquireTokenParameter and acquireTokenSilentParameter, not both at the same time. If the dev sets the scope after setting the resource string, the scope array will be overwritten.

To acquireToken interactively
```java
 AcquireTokenParameters.Builder builder = new AcquireTokenParameters.Builder();
 AcquireTokenParameters acquireTokenParameters = builder.startAuthorizationFromActivity(this)
                    .withResource(mResource)
                    .withUiBehavior(mUiBehavior)
                    .withAuthorizationQueryStringParameters(mExtraQp)
                    .callback(getAuthenticationCallback())
                    .withLoginHint(mLoginHint)
                    .build();

 mApplication.acquireTokenAsync(acquireTokenParameters);
```

To aquireToken silently
```java
mApplication.getAccounts(new PublicClientApplication.AccountsLoadedCallback() {
            @Override
            public void onAccountsLoaded(final List<IAccount> accounts) {
                IAccount requestAccount = null;

                for (final IAccount account : accounts) {
                    if (account.getUsername().equalsIgnoreCase(requestOptions.getLoginHint().trim())) {
                        requestAccount = account;
                        break;
                    }
                }

                if (null != requestAccount) {
                    AcquireTokenSilentParameters.Builder builder = new AcquireTokenSilentParameters.Builder();
                    AcquireTokenSilentParameters acquireTokenSilentParameters =
                            builder.withResource(mResource)
                                    .forAccount(requestAccount)
                                    .forceRefresh(mForceRefresh)
                                    .callback(getAuthenticationCallback())
                                    .build();

                    mApplication.acquireTokenSilentAsync(acquireTokenSilentParameters);
                } else {
                    showMessage("No account found matching loginHint");
                }
            }
        });
```

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -716,40 +716,6 @@ public final class PublicClientApplication {
      * Acquire token interactively, will pop-up webUI. Interactive flow will skip the cache lookup.
      * Default value for {@link UiBehavior} is {@link UiBehavior#SELECT_ACCOUNT}.
      *
-     * @param activity    Non-null {@link Activity} that is used as the parent activity for launching the {@link AuthenticationActivity}.
-     *      *                 All apps doing an interactive request are required to call the
-     *      *                 {@link PublicClientApplication#handleInteractiveRequestRedirect(int, int, Intent)} within the calling
-     *      *                 activity {@link Activity#onActivityResult(int, int, Intent)}.
-     * @param resource    The non-null string of resource identifier.
-     * @param callback    The {@link AuthenticationCallback} to receive the result back.
-     *      *                 1) If user cancels the flow by pressing the device back button, the result will be sent
-     *      *                 back via {@link AuthenticationCallback#onCancel()}.
-     *      *                 2) If the sdk successfully receives the token back, result will be sent back via
-     *      *                 {@link AuthenticationCallback#onSuccess(IAuthenticationResult)}
-     *      *                 3) All the other errors will be sent back via
-     *      *                 {@link AuthenticationCallback#onError(MsalException)}.
-     */
-    public void acquireToken(@NonNull final Activity activity,
-                             @NonNull final String resource,
-                             @NonNull final AuthenticationCallback callback) {
-        acquireToken(
-                activity,
-                mapResourceToDefaultScope(resource),
-                null, // account
-                null, // uiBehavior
-                null, // extraQueryParams
-                null, // extraScopes
-                null, // authority
-                callback,
-                null, // loginHint
-                null // claimsRequest
-        );
-    }
-
-    /**
-     * Acquire token interactively, will pop-up webUI. Interactive flow will skip the cache lookup.
-     * Default value for {@link UiBehavior} is {@link UiBehavior#SELECT_ACCOUNT}.
-     *
      * @param activity  Non-null {@link Activity} that will be used as the parent activity for launching the {@link AuthenticationActivity}.
      *                  All the apps doing interactive request are required to call the
      *                  {@link PublicClientApplication#handleInteractiveRequestRedirect(int, int, Intent)} within the calling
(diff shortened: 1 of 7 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -750,6 +784,43 @@ public void acquireToken(@NonNull final Activity activity,
...
+     *                  {@link AuthenticationCallback#onError(MsalException)}.
+     */
+    public void acquireToken(@NonNull final Activity activity,
+                             @NonNull final String resource,
+                             @Nullable final String loginHint,
+                             @NonNull final AuthenticationCallback callback) {
+        acquireToken(
+                activity,
+                mapResourceToDefaultScope(resource),
+                null, // account
```
iambmelt: Thanks for commenting next to the `null` parameters - so much easier to read 😄

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -712,6 +712,40 @@ public void acquireToken(@NonNull final Activity activity,
...
+     *      *                 {@link AuthenticationCallback#onSuccess(IAuthenticationResult)}
+     *      *                 3) All the other errors will be sent back via
+     *      *                 {@link AuthenticationCallback#onError(MsalException)}.
+     */
+    public void acquireToken(@NonNull final Activity activity,
+                             @NonNull final String resource,
+                             @NonNull final AuthenticationCallback callback) {
+        acquireToken(
+                activity,
+                mapResourceToDefaultScope(resource),
```
iambmelt: Although we've tagged the `resource` parameter here as `@NonNull` I think we shouldn't trust the end-developer to adhere... can we null-check this param before we call `acquireToken()`?

iambmelt: I also just realized we have two methods to do this `mapResourceToDefaultScope` conversion.... the other is in `TokenCacheItemMigrationAdapter` though the return types are slightly different and _that_ implementation doesn't call `.trim()` which seems like a good protective measure....

If you're willing to take combining these bits of code as an addendum to this PR it'd be a nice bit of "gold plating" but I won't request changes based on this alone

iambmelt: Oh, just kidding, I didn't realize that was a _new_ method. In that case, let's try to share code if we can...

I think having a method to transform a resource into a single scope makes sense, maybe put a wrapper function around it to box the result into a `String[]` - feel free to move out of `TokenCacheMigrationAdapter` and maybe move into `StringUtil` or some other appropriate class

shoatman: I'd be inclined to put this kind of helper function on the MicrosoftOAuth2Strategy.... since it's AAD specific logic.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -750,6 +784,43 @@ public void acquireToken(@NonNull final Activity activity,
...
+     *                  which will have the UPN pre-populated.
+     * @param callback  The Non-null {@link AuthenticationCallback} to receive the result back.
+     *                  1) If user cancels the flow by pressing the device back button, the result will be sent
+     *                  back via {@link AuthenticationCallback#onCancel()}.
+     *                  2) If the sdk successfully receives the token back, result will be sent back via
+     *                  {@link AuthenticationCallback#onSuccess(IAuthenticationResult)}
+     *                  3) All the other errors will be sent back via
+     *                  {@link AuthenticationCallback#onError(MsalException)}.
+     */
+    public void acquireToken(@NonNull final Activity activity,
```
shoatman: I don't think we need to add resourceId as a separate parameter and introduce a whole new set of overloads.  I don't expect many users to use these overloads... instead I suspect that most users will prefer the newly added overload that accepts the AcquireTokenParameters object.  I'd just add resource id to that object and not introduce additional overloads to MSAL

heidijinxujia: First, yes, agree that we should not add new set of overloads. Second, instead of adding new parameters, I think we should add different initialization methods of AcquireTokenParameter and AcquireTokenSilentParameter, like initWithScope() and initWithResource(). And let the developer to choose which to use. As at our end, there is no way (at lease haven't find a perfect one) to distinguish between resource string/id and scopes.

heidijinxujia:
@iambmelt Let me know if everything look good for you when you get time. Thanks.

msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
```diff
@@ -202,6 +204,18 @@ public B callback(AuthenticationCallback callback) {
             return self();
         }
 
+        public B withResource(final String resource) {
+            if (!StringUtil.isEmpty(resource)) {
+                mScopes = new ArrayList<String>() {{
+                    add(resource.toLowerCase().trim() + "/.default");
+                }};
+            } else {
+                mScopes = new ArrayList<>();
```
iambmelt: When would this ever be valid? We should throw an `IllegalArgumentException`, rather than let developers pass null and get some _other_ cryptic error elsewhere

msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
```diff
@@ -202,6 +204,18 @@ public B callback(AuthenticationCallback callback) {
             return self();
         }
 
+        public B withResource(final String resource) {
+            if (!StringUtil.isEmpty(resource)) {
+                mScopes = new ArrayList<String>() {{
```
iambmelt: We should check if `null != mScopes` and, if it is, throw an `Exception` (perhaps `IllegalStateException` or `IllegalArgumentException`) [since scopes and resources cannot be combined in a single request](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent#the-default-scope).

The same logic would need to be added to `withScopes(List)` to ensure the behavior is the same whether `withScopes()` or `withResource()` is called first

heidijinxujia: Thank you, Brian! Addressed.

PR #596, opened 2019-05-02
Update test app

rpdome:
- Add "PublicApplication mode" to the UI.
- Add support to the new PublicApplication.create() method and the two new PublicApplication modes, SingleAccount (shared device mode) and MultipleAccount (default MSAL behavior).
- Putting MSAL operation in a wrapper.
- Using the listener to update UI (to support MSAL asynchronous operations).
- AcquireTokenSilent now only performs silent call. (remove fallback logic to invoke acquireToken - as that sort of defeat purpose of testing acquireTokenSilent)
- Clean up some unused codes.

Will change the branch target to dev once #591 is merged.

### DIFF
--- a/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
+++ b/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
@@ -37,12 +37,14 @@ import android.widget.Spinner;
 import android.widget.Switch;
 import android.widget.TextView;
 
+import com.microsoft.identity.client.IAccount;
 import com.microsoft.identity.client.UiBehavior;
 import com.microsoft.identity.common.exception.ClientException;
 import com.microsoft.identity.common.internal.ui.browser.Browser;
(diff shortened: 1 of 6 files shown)


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
```diff
@@ -166,7 +170,14 @@ private void setCurrentDefaultBrowserValue() {
     public void onResume() {
         super.onResume();
         if (mOnFragmentInteractionListener != null) {
-            mOnFragmentInteractionListener.bindSelectAccountSpinner(mSelectAccount);
+            MsalWrapper.sharedInstance().registerPostAccountLoadedJob("bindSelectAccountSpinner",
```
shoatman: getInstance() instead of sharedInstance()?

rpdome:
Switch the target branch to dev as the 'Factory' branch is already merged into dev. (#591)

rpdome:
Thanks @shoatman!

PR #597, opened 2019-05-02
Rremove timeout for Auth request

kreedula:
Also fixed an issue where BrokerResult is retrieved as serializable instead of string


kreedula:
Thanks @shoatman

PR #599, opened 2019-05-03
Implements SingleAccount and MultipleAccountPublicCilentApplication

rpdome:
This is a part of the FLW/End my Shift feature.

- Move codes (that should be under IMultipleAccountPublicCilentApplication) from PublicClientApplicaiton
- Implements SingleAccountPublicClientApplication
- Implements globalSignOut() and getCurrentAccount() in BrokerMsalController (for invoking broker operations).
- PublicClientApplication.create() now returns the right PublicClientApplication (based on Broker mode).

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
@@ -53,7 +53,7 @@ public interface ISingleAccountPublicClientApplication extends IPublicClientAppl
      * @param callback a callback to be invoked when the operation finishes.
      * @throws MsalClientException if this function is invoked when the app is no longer in the single account mode.
      */
-    void globalSignOut(final AccountRemovedListener callback) throws MsalClientException;
+    void removeCurrentAccount(final AccountRemovedListener callback) throws MsalClientException;
 
     /**
      * Callback for asynchronous loading of the msal IAccount account.
--- a/msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
@@ -76,17 +76,4 @@ public interface ISingleAccountPublicClientApplication extends IPublicClientAppl
          * */
         void onAccountChanged(final IAccount priorAccount, final IAccount currentAccount);
     }
-
-    /**
-     * Callback for asynchronous loading of broker AccountRecord account (in Single account mode).
-     */
-    interface SingleAccountLoadedCallback {
-
-        /**
-         * Called once the signed-in account (if there is any), has been loaded from the broker.
-         * @param accountRecord The accountRecord in broker. This could be null.
-         */
-        void onAccountLoaded(@Nullable final AccountRecord accountRecord);
-    }
-
 }
--- a/msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
@@ -23,7 +23,10 @@
 
 package com.microsoft.identity.client;
 
+import android.support.annotation.Nullable;
+
 import com.microsoft.identity.client.exception.MsalClientException;
+import com.microsoft.identity.common.internal.dto.AccountRecord;
 
 /**
  * An interface that contains list of operations that are available when MSAL is in 'single account' mode.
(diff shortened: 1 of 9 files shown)


iambmelt:
... Reviewing now

msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
```diff
@@ -72,6 +75,18 @@
...
+    /**
+     * Callback for asynchronous loading of broker AccountRecord account (in Single account mode).
+     */
+    interface SingleAccountLoadedCallback {
 
+        /**
+         * Called once the signed-in account (if there is any), has been loaded from the broker.
+         * @param accountRecord The accountRecord in broker. This could be null.
+         */
+        void onAccountLoaded(@Nullable final AccountRecord accountRecord);
```
iambmelt: Just verifying that this is not intended for public consumption, right?

iambmelt: (as in, `AccountRecord` is an internal class and shouldn't be exposed to users of the library directly) - just clarifying intent here.... can't recall if inner-interfaces of `public interfaces` are implicitly `public` or not

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -384,4 +444,66 @@ private Bundle getRequestBundleForRemoveAccount(@Nullable final IAccount account
...
+                            }
+                        }
+                    });
+                } catch (final BaseException | InterruptedException | ExecutionException | RemoteException e) {
+                    com.microsoft.identity.common.internal.logging.Logger.error(
+                        TAG,
+                        "Exception is thrown when trying to perform global sign-out."
+                            + e.getMessage(),
+                        e);
+                    callback.onAccountRemoved(false);
```
iambmelt: Hmm... I wish we had some way to bubble the `Exception` up and tell why.... Maybe a future enhancement if the need arises

msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
iambmelt: Yeah, it looks like this would be implicitly `public` from this post: https://beginnersbook.com/2016/03/nested-or-inner-interfaces-in-java/

>Nested interfaces declared inside class can take any access modifier, however nested interface declared inside interface is public implicitly.

rpdome: This is a callback function that is used by BrokerMsalController. User shouldn't need this.

As the concept of package-private interface doesn't exist in java, here are the things I'd propose.

1. Rename the function and add the description that this is for internal use only.
2. Move this out of the ISingleAccount interface, and put it in the actual object itself. (So that the interface would only contains dev-exposing interfaces.

What do you think?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
rpdome: For now I'm following the pattern we use for all binding functions, but I agree that bubbling this up might be helpful for devs.

msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
iambmelt: #2 sounds good to me

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
iambmelt: We can defer this for later. I imagine that if a substantial amount of customers hit this we'll be adding telemetry and more comprehensive error handling around it. I don't want this to be blocker - we can always capture device logs if need be

msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
```diff
@@ -34,6 +34,17 @@
     implements IMultipleAccountPublicClientApplication {
     private static final String TAG = MultipleAccountPublicClientApplication.class.getSimpleName();
 
+    /**
+     * Callback for asynchronous loading of broker AccountRecord accounts (in Multiple account mode).
+     */
+    public interface GetAccountRecordsFromBrokerCallback {
```
iambmelt: Can we declare this inside of the `BrokerMsalController` instead so that even though this is `public` it remains inside the `*.internal.*` package space?

PR #612, opened 2019-05-20
Remove stale sample testapp

rpdome:


rpdome:
fyi. This will need to go in in order for MSAL to build @iambmelt @shoatman @heidijinxujia.  (It's a stale test app that is using the 'old' PCA constructor.)

iambmelt:
@rpdome Let's see what this conflict is in `AuthUtil` and if someone is using it.

PR #618, opened 2019-05-22
Add the getAccount api to retrieve account by different overload

heidijinxujia:
The current requirement is enabled to retrieve the IAccount object by `HomeAccountIdentifier`, `LocalAccountIdentifier` or `oid` or `username` or `upn`.

There are several possible solutions here.
1. Introduce multiple getAccount API to take different overload. The drawback will be on the future extensibility on the overloads.
2. Create a parameter class and take it as input para for getAccount API. This will be problematic when the dev set both homeAccountIdentifier and localAccountIdentifier mistakenly.
3. ~~Introduce the query type enum, which describes which overload is used. And take the query type, the query string as input for the getAccount API.~~
Use the identifier string directly and use it to search all the possible fields.

- The getAccount API takes two parameters, queryString and the callback object. Callback is needed here because the necessary background call to retrieve the accounts from Broker.
```
public void getAccount(
                           @NonNull final String queryString,
                           @NonNull final GetAccountCallback callback)
{
    ...
}

public interface GetAccountCallback extends TaskCompletedCallbackWithError<IAccount, Exception> {
        /**
         * Called once succeed and pass the result object.
         *
         * @param result the success result.
         */
        void onTaskCompleted(IAccount result);

        /**
         * Called once exception thrown.
         *
         * @param exception
         */
        void onError(Exception exception);
    }
```

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -500,22 +500,9 @@ public final class PublicClientApplication {
     }
 
     /**
-     * Enum for the getAccount query types,
-     * which enable the user to getAccount with homeAccountIdentifier,
-     * localAccountIdentifier or username.
-     *
-     */
-    public enum AccountQueryType {
-        HOME_ACCOUNT_IDENTIFIER,
-        LOCAL_ACCOUNT_IDENTIFIER,
-        USERNAME
-    }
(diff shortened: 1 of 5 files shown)


shoatman:
How do you feel about just accepting a string parameter called "identifier" and then using that identifier to search all of the fields?

heidijinxujia:
Thank you, @shoatman Addressed.

PR #625, opened 2019-05-25
Fix the redirect initialization for PCA configuration

heidijinxujia:
Fix the redirect initialization for PCA configuration to apply the new redirect uri schema.
Removed the unused private function.


shoatman:
lgtm

PR #629, opened 2019-05-29
TenantProfile Implementation

iambmelt:
Supersedes #565


iambmelt:
Notes about root `IAccount` behavior

- In the result of an `acquireToken` or `acquireTokenSilent` call, the root `IAccount` is the account that was last authorized
- When calling `getAccounts`...
    - What if there is no home account?

iambmelt:
TODO (5/29):
- Finish implementation of `MultipleAccountPublicClientApplication#removeAccount()`
- Broker APIs to support TenantProfile

msal/src/main/java/com/microsoft/identity/client/tenantprofile/ITenantProfile.java
```diff
@@ -0,0 +1,31 @@
...
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.client.tenantprofile;
+
+import android.support.annotation.NonNull;
+
+public interface ITenantProfile extends IAccount {
```
shoatman: I thought about this a bit.  I think it comes down to whether we think we will ever add anything to Account that will not also be a part of TenantProfile.  At the moment, I can't think of anything specifically....

msal/src/main/java/com/microsoft/identity/client/tenantprofile/MultiTenantAccount.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.Collections;
+import java.util.Map;
+
+public class MultiTenantAccount extends Account implements IMultiTenantAccount {
+
+    private Map<String, ITenantProfile> mTenantProfiles;
+
+    public MultiTenantAccount(@Nullable final String tenantRawIdToken) {
+        super(tenantRawIdToken);
```
shoatman: This is also the homeTenantRawIdToken.  Is it not?

msal/src/main/java/com/microsoft/identity/client/tenantprofile/AccountAdapter.java
```diff
@@ -0,0 +1,261 @@
...
+                new HomeAccountFilter(false)
+        );
+
+        // Then, get all of the guest accounts...
+        final List<ICacheRecord> guestCacheRecords = filterCacheRecords(
+                allCacheRecords,
+                new HomeAccountFilter(true)
+        );
+
+        final List<ICacheRecord> orphanCacheRecords = filterCacheRecords(
```
shoatman: Probably worth explaining what you mean by orphanCacheRecords.... I believe you are using this term to convey a cache record for which their is not corresponding root cache record.  We should probably adopt a consistent terminology Parent/Child.  Local/Foreign.  Root/Projection.  Local/Guest to be able to differentiate between these

msal/src/main/java/com/microsoft/identity/client/tenantprofile/AccountAdapter.java
```diff
@@ -0,0 +1,261 @@
...
+        guestCacheRecords.removeAll(orphanCacheRecords);
+
+        final List<IAccount> rootAccounts = createRootAccounts(homeCacheRecords);
+        appendChildren(rootAccounts, guestCacheRecords);
+        rootAccounts.addAll(createOrphanRoots(orphanCacheRecords));
+
+        return rootAccounts;
+    }
+
+    private static List<IAccount> createOrphanRoots(
```
shoatman: These likely are not "roots" since they are orphaned.... I think we can revisit this naming concern in the future though.

msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
```diff
@@ -281,4 +287,96 @@ public void removeAccount(@Nullable final IAccount account,
...
+        AccountMatcher(@NonNull final AccountMatcher... delegateMatchers) {
+            mDelegateMatchers = delegateMatchers;
+        }
+
+        boolean matches(@NonNull final String identifier,
+                        @NonNull final com.microsoft.identity.client.tenantprofile.IAccount account) {
+            boolean matches = true;
+
+            for (final AccountMatcher matcher : mDelegateMatchers) {
+                matches = matches && matcher.matches(identifier, account);
```
iambmelt: 🐛 Add early escape here -- only 1 matcher needs to match, not all of them...

msal/src/main/java/com/microsoft/identity/client/tenantprofile/Account.java
```diff
@@ -0,0 +1,98 @@
...
+import java.text.ParseException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class Account implements IAccount {
+
+    private static final String TAG = Account.class.getSimpleName();
+
+    private final String mRawIdToken;
+    private String mHomeAccountId;
```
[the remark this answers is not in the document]
iambmelt: This is a variable naming issue; I'll make this more clear. It is the `oid`

msal/src/main/java/com/microsoft/identity/client/tenantprofile/ITenantProfile.java
iambmelt: For some future-proofiness, I'll pull `getClaims()` into its own interface.... something like `IClaimable`

msal/src/main/java/com/microsoft/identity/client/tenantprofile/Account.java
```diff
@@ -0,0 +1,98 @@
...
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.client.tenantprofile;
```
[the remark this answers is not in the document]
iambmelt: My plan is to move this all to the `com.microsoft.identity.client` namespace once the changes are complete

msal/src/main/java/com/microsoft/identity/client/tenantprofile/AccountAdapter.java
```diff
@@ -0,0 +1,261 @@
...
+            @NonNull final List<ICacheRecord> orphanCacheRecords) {
+        // First, bucket the records by homeAccountId to create affinities
+        final Map<String, List<ICacheRecord>> bucketedRecords = new HashMap<>();
+
+        for (final ICacheRecord cacheRecord : orphanCacheRecords) {
+            final String cacheRecordHomeAccountId = cacheRecord.getAccount().getHomeAccountId();
```
iambmelt: This should be the `oid`

msal/src/main/java/com/microsoft/identity/client/tenantprofile/AccountAdapter.java
```diff
@@ -0,0 +1,261 @@
...
+
+        // Now that all of the tokens have been bucketed by an account affinity
+        // box those into a 'rootless' IAccount
+        for (final Map.Entry<String, List<ICacheRecord>> entry : bucketedRecords.entrySet()) {
+            // Create our empty root...
+            final MultiTenantAccount emptyRoot = new MultiTenantAccount(
+                    null // home tenant IdToken.... doesn't exist!
+            );
+
+            // Set the home_account_id of the root, even though we don't have the IdToken...
```
iambmelt: `oid`

msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
iambmelt: Resolved in 37ee73d24a805ae7f75f33e66aa75c6a2b32954a

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
```diff
@@ -39,30 +39,28 @@
...
-
-    public AuthenticationResult(@NonNull final ICacheRecord cacheRecord) {
-        mAccessToken = cacheRecord.getAccessToken();
-        mTenantId = cacheRecord.getAccount().getRealm();
-        mUniqueId = cacheRecord.getAccount().getHomeAccountId();
-        mRawIdToken = cacheRecord.getIdToken().getSecret();
-        mAccount = AccountAdapter.adapt(cacheRecord.getAccount());
+    private final com.microsoft.identity.client.tenantprofile.IAccount mAccount;
+
+    public AuthenticationResult(@NonNull final List<ICacheRecord> cacheRecords) {
```
rpdome: nit: is it possible that this could be an empty list?

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
```diff
@@ -39,30 +39,28 @@
...
-                                @Nullable String rawIdToken,
-                                @NonNull IAccountRecord accountRecord) {
-        mAccessToken = accessToken;
-        mTenantId = accessToken.getRealm();
-        mUniqueId = accessToken.getHomeAccountId();
-        mRawIdToken = rawIdToken;
-        mAccount = AccountAdapter.adapt(accountRecord);
-
-    }
+//    public AuthenticationResult(@NonNull AccessTokenRecord accessToken,
```
heidijinxujia: Are you planning to delete or rewrite this method?

msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -22,73 +22,102 @@
...
+    @Nullable
+    public String getTenantId() { // TODO make this package private...
+        return mHomeTenantId;
     }
 
-    @Override
-    public String getUsername() {
-        return mUsername;
+    @Nullable
+    public String getHomeAccountId() { // TODO make this package private
```
iambmelt: TODO

msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -22,73 +22,102 @@
...
 
-    /**
-     * Sets the username.
-     *
-     * @param username The username to set.
-     */
-    void setUsername(final String username) {
-        mUsername = username;
+    @Nullable
+    public String getTenantId() { // TODO make this package private...
```
iambmelt: TODO

msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -22,73 +22,102 @@
...
-     */
-    void setEnvironment(final String environment) {
+    void setEnvironment(@NonNull final String environment) {
         mEnvironment = environment;
     }
 
-    @Override
-    public String getEnvironment() {
+    @Nullable
+    public String getEnvironment() { // TODO make this package private
```
iambmelt: TODO

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -931,7 +921,11 @@ public void run() {
 
     private Bundle getRequestBundleForGlobalSignOut(@NonNull final IAccount account) {
         final Bundle requestBundle = new Bundle();
-        requestBundle.putString(ACCOUNT_LOGIN_HINT, account.getUsername());
+        // TODO I think a home_account_id might be better here...
+        // TODO If we really want to use a login_hint consider that you may need to dig around
+        // TODO inside of the IAccount to find it... If there is no home tenant, you'll need to inspect
+        // TODO the claims of one of the ITenantProfiles to find what you're looking for
+        requestBundle.putString(ACCOUNT_LOGIN_HINT, "");
```
iambmelt: @rpdome ^^

msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
```diff
@@ -114,15 +114,15 @@ public AcquireTokenResult acquireToken(@NonNull final AcquireTokenOperationParam
...
+                final List<ICacheRecord> cacheRecord = saveTokens(
                         oAuth2Strategy,
                         mAuthorizationRequest,
                         tokenResult.getTokenResponse(),
                         parameters.getTokenCache()
                 );
 
                 acquireTokenResult.setLocalAuthenticationResult(
-                        new LocalAuthenticationResult(cacheRecord)
+                        new LocalAuthenticationResult(cacheRecord.get(0), cacheRecord)
```
iambmelt: Try cleaning up all of these calls to `get(0)`

msal/src/main/java/com/microsoft/identity/client/tenantprofile/Account.java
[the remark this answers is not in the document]
iambmelt: Resolved in 9a753d8314a236f4d58e40b81a9a8b686a5910c0

msal/src/main/java/com/microsoft/identity/client/tenantprofile/Account.java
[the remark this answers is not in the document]
iambmelt: Resolved in 9a753d8314a236f4d58e40b81a9a8b686a5910c0

msal/src/main/java/com/microsoft/identity/client/tenantprofile/MultiTenantAccount.java
iambmelt: It is. Naming corrected in e7d47da577a28423891d6f32c763347047d2edcc

msal/src/main/java/com/microsoft/identity/client/tenantprofile/AccountAdapter.java
iambmelt: The orphans do have a root - a sort of 'stub' `MultiTenantAccount` that has an id, but no claims. We can work on the naming here

msal/src/main/java/com/microsoft/identity/client/tenantprofile/AccountAdapter.java
iambmelt: Keeping this bit of the code as-is - resolved in 9a753d8314a236f4d58e40b81a9a8b686a5910c0

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
iambmelt: Not presently --  we only create this object in the event of success. Failure cases fall into the `MsalException` path where this object never gets constructed

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
iambmelt: Resolved in 71aa5d5760998128a65685013e351387d5338ac3

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -667,18 +656,19 @@ public void run() {
...
             Logger.verbose(TAG + methodName, "Get the broker accounts from auth service.");
-            return getBrokerAccountsWithAuthService(parameters);
+            //return getBrokerAccountsWithAuthService(parameters);
         } else {
             Logger.verbose(TAG + methodName, "Is microsoft auth service supported? " + "[no]");
             Logger.verbose(TAG + methodName, "Get the broker accounts from Account Manager.");
-            return getBrokerAccountsFromAccountManager(parameters);
+            //return getBrokerAccountsFromAccountManager(parameters);
         }
+        return new ArrayList<>(); // TODO remove after implementation
```
iambmelt: TODO light this up once @rpdome's FLW changes are ready

msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java
```diff
@@ -213,16 +247,22 @@ public static AcquireTokenSilentOperationParameters createAcquireTokenSilentOper
...
+
             final AccountRecord requestAccountRecord = new AccountRecord();
-            requestAccountRecord.setEnvironment(account.getEnvironment());
-            requestAccountRecord.setUsername(account.getUsername());
-            requestAccountRecord.setHomeAccountId(account.getHomeAccountIdentifier().getIdentifier());
-            requestAccountRecord.setLocalAccountId(account.getAccountIdentifier().getIdentifier());
+            requestAccountRecord.setEnvironment(multiTenantAccount.getEnvironment());
+            requestAccountRecord.setHomeAccountId(multiTenantAccount.getHomeAccountId());
+
+            // TODO do you mean the home account? The guest account? Where do I look??
```
iambmelt: TODO

msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java
```diff
@@ -213,16 +247,22 @@ public static AcquireTokenSilentOperationParameters createAcquireTokenSilentOper
...
+            requestAccountRecord.setUsername(SchemaUtil.getDisplayableId(multiTenantAccount.getClaims()));
+            // TODO which tenant do you want?????? Using home for now...
+            requestAccountRecord.setLocalAccountId(multiTenantAccount.getId());
```
iambmelt: TODO

msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java
```diff
@@ -337,21 +377,18 @@ public static String getAuthorityFromAccount(final IAccount account) {
...
-                && !StringUtil.isEmpty(aadIdentifier.getTenantIdentifier())) {
+        if (account instanceof MultiTenantAccount) {
+            MultiTenantAccount multiTenantAccount = (MultiTenantAccount) account;
+
             return "https://"
-                    + account.getEnvironment()
+                    + multiTenantAccount.getEnvironment()
                     + "/"
-                    + aadIdentifier.getTenantIdentifier()
+                    + multiTenantAccount.getTenantId() // TODO this assumes home account, but should the default authority be used instead??
```
iambmelt: TODO

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
```diff
@@ -292,20 +287,20 @@ public void onApplicationLoaded() {
...
-                if (accounts != null) {
-                    for (IAccount account : accounts)
-                        add(account.getUsername());
-                }
-            }}
+                getApplicationContext(), android.R.layout.simple_spinner_item,
+                new ArrayList<String>() {{
+                    if (accounts != null) {
+                        for (IAccount account : accounts)
+                            add((String) account.getClaims().get("preferred_username")); // TODO this will break for guest accounts with no home...
```
iambmelt: TODO

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalWrapper.java
```diff
@@ -421,10 +421,10 @@ public void onAccountLoaded(IAccount activeAccount) {
 
                     @Override
                     public void onAccountChanged(IAccount priorAccount, IAccount currentAccount) {
-                        notifyCallback.notify("single account is changed from " +
-                                (priorAccount == null ? "null" : priorAccount.getUsername()) +
-                                " to " +
-                                (currentAccount == null ? "null" : currentAccount.getUsername()));
+                        notifyCallback.notify("single account is changed from "
```
iambmelt: TODO How does this work with guest accounts?

msal/src/main/java/com/microsoft/identity/client/tenantprofile/ITenantProfile.java
iambmelt: Resolved in b5789081ab706f9f537d4432c8fa3523fe4882e6

msal/src/main/java/com/microsoft/identity/client/tenantprofile/AccountAdapter.java
iambmelt: I've renamed this to `guestCacheRecordsWithNoHomeAccount` -- hopefully this is clearer.

Resolved in af766f7ba4aaf018ef97a2fe592dc50703a65342

msal/src/main/java/com/microsoft/identity/client/tenantprofile/AccountAdapter.java
iambmelt: Renamed this method to `createIAccountsForGuestsNotSignedIntoHomeTenant` in af766f7ba4aaf018ef97a2fe592dc50703a65342

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: Resolved in 41b8ca4d0c9b27368c864655858aafea963167cb

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: Resolved in 41b8ca4d0c9b27368c864655858aafea963167cb

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: Resolved in 41b8ca4d0c9b27368c864655858aafea963167cb

msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java
iambmelt: Resolved in 41b8ca4d0c9b27368c864655858aafea963167cb

msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
iambmelt: Resolved in 2b8140384d5bde3240624ca9ec729e49337c17c5

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
```diff
@@ -1,125 +1,296 @@
...
+        final List<ICacheRecord> homeCacheRecords = filterCacheRecords(
+                allCacheRecords,
+                new HomeAccountFilter(false)
         );
 
-        Logger.verbosePII(
-                TAG + methodName,
-                "Environment: [" + accountIn.getEnvironment() + "]"
+        // Then, get all of the guest accounts...
+        final List<ICacheRecord> guestCacheRecords = filterCacheRecords(
```
rpdome: nit: I'd explicitly specify that guestCacheRecords and guestCacheRecordsWithNoHomeAccount are mutually exclusive - not a subset of the other.

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
```diff
@@ -1,125 +1,296 @@
...
+                            .getAccount()
+                            .getEnvironment()
+            );
+
+            result.add(rootAccount);
+        }
+
+        return result;
+    }
+
```
rpdome: nit: @Nullable

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
rpdome: I'm actually removing this - in FLW scenario, there should be only 1 sign-in account in broker - if there's more, we consider it an error state and wipe everything.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
rpdome: These are invoked in MultipleAccountPCA though. SingleAccountPCA only uses getCurrentAccount()

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalWrapper.java
rpdome: If the root profile doesn't exist, we could try iterating through the profiles to find the username.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java
iambmelt: Resolved in 07eff8ae82868d6e0297021c9f8889ebe142bd7d

msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java
iambmelt: Resolved in 07eff8ae82868d6e0297021c9f8889ebe142bd7d

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
iambmelt: I've clarified this via additional comments in 9188954824092606374d0523b66a25839bc1de9e - let me know if you'd like to see anything else here. Thanks @rpdome

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
iambmelt: @rpdome Unless there's something I'm not thinking of these should be `@NonNull` -- I'll annotate it as such

iambmelt: Added in d51181ca79ce5837d96fd710d00db4ee846e9b5f

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalWrapper.java
iambmelt: Resolved in c30ce285f18799c6253e8748f6da83df36c8f230

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
iambmelt: Resolved in 40e1dd030c3de9eb804f5e08730bcd3a2338e1e5

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
heidijinxujia: This is not for FLW, but we need to change the result bundle type from AccountRecord to CacheRecord from MicrosoftAuthService.getAccounts().

iambmelt: Yes, this is a big TODO -- not necessary for Teams build, but definitely necessary

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -49,20 +52,25 @@ public void getCurrentAccount(final CurrentAccountListener listener) throws Msal
...
-
-                        mLocalAccountRecord = accountRecordInBroker;
-                        listener.onAccountLoaded(accountInBroker);
+                    public void onAccountLoaded(@Nullable final List<ICacheRecord> cacheRecords) {
+                        // TODO implement...
+                        // TODO check if the cache records are null before trying to adapt...
+                        // TODO the List that is returned should adapt into a single IAccount
+                        // TODO Dome... you can use this return type to support FLW
+                        final List<IAccount> account = AccountAdapter.adapt(cacheRecords);
+//                        IAccount localAccount = mLocalAccountRecord == null ? null : AccountAdapter.adapt(mLocalAccountRecord);
```
iambmelt: TODO remove in this PR

iambmelt: Resolved in 3ae4ca76f4a909d24bad1a6ed233c540beaf1359

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -706,12 +674,14 @@ public void acquireTokenAsync(@NonNull final AcquireTokenParameters acquireToken
...
+        final MultiTenantAccount multiTenantAccount = (MultiTenantAccount) account;
+
+        if (null != multiTenantAccount && null != multiTenantAccount.getHomeAccountId()) {
             return AccountAdapter.getAccountInternal(
                     mPublicClientConfiguration.getClientId(),
                     mPublicClientConfiguration.getOAuth2TokenCache(),
-                    account.getHomeAccountIdentifier().getIdentifier(),
-                    AccountAdapter.getRealm(account)
+                    multiTenantAccount.getHomeAccountId(),
+                    multiTenantAccount.getTenantId() // TODO This assumes the home account, but what about guests??
```
iambmelt: TODO?

iambmelt: @shoatman I'll resolve this in the morning (6/5/19) -- feel free to review the rest and I'll submit this fix before merge

heidijinxujia:
Could you have a look with this issue? https://github.com/AzureAD/ad-accounts-for-android/issues/1012

msal/src/main/java/com/microsoft/identity/client/AuthenticationResultAdapter.java
```diff
@@ -29,9 +29,7 @@
     static IAuthenticationResult adapt(ILocalAuthenticationResult localAuthenticationResult) {
 
         IAuthenticationResult authenticationResult = new AuthenticationResult(
-                localAuthenticationResult.getAccessTokenRecord(),
-                localAuthenticationResult.getIdToken(),
-                localAuthenticationResult.getAccountRecord()
+                localAuthenticationResult.getCacheRecordWithTenantProfileData()
```
heidijinxujia: NPE thrown here when I tried to make an ATS call with broker.

06-04 17:55:07.269 25259-25259/com.microsoft.identity.client.sample.local E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.microsoft.identity.client.sample.local, PID: 25259
    java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.Object java.util.List.get(int)' on a null object reference
        at com.microsoft.identity.client.AuthenticationResult.<init>(AuthenticationResult.java:48)
        at com.microsoft.identity.client.AuthenticationResultAdapter.adapt(AuthenticationResultAdapter.java:32)
        at com.microsoft.identity.client.PublicClientApplication$5.onSuccess(PublicClientApplication.java:973)
        at com.microsoft.identity.common.internal.controllers.ApiDispatcher$4$2.run(ApiDispatcher.java:418)
        at android.os.Handler.handleCallback(Handler.java:739)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:148)
        at android.app.ActivityThread.main(ActivityThread.java:5417)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)

iambmelt:
@shoatman af92ddd367722358ce106c3950ee634a3467934f contains a workaround that I think can be tidied up further as fyi. Don't have an immediate better solution in mind currently

msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -22,73 +22,102 @@
...
-    public IAccountIdentifier getAccountIdentifier() {
-        return mAccountIdentifier;
-    }
+    public String getId() {
+        String id;
+
+        if (null != mRawIdToken && null != getClaims()) {
+            id = (String) getClaims().get(MicrosoftIdToken.OBJECT_ID);
+        } else {
+            id = mHomeOid;
```
Om83: Why are we returning Home OID when OID is not there? Trying to understand the scenario here.

msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -22,73 +22,102 @@
...
+
+        final Map<String, Object> result = new HashMap<>();
+
+        try {
+            final JWT jwt = JWTParser.parse(rawIdToken);
+            final JWTClaimsSet claimsSet = jwt.getJWTClaimsSet();
+            result.putAll(claimsSet.getClaims());
+        } catch (ParseException e) {
+            Logger.error(
+                    TAG + methodName,
```
Om83: Instead of explicitly capturing the methodName wouldn't be better to use something like getEnclosingMethod()

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
```diff
@@ -1,125 +1,311 @@
...
+        private final boolean mInverseMatch;
 
-                // Set the uid as the objectId
-                setObjectIdentifier(components[0]);
+        /**
+         * Creates a new HomeAccountFilter.
+         *
+         * @param inverseMatch If true, only guest accounts will be returned. If false, only home.
+         */
+        HomeAccountFilter(final boolean inverseMatch) {
```
Om83: Also for better readability you can probably use Enum 'AccountTypeToFilter' with values Home and  Guest.

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
```diff
@@ -120,7 +124,8 @@ void acquireToken(@NonNull final Activity activity,
                       @Nullable final String loginHint,
                       @NonNull final UiBehavior uiBehavior,
                       @Nullable final List<Pair<String, String>> extraQueryParameters,
-                      @NonNull final AuthenticationCallback callback);
+                      @NonNull final AuthenticationCallback callback
+    );
```
Om83: nit:Why do we need to have these diffs i.e. why moving the bracket to the next lines?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -157,6 +148,12 @@
     private static final String INTERNET_PERMISSION = "android.permission.INTERNET";
     private static final String ACCESS_NETWORK_STATE_PERMISSION = "android.permission.ACCESS_NETWORK_STATE";
 
+    /**
+     * Constant used to signal a home account's tenant id should be used when performing cache lookups
+     * relative to creating OperationParams.
+     */
+    private static final String FORCE_HOME_LOOKUP = "force_home";
```
Om83: "force_home_lookup" instead?

msal/src/main/java/com/microsoft/identity/client/Account.java
shoatman: I believe the reason why this isn't being done is because that method requires the use of reflection.  Also which object did you want to get the class for within the method to determine the enclosing method?

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
shoatman: Since you've been using polymorphism for other filters.... although it's convenient in this case to just "switch" the mode.  I'd suggest sticking with polymorphism and implementing a GuestAccountFilter.

msal/src/main/java/com/microsoft/identity/client/Account.java
shoatman: This is supposed to be the home tenant object id if available.  Let's review the identifier situation together.

msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -22,73 +22,102 @@
...
+    String getTenantId() {
+        return mHomeTenantId;
     }
 
-    @Override
-    public String getUsername() {
-        return mUsername;
+    @Nullable
+    String getHomeAccountId() {
+        return getId() + "." + mHomeTenantId;
```
shoatman: We want to use the objectid of the home tenant as well.

msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -22,73 +22,102 @@
...
         return mEnvironment;
     }
+
+    @Nullable
+    @Override
+    public Map<String, ?> getClaims() {
+        Map<String, ?> claims = null;
+
+        if (null != mRawIdToken) {
+            claims = getClaims(mRawIdToken);
```
shoatman: getClaims always returns a map... even if the mRawIdToken is null.... I noticed that rawIdToken was nullable... i'm assuming this is to address the scenario when no home account is available.

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: When there is no idtoken present (say, in the case where a guest has authorized without authorizing to their home tenant) I am populating this value with the oid returned from `client_info`

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: In general reflection comes at fairly high price in terms of performance and the tradeoff here is this:

```java
String methodName = ":getFoo";
```

vs

```java
String methodName = getClass().getEnclosingMethod().getName();
```

IntelliJ / AndroidStudio is smart enough to change these values when autorefactor is used to renamed methods. Since string allocations are pretty cheap in terms of perf I'm inclined to leave as-is

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: That's right -- are you looking for a change here?

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
iambmelt: Hah - I actually _had_ a guest account filter and then refactored it out. Easy to replace 👍

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
iambmelt: The Google Style Guide (which we follow for the Android code) doesn't _explicitly_ address this for interfaces, but the code style is generally to line-break/left-justify the closing braces. In this case since there are no braces to scope-on, I use the parens and apply the same style.

> [Line break before the closing brace.](https://google.github.io/styleguide/javaguide.html#s4.1.2-blocks-k-r-style)

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: Resolved in b3acd1b8a29842e7c7404364ac29242f6fa273bf

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
iambmelt: PS: Acknowledge this creates noise in the PR

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: I'll look at this. Will definitely add some supporting comments explaining what's going on

msal/src/main/java/com/microsoft/identity/client/Account.java
shoatman: I think that Account Id should always be home object id though.

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
iambmelt: Resolved in ad770bbde46de73922851e1ec8151d7152d2ffcf

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: Yep, it [currently] is

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: Yep - this is already implemented [here](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/iambmelt/tenant-profile-redux/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java#L212-L215)

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: I did some refactoring around this to better share code with `IDToken` in `common`.

Commits which address this can be found here:
- `common` -> [adf6d5bbe291da2df4b7c8fc5ff7157eb194dc13](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/461/commits/adf6d5bbe291da2df4b7c8fc5ff7157eb194dc13)
- `msal` -> ac88ab8a8fa651723d8fd5b884489d664c50c812

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
```diff
@@ -199,7 +199,7 @@ public boolean onNavigationItemSelected(final MenuItem item) {
             final Bundle bundle = new Bundle();
             if (mAuthResult != null) {
                 bundle.putString(ResultFragment.ACCESS_TOKEN, mAuthResult.getAccessToken());
-                bundle.putString(ResultFragment.DISPLAYABLE, (String) mAuthResult.getAccount().getClaims().get("preferred_username"));
+                bundle.putString(ResultFragment.DISPLAYABLE, (String) mAuthResult.getAccount().getClaims().get("preferred_username")); // TODO this doesn't work for guests with no home
```
iambmelt: TODO

iambmelt: Resolved in cdb93e4dbcf8c4fc00181e575aaa182888cbe784

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: Also improved the javadoc and nullability annots here: ac0b5714d96fe11bd7aa490621f75acf4a6b03d7

msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
```diff
@@ -239,11 +241,14 @@ public void onError(final Exception exception) {
...
+        // First, cast the input IAccount to a MultiTenantAccount
+        final MultiTenantAccount multiTenantAccount = (MultiTenantAccount) account;
+
         //create the parameter
         try {
-            if (null == account
-                    || null == account.getHomeAccountIdentifier()
-                    || StringUtil.isEmpty(account.getHomeAccountIdentifier().getIdentifier())) {
+            if (null == multiTenantAccount
+                    || null == multiTenantAccount.getHomeAccountId()) {
```
kreedula: Should we do `TextUtils.isEmpty()` check as well?

iambmelt: Sure -- I don't see a way we could end up with `""` as our value, but for consistency with other parts of the code that evaluate Strings I'll make this change

iambmelt: Oh, nevermind. I've made changes to assert that this can't be null. Resolved in e8899e5832120f1468de9c913d2084ad7b8101f1

PR #651, opened 2019-06-11
Add tenantProfile support in FLW flows

rpdome:
- Get back ICacheRecord lists from Broker
- Persists ICacheRecords of the current account in SharedPreference, and use that to keep track of the currently signed-in account. (#624)
- In BrokerMsalController.java, Extract boiler plate code into performBrokerTask().
- ISingleAccountPublicApplication now uses callback that supports exception (instead of just throwing it).

Flows are verified. Broker changes will follow.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
@@ -17,6 +17,7 @@ import com.microsoft.identity.common.internal.cache.SharedPreferencesFileManager
 import com.microsoft.identity.common.internal.controllers.TaskCompletedCallbackWithError;
 import com.microsoft.identity.common.internal.dto.AccountRecord;
 import com.microsoft.identity.common.internal.dto.IdTokenRecord;
+import com.microsoft.identity.common.internal.result.MsalBrokerResultAdapter;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -26,25 +27,43 @@ public class SingleAccountPublicClientApplication extends PublicClientApplicatio
     private static final String TAG = SingleAccountPublicClientApplication.class.getSimpleName();
 
     /**
-     * Name of the shared preference cache for storing current account.
-     * */
+     * Name of the shared preference cache for storing SingleAccountPublicClientApplication data.
+     */
     public static final String SINGLE_ACCOUNT_CREDENTIAL_SHARED_PREFERENCES =
             "com.microsoft.identity.client.single_account_credential_cache";
+
+    /**
+     * SharedPreference key for storing current account.
+     */
+    public static final String CURRENT_ACCOUNT_SHARED_PREFERENCE_KEY =
+            "com.microsoft.identity.client.single_account_credential_cache.current_account";
+
+    private SharedPreferencesFileManager sharedPreferencesFileManager;
 
     protected SingleAccountPublicClientApplication(@NonNull final Context context,
                                                    @NonNull final PublicClientApplicationConfiguration developerConfig) {
         super(context, developerConfig);
+        initializeSharedPreferenceFileManager(context);
     }
 
     protected SingleAccountPublicClientApplication(@NonNull final Context context,
                                                    @NonNull final String clientId) {
         super(context, clientId);
+        initializeSharedPreferenceFileManager(context);
     }
 
     protected SingleAccountPublicClientApplication(@NonNull final Context context,
                                                    @NonNull final String clientId,
                                                    @NonNull final String authority) {
         super(context, clientId, authority);
+        initializeSharedPreferenceFileManager(context);
+    }
+
+    private void initializeSharedPreferenceFileManager(@NonNull final Context context) {
+        sharedPreferencesFileManager = new SharedPreferencesFileManager(
+                context,
+                SINGLE_ACCOUNT_CREDENTIAL_SHARED_PREFERENCES,
+                new StorageHelper(context));
     }
 
     @Override
@@ -56,18 +75,18 @@ public class SingleAccountPublicClientApplication extends PublicClientApplicatio
                 new TaskCompletedCallbackWithError<List<ICacheRecord>, Exception>() {
                     @Override
                     public void onTaskCompleted(List<ICacheRecord> cacheRecords) {
-                        MultiTenantAccount currentAccount = getPersistedCurrentAccount(configuration.getAppContext());
+                        MultiTenantAccount currentAccount = getPersistedCurrentAccount();
                         MultiTenantAccount newAccount = getAccountFromICacheRecordList(cacheRecords);
 
                         String currentAccountHomeAccountId = currentAccount == null ? "" : currentAccount.getHomeAccountId();
-                        String newAccountHomeAccountId = newAccount == null ? "" :  newAccount.getHomeAccountId();
+                        String newAccountHomeAccountId = newAccount == null ? "" : newAccount.getHomeAccountId();
 
                         boolean isCurrentAccountChanged = !currentAccountHomeAccountId.equalsIgnoreCase(newAccountHomeAccountId);
-                        if (isCurrentAccountChanged){
-                            persistCurrentAccount(configuration.getAppContext(), cacheRecords);
+                        if (isCurrentAccountChanged) {
                             callback.onAccountChanged(currentAccount, newAccount);
                         }
 
+                        persistCurrentAccount(cacheRecords);
                         callback.onAccountLoaded(newAccount);
                     }
 
@@ -88,7 +107,7 @@ public class SingleAccountPublicClientApplication extends PublicClientApplicatio
                     @Override
                     public void onTaskCompleted(Boolean success) {
                         if (success) {
-                            persistCurrentAccount(configuration.getAppContext(), null);
+                            persistCurrentAccount(null);
                         }
 
                         callback.onTaskCompleted(success);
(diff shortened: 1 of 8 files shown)


msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -95,4 +100,114 @@ public void onError(Exception exception) {
...
+
+            accountCredentialCache.clearAll();
+            return null;
+        }
+
+        AccountRecord accountRecord = accountRecords.get(0);
+        List<IdTokenRecord> idTokenRecordList = msalOAuth2TokenCache.getIdTokensForAccountRecord(null, accountRecord);
+
+        List<ICacheRecord> cacheRecordList = new ArrayList<>();
+        for (final IdTokenRecord idTokenRecord : idTokenRecordList) {
```
iambmelt: 🐛 This is mixing `AccountRecords` from 1 tenant with `IdTokenRecords` from another -- because `AccountRecords` are _tenant specific_, there should be 1 `AccountRecord` + 1 `IdTokenRecord` _per tenant_.

The `realm` field of an `AccessTokenRecord` or `IdTokenRecord` tells you which tenant it belongs to

Some quick examples:
 #### An `IAccount` where the user is signed-in to only their home tenant...
- 1 `ICacheRecord` with 1 `AccountRecord` & 1 `IdTokenRecord` (both for the home tenant) inside

#### An `IAccount` where the user is signed-in to only a guest tenant...
- 1 `ICacheRecord` with 1 `AccountRecord` & 1 `IdTokenRecord` (both for the home tenant) inside

#### An `IAccount` where the user is signed-in to their home tenant and to a guest tenant...
- 2 `ICacheRecords`
    * 1 is for the home tenant, contains 1 `AccountRecord` + 1 `IdTokenRecord`
    * 1 is for the guest tenant, contains 1 `AccountRecord` + 1 `IdTokenRecord`

If a user were to sign-in to multiple guest tenants, then add additional `ICacheRecords` for each authorized tenant, with each containing an additional `AccessTokenRecord` + `IdTokenRecord`

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -95,4 +100,114 @@ public void onError(Exception exception) {
...
+    /**
+     * Get current account that is persisted in shared preference.
+     *
+     * @param appContext
+     * @return a persisted MultiTenantAccount. This could be null.
+     * */
+    @Nullable
+    private MultiTenantAccount getPersistedCurrentAccount(@NonNull final Context appContext) {
+        final String methodName = ":getPersistedCurrentAccount";
+        final IAccountCredentialCache accountCredentialCache = getCurrentAccountCredentialCache(appContext);
```
iambmelt: `IAccountCredentialCache` should really be _internal_ to the cache -- `PublicClientApplication` should call methods on `MsalOauth2TokenCache` to serve its needs

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -95,4 +100,114 @@ public void onError(Exception exception) {
...
+    /**
+     * Persists current account to shared preference.
+     *
+     * @param appContext
+     * @param cacheRecords list of cache record that belongs to an account.
+     *                     Please note that this layer will not verify if the list belongs to a single account or not.
+     * */
+    private void persistCurrentAccount(@NonNull final Context appContext,
+                                       @Nullable List<ICacheRecord> cacheRecords) {
+        final IAccountCredentialCache accountCredentialCache = getCurrentAccountCredentialCache(appContext);
```
iambmelt: Same here, `IAccountCredentialCache` should not be used directly -- if there are APIs you need here that aren't available on `MsalOauth2TokenCache` let me know and we can add additional APIs

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -95,4 +100,114 @@ public void onError(Exception exception) {
...
+    @Nullable
+    private MultiTenantAccount getPersistedCurrentAccount(@NonNull final Context appContext) {
+        final String methodName = ":getPersistedCurrentAccount";
+        final IAccountCredentialCache accountCredentialCache = getCurrentAccountCredentialCache(appContext);
+        final MsalOAuth2TokenCache msalOAuth2TokenCache = getTokenCache(appContext, accountCredentialCache);
+
+        List<AccountRecord> accountRecords = accountCredentialCache.getAccounts();
+        if (accountRecords.size() == 0) {
+            return null;
+        } else if (accountRecords.size() > 1) {
```
iambmelt: 🐛 Not true -- even in `SingleAccountMode` a single _user principle_ may have multiple `AccountRecords`.

See [this comment](https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/651/files#r292682883) below.

Let's say a user signs in to Teams (in `SingleAccountMode`/`SharedDeviceMode`) and that user is in `Contoso` tenant, if that user is a _guest_ in `Fabrikam`, they may authorize to that tenant as well, in which case, this user will have two `AccountRecords`, 2 `IdTokenRecords`, 2 `AccessTokenRecords` and 1 `RefreshTokenRecord`.

Use case here might a user who is FLW, but needs to chat with employees in multiple tenants at once

iambmelt: It's worth calling out that _all_ of a user's accounts will have the same `home_account_id`, irrespective of tenant. The `local_account_id` will change

iambmelt: For additional clarity on this point, see [`MsalOAuth2TokenCache#getAccountsWithAggregatedAccountData()`](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/iambmelt/tenant-profile/common/src/main/java/com/microsoft/identity/common/internal/cache/MsalOAuth2TokenCache.java#L621-L651) -- this method loads all of the `AccountRecords` & `IdTokenRecords` for a principal so as to later construct the full `IAccount` inside of [`AccountAdapter#adapt()`](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/iambmelt/tenant-profile-redux/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java#L143-L183)

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -132,28 +132,25 @@ private MultiTenantAccount getPersistedCurrentAccount(@NonNull final Context app
...
+                com.microsoft.identity.common.internal.logging.Logger.verbose(
+                        TAG + methodName,
+                        "Find " + idTokenRecordList.size() + " associated IdTokenRecord. Skip this accountRecord.");
+                continue;
+            }
+
             CacheRecord cacheRecord = new CacheRecord();
             cacheRecord.setAccount(accountRecord);
-            cacheRecord.setIdToken(idTokenRecord);
+            cacheRecord.setIdToken(idTokenRecordList.get(0));
```
iambmelt: You'll need to check the `IdTokenRecord` version here -- since the broker may contain v1 and v2 IdTokens (v1 used for ADAL, B2C but that doesn't apply here I wouldn't think)

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -26,25 +27,43 @@
...
 
     protected SingleAccountPublicClientApplication(@NonNull final Context context,
                                                    @NonNull final String clientId,
                                                    @NonNull final String authority) {
         super(context, clientId, authority);
+        initializeSharedPreferenceFileManager(context);
+    }
+
+    private void initializeSharedPreferenceFileManager(@NonNull final Context context) {
+        sharedPreferencesFileManager = new SharedPreferencesFileManager(
```
rpdome: @shoatman I'll be using this to stored the deserialized MultiTenantAccount. 

Regarding Synchronization, If we want to protect against race condition, we could potentially put the callback function of getCurrentAccount and removeCurrentAccount inside a block. IMHO that's kind of overkill. Please let me know if you think otherwise.

iambmelt: Synced offline - one concern here would be an app that calls `removeAccount()` and `getCurrentAccount()` simultaneously.

I would argue that this falls under a usage issue and doesn't warrant a great deal of bulletproofing in the MSAL-side, especially given the _other_ corner-cases we might face regarding `removeAccount()`.... for instance, 1 app calls `removeAccount()` while another app is mid token acquisition... Then we get into a larger discussion about globally synchronizing broker so that `removeAccount()` cannot be called while _any app_ is acquiring tokens. If we're trying to guard against strange race conditions this would seem to be a better angle of attack.

Let's sync in scrum; synchronizing `removeAccount()` in broker such that it can't be invoked while other apps are acquiringTokens might be useful but it would also invite other problems: for instance, an app starts interactive auth and then moves to the background. The lock could be auto-freed after 10 mins, but we can't make anyone wait that long to remove an account realistically.

The more I think about this, the more I think we should fire-off the actions as best-effort and retrofit more robust synchronization if actual, real-life issues arise.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -95,4 +119,65 @@ public void onError(Exception exception) {
...
                 });
     }
+
+    /**
+     * Get current account that is persisted in shared preference.
+     *
+     * @return a persisted MultiTenantAccount. This could be null.
+     */
+    @Nullable
+    private MultiTenantAccount getPersistedCurrentAccount() {
```
iambmelt: Ideally we might separate this functionality into another delegate class for `SingleAccountPublicClientApplication`, I'm fine to see this change in a follow-up PR -- persisting/restoring doesn't seem like it should be the direct responsibility of this class.

Issue #653, opened 2019-06-12
Remove Account - Potential Issue


shoatman:
Based on reading through the code.  I believe there is currently a bug in remove account specifically related to the scenario where:

- Account has been added locally (when no broker was installed on device)
- Broker is then installed
- removeAccount is called

It looks like in this scenario the code will attempt to remove the account from the broker cache rather than removing it from the local cache. 

https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/4e6c47303aac681c47ecc354fa50e17cd968bcde/msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java#L268-L276

heidijinxujia:
Yes, it is a bug. It should be 
```java
final RemoveAccountCommand command = new RemoveAccountCommand( 
         params, 
         MSALControllerFactory.getAcquireTokenSilentControllers( 
                 mPublicClientConfiguration.getAppContext(), 
                 params.getAuthority(), 
                 mPublicClientConfiguration 
         ), 
         callback 
 );
```

heidijinxujia:
@rpdome Do you need a similar change for FLW?

iambmelt:
@heidijinxujia @rpdome Also, will UI ever be needed for this `removeAccount` path? If yes, can't use silent controller, if not then no issue. I just remember there was some discussion about localizing messages around account removal

rpdome:
@heidijinxujia Given the current requirement (AuthApp must be installed and provisioned as shared BEFORE any other apps), FLW flow shouldn't use local cache at all. Although for extensibility, it might be a good idea to perform local cache clean up when the mode is switched (from multiple -> single) for the very first time.

@iambmelt  The current FLW removeAccount hasn't migrated to Command yet. It's still using the old pattern - directly call the function inside the controller, set up a connect and make a function call.

heidijinxujia:
The issue is fixed here 
https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/654

heidijinxujia:
Reopen this issue. Waiting for the approval for the PR.

PR #654, opened 2019-06-13
Fix the removeAccount for local cache when broker installed

heidijinxujia:
**Problem**
#653
The getAcquireTokenController only generator one controller out of local controller and broker controller. For getAccount, getAccounts and removeAccount call, essentially, we should load the account both locally and from broker as well.

**Solution**
Call MSALControllerFactory.getAcquireTokenSilentControllers() to construct the account command. Thus, when the broker eligible, it will trigger both local controller and broker controller. If not, it will only trigger the local controller. 

In this PR, I also update our msal test app to utilize the lab's test account. This change was blocked by the redirect uri registration.

**E2E**
Tested with both CP and Authenticator (ad-accounts RC26)
1. Install MSAL test app, login with cloud account A.
2. Click "GET USERS", it will show the info for account A.
3. Install CP or Authenticator
4. Open MSAL test app and login with a MAM account B, finish the registration and you should see the token returned.
5. Click "GET USERS", it will show the info for both Account A and Account B
6. Type in the loginhint for Account A, and click "REMOVE ACCOUNT", it shows "the account is removed successfully"
7. Click "GET USERS" it will show the account B only
8. Type in the loginhint for Account B, and click "REMOVE ACCOUNT", it shows "the account is removed successfully"
9. Click "REMOVE USER", it shows an empty list.


heidijinxujia:
@heidijinxujia This debug.keystore is from [our repo](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/testapps/testapp/debug/debug.keystore).

iambmelt:
Scenario where both controllers _should_ be hit:
- AppA signs in to MSAL without broker using account `user@contoso.com`
- AppB signs in to a MAM protected app using `user@contoso.com`, broker is installed
- AppA calls `removeAccount()` on `user@contoso.com`

kreedula:
Thanks @iambmelt , here is the code where we always set the result returned from last controller, which is Broker controller https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/controllers/RemoveAccountCommand.java#L67

heidijinxujia:
@iambmelt @kreedula The removeAccount will only return true when the account is removed successfully from local (and broker if the broker is eligible to use) or no account is found in local cache (and broker cache if broker is eligible). It will not return false, instead, it will throw the exception with error code and detailed reasons if anything goes wrong during the removing.

PR #655, opened 2019-06-13
SingleAccountPublicClientApplication Changes - Ready for Review

shoatman:
Hi Folks,

Please feel free to take a look at these changes.  I will rebase once the Additional Tenant Profile and SingleAccount storage changes are merged by Brian and Dome.  

Thanks

shane


msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
```diff
@@ -45,20 +47,78 @@
      * @param listener a callback to be invoked when the operation finishes.
```
rpdome: we should expose the isSharedDevice flag too.

shoatman: Done.

testapps/testapp/.settings/org.eclipse.buildship.core.prefs
```diff
@@ -0,0 +1,2 @@
+connection.project.dir=../../..
```
rpdome: Do you know what this file is?

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -66,10 +107,100 @@ private void initializeSharedPreferenceFileManager(@NonNull final Context contex
...
+
     @Override
-    public void getCurrentAccount(@NonNull final CurrentAccountCallback callback) {
+    public void getCurrentAccount(final CurrentAccountCallback callback) {
+        final String methodName = ":getCurrentAccount";
         final PublicClientApplicationConfiguration configuration = getConfiguration();
 
+
+        try {
+            if (MSALControllerFactory.brokerEligible(
```
rpdome: brokerEligible() is already in BrokerMsalController.getCurrentAccount

rpdome: (We might also want to rename this function and its .aidl counterpart to avoid confusion)

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -93,39 +224,168 @@ public void onTaskCompleted(List<ICacheRecord> cacheRecords) {
...
-                            persistCurrentAccount(null);
-                        }
 
-                        callback.onTaskCompleted(success);
-                    }
+        try{
+            if (MSALControllerFactory.brokerEligible(
+                    configuration.getAppContext(),
+                    configuration.getDefaultAuthority(),
+                    configuration) && mIsSharedDevice) {
```
rpdome: BrokerEligible() is already invoked in BrokerMsalController. removeAccountFromSharedDevice()

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -93,39 +224,168 @@ public void onTaskCompleted(List<ICacheRecord> cacheRecords) {
...
+                                mPublicClientConfiguration.getAppContext(),
+                                params.getAuthority(),
+                                mPublicClientConfiguration
+                        ),
+                        callback
+                );
+
+                ApiDispatcher.removeAccount(command);
+            } else {
+                callback.onError(new MsalClientException(MsalClientException.NO_ACCOUNT_TO_SIGN_OUT, "No account is currently signed in to your Single Account Public Client Application"));
```
rpdome: What would be the upgrade path for apps using ADAL who wants to switch to use single account mode?

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -93,39 +224,168 @@ public void onTaskCompleted(List<ICacheRecord> cacheRecords) {
...
+                params.setAccount(requestAccountRecord);
+
+                final RemoveAccountCommand command = new RemoveAccountCommand(
+                        params,
+                        MSALControllerFactory.getAcquireTokenController(
+                                mPublicClientConfiguration.getAppContext(),
+                                params.getAuthority(),
+                                mPublicClientConfiguration
+                        ),
+                        callback
```
rpdome: persistCurrentAccount(null) should be called as part of this callback as well.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -66,10 +107,100 @@ private void initializeSharedPreferenceFileManager(@NonNull final Context contex
...
+                                        "No account found.");
+                                checkCurrentAccountNotifyCallback(callback, null);
+                            } else {
+                                // First, transform the result into IAccount + TenantProfile form
+                                final List<IAccount>
+                                        accounts = AccountAdapter.adapt(result);
+
+                                final String trimmedIdentifier = getPersistedCurrentAccount().getHomeAccountId();
+
+                                // Evaluation precedence...
```
rpdome: Given that trimmedIdentifier is fixed in this case (always homeAccountId, unlike in MultipleAccountPCA.getAccount), we won't need to iterate through the list.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -66,10 +107,100 @@ private void initializeSharedPreferenceFileManager(@NonNull final Context contex
...
+                                final AccountMatcher accountMatcher = new AccountMatcher(
+                                        homeAccountMatcher,
+                                        localAccountMatcher,
+                                        usernameMatcher
+                                );
+
+                                for (final IAccount account : accounts) {
+                                    if (accountMatcher.matches(trimmedIdentifier, account)) {
+                                        //callback.onTaskCompleted(account);
+                                        checkCurrentAccountNotifyCallback(callback, result);
```
rpdome: This means that as an app, I can switch back and forth between SingleAccount and MultipleAccount mode?

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -93,39 +224,168 @@ public void onTaskCompleted(List<ICacheRecord> cacheRecords) {
                     @Override
                     public void onError(Exception exception) {
                         callback.onError(exception);
+
                     }
                 });
     }
 
+    private void checkCurrentAccountNotifyCallback(final CurrentAccountCallback callback, List<ICacheRecord> newAccountRecords){
```
rpdome: Either replace this with the code in getCurrentAccountFromSharedDevice's callback, or call this function over there.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -93,39 +224,168 @@ public void onTaskCompleted(List<ICacheRecord> cacheRecords) {
...
+        IAccount localAccount = getPersistedCurrentAccount();
+        IAccount newAccount = newAccountRecords == null ? null : getAccountFromICacheRecordList(newAccountRecords);
+
+        if (localAccount == null) {
```
rpdome: put this in didExistingCurrentAccountChange()

testapps/testapp/.settings/org.eclipse.buildship.core.prefs
shoatman: I think this needs to be added to git ignore.  Not sure where it came from.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -180,4 +440,34 @@ private MultiTenantAccount getAccountFromICacheRecordList(@NonNull List<ICacheRe
 
         return (MultiTenantAccount) account.get(0);
     }
+
+    @Override
+    public void acquireTokenSilentAsync(@NonNull final String[] scopes,
+                                        @NonNull final AuthenticationCallback callback) {
+        acquireTokenSilent(
+                scopes,
+                getPersistedCurrentAccount(),
```
rpdome: Are we planning to support ADAL - > Single Account MSAL migration?

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -66,10 +107,100 @@ private void initializeSharedPreferenceFileManager(@NonNull final Context contex
...
+                                com.microsoft.identity.common.internal.logging.Logger.verbose(
+                                        TAG + methodName,
+                                        "No account found.");
+                                checkCurrentAccountNotifyCallback(callback, null);
+                            } else {
+                                // First, transform the result into IAccount + TenantProfile form
+                                final List<IAccount>
+                                        accounts = AccountAdapter.adapt(result);
+
+                                final String trimmedIdentifier = getPersistedCurrentAccount().getHomeAccountId();
```
rpdome: getPersistedCurrentAccount() could be nil.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
rpdome: If yes..

For non-FLW, single account mode, I'm expecting that the user should not have to call getCurrentAccount() at all, so we could try to do one-time getCurrentAccount() here - for example.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -231,15 +373,220 @@ public static IMultipleAccountPublicClientApplication createMultipleAccountPubli
...
+            //Exception thrown
+            MsalException ex = result.getException();
+            throw ex;
+        }
+
+        IPublicClientApplication application = result.getPublicClientApplication();
+        if(application instanceof ISingleAccountPublicClientApplication){
+            return (ISingleAccountPublicClientApplication) application;
+        }else{
+            if(configuration.mAccountMode != AccountMode.SINGLE){
```
rpdome: Should this be part of early-return?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -201,21 +214,150 @@ public static void create(@NonNull final Context context,
...
+    @WorkerThread
     public static IMultipleAccountPublicClientApplication createMultipleAccountPublicClientApplication(@NonNull final Context context,
-                                                                                                       final int configFileResourceId) {
+                                                                                                       final int configFileResourceId) throws MsalException, InterruptedException {
         if (context == null) {
             throw new IllegalArgumentException("context is null.");
         }
 
-        return new MultipleAccountPublicClientApplication(context, loadConfiguration(context, configFileResourceId));
+        throwOnMainThread("createMultipleAccountPublicClientApplication");
```
rpdome: nit: We could probably consolidate these boilerplate code.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: Here's i'm opting between shared mode and non shared mode.  I suppose I can assume that if the device is in shared mode that the broker is installed.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: That was intended.  Let's discuss why you were thinking that though.  I'm guessing I missed something.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: Let's discuss this one in person

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: Agreed.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: Agreed.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: Agreed.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: Fixed.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: Agreed.  Fixed.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: I didn't follow this comment.  Let's discuss in person.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: Agreed.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: Agreed.  Fixed.

msal/src/main/java/com/microsoft/identity/client/IMultipleAccountPublicClientApplication.java
```diff
@@ -64,4 +86,43 @@ void getAccount(@NonNull final String identifier,
...
     );
+
+    /**
+     * Perform acquire token silent call. If there is a valid access token in the cache, the sdk will return the access token; If
+     * no valid access token exists, the sdk will try to find a refresh token and use the refresh token to get a new access token. If refresh token does not exist
+     * or it fails the refresh, exception will be sent back via callback.
+     *
+     * @param scopes   The non-null array of scopes to be requested for the access token.
+     *                 MSAL always sends the scopes 'openid profile offline_access'.  Do not include any of these scopes in the scope parameter.
+     * @param account  {@link IAccount} represents the account to silently request tokens.
```
Om83: nit : to silently request tokens for.

msal/src/main/java/com/microsoft/identity/client/IMultipleAccountPublicClientApplication.java
```diff
@@ -20,11 +20,33 @@
 //  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 //  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 //  THE SOFTWARE.
-
+//  Copyright (c) Microsoft Corporation.
```
iambmelt: Nit: license included twice

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
```diff
@@ -292,4 +118,11 @@ void acquireTokenSilentAsync(@NonNull final String[] scopes,
      * @return
      */
     PublicClientApplicationConfiguration getConfiguration();
+
+    /**
+     * Returns whether the application is being run on a device that is marked as a shared.
+     * Only SingleAccountPublicClientApplications may be used on shared devices
+     * @return
+     */
+    Boolean getIsSharedDevice();
```
iambmelt: Regarding naming, I'd be fine with this being `isSharedDevice()` rather than an explicit getter

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -201,21 +214,95 @@ public static void create(@NonNull final Context context,
...
+        if(application instanceof IMultipleAccountPublicClientApplication){
+            return (IMultipleAccountPublicClientApplication)application;
+        }else{
+            if(configuration.mAccountMode == AccountMode.MULTIPLE && application.getIsSharedDevice())  {
+                throw new MsalClientException("AccountMode in configuration is set to multiple; however the device is marked as shared");
+            }
+            if(configuration.mAccountMode != AccountMode.MULTIPLE){
+                throw new MsalClientException("AccountMode in configuration is not set to multiple");
+            }
+            return null;
```
iambmelt: Should this be another exception instead of returning `null`?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -231,15 +318,168 @@ public static IMultipleAccountPublicClientApplication createMultipleAccountPubli
...
```
iambmelt: Same question here - feels like it needs a fall-through error

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -66,66 +107,270 @@ private void initializeSharedPreferenceFileManager(@NonNull final Context contex
...
+            @Override
+            public void onCancel() {
+                authenticationCallback.onCancel();
+            }
+        };
+    }
+
+    private boolean didExistingCurrentAccountChange(MultiTenantAccount newAccount){
+        MultiTenantAccount persistedAccount = getPersistedCurrentAccount();
+        return (persistedAccount != null || newAccount.getId() != persistedAccount.getId()) ? true : false;
```
iambmelt: 🐞 String comparison using `!=` -- should be `... || !newAccount.getId().equals(persistedAccount.getId())`

But I think there's another issue here - if `persistedAccount != null` the right-side of this expression will never be evaluated. Should this be `if (null == persistedAccount)` ?

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
iambmelt: I would expect `getCurrentAccount()` to be called when broker is in-play (as, in the case of FoCI, another app could remove the account) -- since apps won't know that the broker is necessarily being used or not, I would say that developers ought to call `getCurrentAccount()` before trying to authenticate a user

msal/src/main/java/com/microsoft/identity/client/internal/CreateApplicationResult.java
```diff
@@ -0,0 +1,32 @@
+package com.microsoft.identity.client.internal;
```
iambmelt: License info

msal/src/main/java/com/microsoft/identity/client/internal/CreateApplicationResult.java
```diff
@@ -0,0 +1,32 @@
...
+import com.microsoft.identity.client.IPublicClientApplication;
+import com.microsoft.identity.client.exception.MsalException;
+
+public class CreateApplicationResult  {
+
+    private MsalException mMsalException;
+    private IPublicClientApplication mIPublicClientApplication;
+    private Boolean mSuccess = false;
+
+    public CreateApplicationResult(IPublicClientApplication application, MsalException exception){
```
iambmelt: Could apply `@NonNull`, `@Nullable` annots - not a functional change

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
rpdome: For some reason, I was under an impression that we were planning to 'clean the slate' every time the user switches mode. I'm also fine with this thought as it allows existing apps to switch to single account mode.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
rpdome: Oh. It's just that they are doing the same operation.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -66,66 +107,270 @@ private void initializeSharedPreferenceFileManager(@NonNull final Context contex
...
-                            persistCurrentAccount(null);
+
+        try{
+            if (mIsSharedDevice) {
+                removeAccountFromSharedDevice(callback, configuration);
+                return;
+            }
+
+
+            ApiDispatcher.initializeDiagnosticContext();
```
heidijinxujia: Need to remove this call. All the initializeDiagnosticContext() is now called before the commands excution at inside ApiDispatcher. Duplicated initializeDiagnosticContext() calls will result multiple correlationId for one request.

msal/src/main/java/com/microsoft/identity/client/IMultipleAccountPublicClientApplication.java
shoatman: fixed.

msal/src/main/java/com/microsoft/identity/client/IMultipleAccountPublicClientApplication.java
shoatman: fixed.

msal/src/main/java/com/microsoft/identity/client/internal/CreateApplicationResult.java
shoatman: Fixed

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
heidijinxujia: I assume the correlationId initialization should either happen at the PCA level or the ApiDispatcher level.

msal/src/main/java/com/microsoft/identity/client/internal/CreateApplicationResult.java
shoatman: fixed.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -201,21 +214,95 @@ public static void create(@NonNull final Context context,
...
+     *                             <p>
+     *                             For more information on the schema of the MSAL config json please
+     * @param listener             a callback to be invoked when the object is successfully created.
+     * @see <a href="https://developer.android.com/guide/topics/resources/providing-resources">Android app resource overview</a>
+     * @see <a href="https://github.com/AzureAD/microsoft-authentication-library-for-android/wiki">MSAL Github Wiki</a>
+     */
+    public static void createMultipleAccountPublicClientApplication(@NonNull final Context context,
+                              final int configFileResourceId,
+                              @NonNull final ApplicationCreatedListener listener) {
+        if (context == null) {
```
kreedula: Since `ApplicationCreatedListener` is required as well, we can throw an `IllegalArgumentException` if null

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
kreedula: +1

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -231,15 +318,168 @@ public static IMultipleAccountPublicClientApplication createMultipleAccountPubli
      * @see <a href="https://developer.android.com/guide/topics/resources/providing-resources">Android app resource overview</a>
      * @see <a href="https://github.com/AzureAD/microsoft-authentication-library-for-android/wiki">MSAL Github Wiki</a>
      */
+    @WorkerThread
     public static IMultipleAccountPublicClientApplication createMultipleAccountPublicClientApplication(@NonNull final Context context,
-                                                                                                       final File configFile) {
+                                                                                                       @NonNull final File configFile) throws InterruptedException, MsalException {
+
```
kreedula: Null check for `NonNull` variable `context` and `configFile` is missing.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
kreedula: +1

msal/src/main/java/com/microsoft/identity/client/internal/CreateApplicationResult.java
shoatman: fixed

msal/src/main/java/com/microsoft/identity/client/internal/CreateApplicationResult.java
shoatman: fixed.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -231,15 +318,168 @@ public static IMultipleAccountPublicClientApplication createMultipleAccountPubli
...
+        final PublicClientApplicationConfiguration configuration = loadConfiguration(context, configFileResourceId);
+
+        IPublicClientApplication application = createPublicClientApplication(context, configuration);
+        if(application instanceof ISingleAccountPublicClientApplication){
+            return (ISingleAccountPublicClientApplication) application;
+        }else{
+            if(configuration.mAccountMode != AccountMode.SINGLE){
+                throw new MsalClientException("AccountMode in configuration is not set to single");
+            }
+            return null;
```
kreedula: Here as well, better to throw an exception

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -231,15 +318,168 @@ public static IMultipleAccountPublicClientApplication createMultipleAccountPubli
...
+    @WorkerThread
+    public static ISingleAccountPublicClientApplication createSingleAccountPublicClientApplication(@NonNull final Context context,
+                                                                                                       final int configFileResourceId) throws InterruptedException, MsalException {
+        if (context == null) {
+            throw new IllegalArgumentException("context is null.");
+        }
+
+        final PublicClientApplicationConfiguration configuration = loadConfiguration(context, configFileResourceId);
+
+        IPublicClientApplication application = createPublicClientApplication(context, configuration);
```
kreedula: nit:final

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -846,20 +999,22 @@ private void acquireTokenSilent(@NonNull final String[] scopes,
 
         validateSilentParameters(acquireTokenSilentParameters);
 
-        acquireTokenSilentAsync(acquireTokenSilentParameters);
+        acquireTokenSilent(acquireTokenSilentParameters);
     }
 
     private void validateSilentParameters(
             @NonNull final AcquireTokenSilentParameters acquireTokenSilentParameters) {
+        /*
```
heidijinxujia: Confused about the commented out code here. For the AcquireTokenSilentParameters, should we validate the authority and callback?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -476,6 +736,11 @@ public PublicClientApplicationConfiguration getConfiguration() {
         return mPublicClientConfiguration;
     }
 
+    @Override
+    public Boolean getIsSharedDevice() {
```
kreedula: How about renaming to : `isSharedDevice` ?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -672,10 +842,12 @@ private void acquireToken(@NonNull final Activity activity,
...
                 .build();
 
-        acquireTokenAsync(acquireTokenParameters);
+
+
+        acquireToken(acquireTokenParameters);
     }
 
-    private static void validateNonNullArgument(@Nullable Object o,
+      protected static void validateNonNullArgument(@Nullable Object o,
```
kreedula: May this method can be used for all the `context` null checks above introduced

msal/src/main/java/com/microsoft/identity/client/internal/CreateApplicationResult.java
kreedula: +1, also `final`

msal/src/main/java/com/microsoft/identity/client/internal/CreateApplicationResult.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public IPublicClientApplication getPublicClientApplication(){
+        return mIPublicClientApplication;
+    }
+
+    public MsalException getException(){
+        return mMsalException;
+    }
+
+    public Boolean getSuccess(){
```
kreedula: nit : we can use primitive type `boolean` here to be safe of null pointers

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: Makes sense to me.  I copied this code from multiaccountpca... so we likely need to update there as well.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -66,66 +107,270 @@ private void initializeSharedPreferenceFileManager(@NonNull final Context contex
...
+                                if(currentAccount != null) {
+                                    final String trimmedIdentifier = currentAccount.getHomeAccountId();
+
+                                    final AccountMatcher accountMatcher = new AccountMatcher(
+                                            homeAccountMatcher
+                                    );
+
+                                    for (final IAccount account : accounts) {
+                                        if (accountMatcher.matches(trimmedIdentifier, account)) {
+                                            //callback.onTaskCompleted(account);
```
kreedula: nit :  can this commented out line be removed?

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: removed.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -66,66 +107,270 @@ private void initializeSharedPreferenceFileManager(@NonNull final Context contex
...
 
+    private void checkCurrentAccountNotifyCallback(final CurrentAccountCallback callback, List<ICacheRecord> newAccountRecords){
+        IAccount localAccount = getPersistedCurrentAccount();
+        IAccount newAccount = newAccountRecords == null ? null : getAccountFromICacheRecordList(newAccountRecords);
+
+        if (localAccount == null) {
+            if (newAccount != null) {
+                callback.onAccountChanged(null, newAccount);
+            }
+        } else if (localAccount.getId() != newAccount.getId()) {
```
kreedula: This should be `!localAccount.getId().equals(newAccount.getId()))` since it's string comparison

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: fixed.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: fixed.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -66,66 +107,270 @@ private void initializeSharedPreferenceFileManager(@NonNull final Context contex
...
+            }
+
+            @Override
+            public void onCancel() {
+                authenticationCallback.onCancel();
+            }
+        };
+    }
+
+    private boolean didExistingCurrentAccountChange(MultiTenantAccount newAccount){
```
kreedula: Nit : `final` and `NonNull` annotation

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: Fixed.

shoatman: Was actually missing the null check for newAccount as well.  Updated signature to mark as nullable since i'm now reusing this logic from somewhere else.

msal/src/main/java/com/microsoft/identity/client/IMultipleAccountPublicClientApplication.java
```diff
@@ -64,4 +86,43 @@ void getAccount(@NonNull final String identifier,
...
+     *
+     * @param scopes   The non-null array of scopes to be requested for the access token.
+     *                 MSAL always sends the scopes 'openid profile offline_access'.  Do not include any of these scopes in the scope parameter.
+     * @param account  {@link IAccount} represents the account to silently request tokens.
+     * @param callback {@link AuthenticationCallback} that is used to send the result back. The success result will be
+     *                 sent back via {@link AuthenticationCallback#onSuccess(IAuthenticationResult)}.
+     *                 Failure case will be sent back via {
+     * @link AuthenticationCallback#onError(MsalException)}.
+     */
+    void acquireTokenSilentAsync(@NonNull final String[] scopes,
```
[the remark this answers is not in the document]
shoatman: No we agreed to keep the convenience overloads... that had less than 4 arguments.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: Hmm... curious when we're allowed to deviate from standardard getter/setter... is there a guide we are following?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: Makes sense

msal/src/main/java/com/microsoft/identity/client/internal/CreateApplicationResult.java
shoatman: Yes

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: Fixed.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: Fixed.  Ended up being @Nullable as a result of another change

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: Fixed.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -153,6 +162,8 @@
      */
     private static final String FORCE_HOME_LOOKUP = "force_home_lookup";
 
+    protected Boolean mIsSharedDevice;
```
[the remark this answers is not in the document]
shoatman: Fixed.

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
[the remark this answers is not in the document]
shoatman: fixed.

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
```diff
@@ -28,13 +28,11 @@
...
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
-import android.util.Pair;
 
 import com.microsoft.identity.client.exception.MsalException;
 
-import java.util.List;
-
 public interface IPublicClientApplication {
+
```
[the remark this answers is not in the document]
shoatman: Done.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: Made the change anyway....

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -221,15 +302,177 @@ public static IMultipleAccountPublicClientApplication createMultipleAccountPubli
...
+        validateNonNullArgument(context, "context");
+
+        final PublicClientApplicationConfiguration configuration = loadConfiguration(context, configFileResourceId);
+
+        IPublicClientApplication application = createPublicClientApplication(context, configuration);
+        if(application instanceof ISingleAccountPublicClientApplication){
+            return (ISingleAccountPublicClientApplication) application;
+        }else{
+            if(configuration.mAccountMode != AccountMode.SINGLE){
+                throw new MsalClientException("AccountMode in configuration is not set to single");
```
rpdome: nit: move this above createPublicClientApplication()

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
rpdome: nit: move this above createPublicClientApplication()

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -777,10 +949,19 @@ private String getRequestTenantId(@NonNull final PublicClientApplicationConfigur
         }
 
         return result;
+
+    }
+
+    protected void validateAcquireTokenParameters(AcquireTokenParameters parameters){
+        return;
```
rpdome: nit: TODO?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -777,10 +949,19 @@ private String getRequestTenantId(@NonNull final PublicClientApplicationConfigur
...
+    }
+
+    protected void validateAcquireTokenSilentParameters(AcquireTokenSilentParameters parameters){
+        return;
```
rpdome: nit: TODO

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -66,66 +107,273 @@ private void initializeSharedPreferenceFileManager(@NonNull final Context contex
...
+        MultiTenantAccount persistedAccount = getPersistedCurrentAccount();
+
+        if(persistedAccount == null){
+            if(newAccount == null){
+                return false;
+            }else{
+                return true;
+            }
+        }else{
+            if(persistedAccount.getId().equals(newAccount.getId())){
```
rpdome: nit: is Id case-sensitive?
If not, I'd prefer this to be equalIgnoreCase.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: Fixed.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: Not sure.  This is someone elses add... will leave as is for now.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: leaving as is... see above.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
shoatman: Good question... I'll ignore case... these are guids typically and I don't think they are case sensistive
Will fix.

shoatman:
Thanks everyone!  I know this was a large PR.....

Issue #656, opened 2019-06-17
Implement support for TokenShareLibrary (TSL)


iambmelt:
Implementation should mirror/adapt ADAL implementation as necessary.

iambmelt:
Work is complete

PR #657, opened 2019-06-17
TSL Integration

iambmelt:
See https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/656

Expands: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/529

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -36,6 +36,7 @@ import android.util.Pair;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.microsoft.identity.client.claims.ClaimsRequest;
+import com.microsoft.identity.client.exception.MsalClientException;
 import com.microsoft.identity.client.exception.MsalException;
 import com.microsoft.identity.client.internal.MsalUtils;
 import com.microsoft.identity.client.internal.configuration.LogLevelDeserializer;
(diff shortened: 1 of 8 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -433,6 +432,50 @@ private void initializeApplication() {
...
     }
 
+    @Override
+    public String getWrappedFamilyRefreshToken(@NonNull final String identifier) throws MsalClientException {
+        validateNonNullArgument(identifier, "identifier");
+        validateBrokerNotInUse();
+
+        try {
+            return mTokenShareUtility.getWrappedFamilyRefreshToken(identifier);
+        } catch (Exception e) {
```
heidijinxujia: Q: why not catch BaseException instead? getWrappedFamilyRefreshToken is thrown BaseException. Exception might be broad..

heidijinxujia: And nit: final

iambmelt: I chose not to use `BaseException` because `Future.get()` throws numerous exceptions (docs [here](https://developer.android.com/reference/java/util/concurrent/Future#get())). If I catch these exceptions and rethrow them and catch them again [here](https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/657/files#diff-490b9a6859fc9f0cff8b15e9467b39b1R443) then the developer would have to call `msalException.getCause().getCause()` in order to see the original `Exception` (the first exception is the `MsalException`, the second exception is the wrapped `BaseException`, the third is the original error) -- ideally, the developer only calls this once and always gets the originating exception. I'm open to ideas about how to tidy this up, but just changing the exception type creates this problem.

I'll update to be `final` -- but I don't have a good resolution to your first point right now

iambmelt: Partially resolved in b0fa950087e7b3c4c85ee192a9778dc784951097

PR #666, opened 2019-06-21
Telemetry Integration

heidijinxujia:
Wire up the telemetry configuration with json.
- json file
- initilization

TODO
Add events into the calls.


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -380,6 +383,26 @@ protected PublicClientApplication(@NonNull final Context context,
...
 
+    private void setupTelemetry(@NonNull final Context context,
+                                @Nullable final PublicClientApplicationConfiguration developerConfig) {
+        if (null != developerConfig.getTelemetryConfiguration()) {
+            com.microsoft.identity.common.internal.logging.Logger.verbose(
+                    TAG,
+                    "Telemetry configuration is set. Telemetry is enabled."
+            );
+        } else {
+            com.microsoft.identity.common.internal.logging.Logger.verbose(
```
rpdome: what would happen if we're calling the constructor above with configuration as null?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -380,6 +383,26 @@ protected PublicClientApplication(@NonNull final Context context,
         }
     }
 
+    private void setupTelemetry(@NonNull final Context context,
+                                @Nullable final PublicClientApplicationConfiguration developerConfig) {
+        if (null != developerConfig.getTelemetryConfiguration()) {
```
kreedula: `developerConfig` is marked `Nullable` , needs null check for it as well.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -380,6 +383,26 @@ protected PublicClientApplication(@NonNull final Context context,
...
+                    "Telemetry configuration is set. Telemetry is enabled."
+            );
+        } else {
+            com.microsoft.identity.common.internal.logging.Logger.verbose(
+                    TAG,
+                    "Telemetry configuration is null. Telemetry is disabled."
+            );
+        }
+
+        new com.microsoft.identity.common.internal.telemetry.Telemetry.Builder()
```
kreedula: shouldn't this be inside the if loop above where we check for `null != developerConfig.getTelemetryConfiguration()`

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
heidijinxujia: If the telemetry configuration is null, then the Telemetry.isTelemetryEnabled will be set to "false", meaning that the Telemetry is disabled for this client.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
heidijinxujia: developerConfig.getTelemetryConfiguration() is a nullable value here. If the developerConfig.getTelemetryConfiguration() is null means that the client does not set up the Telemetry configuration. In this case, the Telemetry will be disabled for this client.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -628,6 +628,9 @@ public void onError(MsalException exception) {
     protected PublicClientApplication(@NonNull final Context context,
                                       @Nullable final PublicClientApplicationConfiguration developerConfig) {
         setupConfiguration(context, developerConfig);
+
+        setupTelemetry(context, developerConfig);
```
heidijinxujia: Put it into setupConfiguration and rename `setupConfiguration `

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
heidijinxujia: Changed form @Nullable to @NonNull. Moved to into setupConfiguration and using mPublicClientConfiguration which is always nonnull.

heidijinxujia:
@kreedula Hey Krishna, let me know if you have any other reviews. Thanks!

PR #668, opened 2019-06-26
Sync API

shoatman:
Not ready for review.


joshfriend:
Looks fine to me, this seems functionally the same as what I suggested in #548

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -432,6 +517,8 @@ private MultiTenantAccount getAccountFromICacheRecordList(@NonNull List<ICacheRe
     @Override
     public void acquireTokenSilentAsync(@NonNull final String[] scopes,
                                         @NonNull final AuthenticationCallback callback) {
+
+        //TODO: Add null check for getPeristedAccount and throw exception if account not found
```
iambmelt: Something to resolve in this PR?

shoatman: I just did.  Thanks for reminding me.  I noticed this was missing when refactoring for sync.

msal/src/main/java/com/microsoft/identity/client/AuthenticationCallback.java
```diff
@@ -27,6 +27,7 @@
 import com.microsoft.identity.client.exception.MsalException;
 import com.microsoft.identity.client.exception.MsalServiceException;
 import com.microsoft.identity.client.exception.MsalUiRequiredException;
+import com.microsoft.identity.client.exception.MsalUserCancelException;
```
heidijinxujia: MsalUserCancelException seems not being used.

shoatman: Fixed.  Thanks Heidi.  I also fixed the test app at the same time.

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
```diff
@@ -111,4 +125,20 @@ void acquireToken(@NonNull final Activity activity,
...
+         * @param result the success result.
+         */
+        void onTaskCompleted(List<IAccount> result);
+
+        /**
+         * Called once exception thrown.
+         *
+         * @param exception
+         */
+        void onError(MsalException exception);
```
heidijinxujia: I am trying to compare the api of `LoadAccountsCallback`  with `CurrentAccountCallback`, which takes `Exception` class for onError call ```void onError(@NonNull final Exception exception);```.  Should it be consistency with the onError parameter's type for LoadAccountsCallback and CurrentAccountCallback? It could all take MsalException as para, where we can adapt the Exception into MsalException.

rpdome: I missed this. CurrentAccountCallback should also return MsalException.

msal/src/main/java/com/microsoft/identity/client/IMultipleAccountPublicClientApplication.java
```diff
@@ -42,7 +45,13 @@
...
      * @param callback The callback to notify once this action has finished.
      */
-    void getAccounts(@NonNull final PublicClientApplication.LoadAccountCallback callback);
+    void getAccounts(@NonNull final LoadAccountsCallback callback);
+
+    /**
+     * Returns a List of {@link IAccount} objects for which this application has RefreshTokens.
+     */
+    @WorkerThread
+    List<IAccount> getAccounts() throws InterruptedException, MsalException;
```
rpdome: Shouldn't we return only MsalException?

shoatman: I think in the case of sync calls to call InterruptedException... sort of a reminder that you're making a synchronous background call that could be interrupted; however very open wrapping if we agree that's appropriate.  I'm assuming MsalClintException of type InterruptedException.

msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
```diff
@@ -289,52 +332,116 @@ public void onError(final Exception exception) {
...
+            callback.onError(new MsalClientException(MsalClientException.INVALID_PARAMETER));
+            return;
+        }
+
+        final OperationParameters params = OperationParametersAdapter.createOperationParameters(mPublicClientConfiguration);
+
+        // TODO Clean this up, only the cache should make these records...
+        // The broker strips these properties out of this object to hit the cache
+        // Refactor this out...
+        final AccountRecord requestAccountRecord = new AccountRecord();
```
rpdome: @iambmelt do we have a function to extract AccountRecord from a MultiTenantAccount?

msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
```diff
@@ -289,52 +332,116 @@ public void onError(final Exception exception) {
...
+
         } catch (final MsalClientException e) {
             callback.onError(e);
         }
     }
 
+    @Override
+    public boolean removeAccount(@Nullable IAccount account) throws MsalException, InterruptedException {
+
+        final ResultFuture<RemoveAccountResult> future = new ResultFuture();
```
rpdome: AsyncResult?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1058,7 +1027,7 @@ private void validateSilentParameters(
     }
 
     @Override
-    public void acquireTokenSilent(
+    public void acquireTokenSilentAsync(
             @NonNull final AcquireTokenSilentParameters acquireTokenSilentParameters) {
         validateSilentParameters(acquireTokenSilentParameters);
```
rpdome: There's validateAcquireTokenSilentParameters() down there. Is this (or the other) redundant?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1110,6 +1079,43 @@ public void acquireTokenSilent(
         }
     }
 
+    @Override
+    public IAuthenticationResult acquireTokenSilent(@NonNull AcquireTokenSilentParameters acquireTokenSilentParameters) throws InterruptedException, MsalException {
+        if (acquireTokenSilentParameters.getCallback() != null) {
+            throw new IllegalArgumentException("Do not provide callback for synchronous methods");
```
rpdome: This is not listed in 'throws'

msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
shoatman: I didn't in this case since there is not result object.... I suppose I can use Boolean.  I'll update.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: I'll check.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: It's a runtime exception... not a checked exception... hence not in Throws.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -183,36 +181,90 @@ public void onError(final Exception exception) {
...
+
+            @Override
+            public void onAccountChanged(@Nullable IAccount priorAccount, @Nullable IAccount currentAccount) {
+                CurrentAccountResult currentAccountResult = new CurrentAccountResult(currentAccount, priorAccount, false);
+                future.setResult(new AsyncResult<CurrentAccountResult>(currentAccountResult, null));
+            }
+
+            @Override
+            public void onError(@NonNull Exception exception) {
+                //TODO: Need to talk to Dome about exception here rather than MsalException
```
rpdome: getCurrentAccountAsync() already returns MsalException. Once you change the param of this interface to MsalException, it should be good.

shoatman: Cool.  Thanks.  I'll update

heidijinxujia:
The authority URL malformed exception will be thrown. Cause by the code [here](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/2969a68d8ff9b621200628c1d65fc1a94c119f83/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalWrapper.java#L260).

msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
shoatman: Fixed.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: Going to follow up on this in a future PR.

PR #675, opened 2019-07-03
Adding support for MSA to TSL

iambmelt:
~Currently holding this PR until scope behavior for MSA is bulletproofed (in testing, observed behavior where unless MSA scopes _matched_ the provided scopes for the /authorize call, the request was yielding `400`)~

Issue is no longer reproducing without code changes

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/ITokenShare.java
+++ b/msal/src/main/java/com/microsoft/identity/client/ITokenShare.java
@@ -34,12 +34,24 @@ public interface ITokenShare extends ITokenShareInternal {
      * {@inheritDoc}
      */
     @Override
-    String getWrappedFamilyRefreshToken(String identifier) throws MsalClientException;
+    String getOrgIdFamilyRefreshToken(String identifier) throws MsalClientException;
 
     /**
      * {@inheritDoc}
      */
     @Override
-    void saveFamilyRefreshToken(String tokenCacheItemJson) throws MsalClientException;
+    void saveOrgIdFamilyRefreshToken(String ssoStateSerializerBlob) throws MsalClientException;
+
+    /**
+     * {@inheritDoc}
+     */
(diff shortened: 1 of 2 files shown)


msal/src/main/java/com/microsoft/identity/client/ITokenShare.java
```diff
@@ -34,12 +34,24 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    String getMsaFamilyRefreshToken(String identifier) throws MsalClientException;
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    void saveMsaFamilyRefreshToken(String refreshToken) throws MsalClientException;
```
heidijinxujia: Do we plan to add the support to remove the frt when removing account?

iambmelt: For the _current_ app, `removeAccount()` will clear the FRT from the app local cache.

Do you mean will we remove the FRT from _other_ apps we might have shared the token with through TSL? I'm not aware of any such requirement; for apps not integrated with broker I wouldn't expect such functionality to exist.

Lmk if that answers your question 😄

iambmelt: Synced offline for more context.

Basically, 'What should happen if I have two apps on the device integrated with TSL? `AppA` could share token with `AppB`. `AppA` removes account, `AppB` shares it back thereby not _really_ removing the account."

Brian's answer:
    Not sure what's expected here. Perhaps a new API is needed to signal to TSL 'remove tokens for any user with this id'. This method would need to accept an OID or a upn

shoatman: Let's discuss this internally.

PR #692, opened 2019-07-27
Integrate the telemetry events

heidijinxujia:
- Integrate api events for MSAL including,
1. Api.BROKER_ACQUIRE_TOKEN_INTERACTIVE
2. Api.BROKER_GET_ACCOUNTS
3. Api.BROKER_REMOVE_ACCOUNT
4. Api.GET_BROKER_DEVICE_MODE
5. Api.LOCAL_ACQUIRE_TOKEN_INTERACTIVE
6. Api.LOCAL_COMPLETE_ACQUIRE_TOKEN_INTERACTIVE
7. Api.LOCAL_ACQUIRE_TOKEN_SILENT
8. Api.LOCAL_GET_ACCOUNTS

- Integrate the broker events for BrokerAccountManagerStrategy and BrokerAuthServiceStrategy, including
1. getBrokerAuthorizationIntent
2. acquireTokenSilent
3. getBrokerAccounts
4. removeBrokerAccount

- Remove the duplicated initialization of diagnostic context. The diagnostic context is initialized at the beginning of the ApiDispatcher calls.

Related PR for common: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/565

## Work Items
- [x] Api Events
- [x] Ui Events
- [x]  Broker Events 
- [x] Http Events 
- [x] Cache Events
- [x] Mapping from MSAL aggragated telemetry to MATS


iambmelt:
@heidijinxujia Similar to https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/627, if there are known gaps in this PRs implementation, can you call them out? I suspect there flows which are unaccounted for

heidijinxujia:
@iambmelt Do you think it would be better if I listed all the flows this PR captured?

iambmelt:
> @iambmelt Do you think it would be better if I listed all the flows this PR captured?

Yeah, if that's easier 👍

testapps/testapp/src/main/res/raw/msal_config.json
```diff
@@ -13,7 +13,7 @@
     }
   ],
   "telemetry": {
-    "pii_enabled": true,
+    "pii_enabled": false,
```
iambmelt: Should this be configured in the default app config? Is this off by default?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAccountManagerStrategy.java
```diff
@@ -237,6 +246,14 @@ AcquireTokenResult acquireTokenSilent(final AcquireTokenSilentOperationParameter
                         e
                 );
 
+                Telemetry.emit(
+                        new BrokerEndEvent()
+                                .putAction(methodName)
+                                .isSuccessful(false)
+                                .putErrorCode(ErrorStrings.BROKER_REQUEST_CANCELLED)
+                                .putErrorDescription("OperationCanceledException thrown when talking to account manager. The broker request cancelled.")
```
rpdome: nit: create a string const for redundant err messages. Here and below.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerBaseStrategy.java
```diff
@@ -122,12 +126,16 @@ static AcquireTokenResult getAcquireTokenResult(@NonNull final Bundle resultBund
                     resultAdapter.authenticationResultFromBundle(resultBundle)
             );
 
+            Telemetry.emit(new BrokerEndEvent().isSuccessful(true));
```
rpdome: nit: I'd prefer calling "successful" event in the very same function that start() is called. (All entry and exit points of a telemetry event should be in the same function).

It might be redundant, but I think it's better in terms of readability/maintainability. (I know for sure by reading the function that I'm not missing any exit points).

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAuthServiceStrategy.java
```diff
@@ -141,17 +144,25 @@ AcquireTokenResult acquireTokenSilent(AcquireTokenSilentOperationParameters para
...
             //Do we want a time out here?
             service = future.get();
-        } catch (Exception e) {
+        } catch (final InterruptedException | ExecutionException e) {
+            Telemetry.emit(
+                    new BrokerEndEvent()
+                            .putAction(methodName)
+                            .isSuccessful(false)
+                            .putErrorCode(ErrorStrings.IO_ERROR)
+                            .putErrorDescription(e.getLocalizedMessage())
```
rpdome: In some cases, we're putting a generic constant string here. In this case, we're using the error message from the exception object. What's the reasoning behind where to put what?

(Personally I think putting exception message might be helpful, but I assume that we are putting generic message here to make sure we're not leaking PII).

testapps/testapp/src/main/res/raw/msal_config.json
heidijinxujia: I think we should disable pii by default even for test apps. Because this is a public repo on github. I will change this back for now. And you guys can have a follow-up tmr or before ga.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerBaseStrategy.java
heidijinxujia: Sorry, I think I do not understand here. The isSuccessful() is to represent if the call succeeded, not the emit() of event.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAuthServiceStrategy.java
heidijinxujia: This is aligned with ADAL and other platforms.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerBaseStrategy.java
rpdome: What I meant was that, if possible, every BrokerEndEvent() should be emitted in the same function as BrokerStartEvent().

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAuthServiceStrategy.java
rpdome: I'm seeing conflicting implementation here and in the line I commented above (that one uses a constant string), hence the question :)

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerBaseStrategy.java
heidijinxujia: No, it should not. The start event and the end event do not necessary in the same function. This is why we need an atomic function like Telemetry.emit(). Please refer to the usage of UiStartEvent and UiEndEvent.

rpdome: hence 'if possible'.

The reason I suggested putting them in the same function is to avoid missing 'exit points' when we're making code change. Having them in one place, it's easier to spot if we forgot to emit any event when we're returning a value, or throw an exception. Putting them in multiple place, it is more likely that we might miss that (which will be very bad if we're trying to get things like success rate from the flow, since the number would not add up).

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAccountManagerStrategy.java
heidijinxujia: Cool, thanks! Addressed. :)

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAuthServiceStrategy.java
heidijinxujia: Key.ERROR_DESCRIPTION is alwasy treated as oii and we filtered them out in both logging and telemetry when pii/oii is disabled.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerBaseStrategy.java
heidijinxujia: Thank you. Btw, most of the BrokerStartEvent and BrokerEndEvent are emit in the same function if you have a look at the code. The getAcquireTokenResult() in the parent class here is called by the BrokerAccountManagerStrategy and BrokerAuthServiceStrategy when extracting the result. We could either move the function back and duplicate the code or add additional try-catch in the calling method. I prefer the current implementation where to emit the result inside getAcquireTokenResult(). However, feel free to refactor the code in the future. :)

heidijinxujia:
@iambmelt  I listed the covered flows in the description. Is there any other comment you want me to address? Thanks!

iambmelt:
> @iambmelt I listed the covered flows in the description. Is there any other comment you want me to address? Thanks!

Thank you!

PR #693, opened 2019-07-29
Adds getUsername() accessor to IAccount, ITenantProfile

iambmelt:
Makes a couple of minor changes:
- Adds `getUsername()` to both `IAccount` and `ITenantProfile` (through shared superinterface `IClaimable`)
- Enforces `@NonNull` by eagerly initializing to `""` and overwriting if the respective IdToken claim is not-null
- Applies similar behavior to `getTenantId()` - formerly available only on `ITenantProfile` this is now surfaced on `IAccount` too through `IClaimable`
    * Similar nullability behavior applies

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/Account.java
+++ b/msal/src/main/java/com/microsoft/identity/client/Account.java
@@ -68,7 +68,8 @@ public class Account implements IAccount {
     }
 
     @NonNull
-    String getTenantId() {
+    @Override
+    public String getTenantId() {
         return mHomeTenantId;
     }
 
--- a/msal/src/main/java/com/microsoft/identity/client/Account.java
+++ b/msal/src/main/java/com/microsoft/identity/client/Account.java
@@ -97,4 +97,20 @@ public class Account implements IAccount {
     public Map<String, ?> getClaims() {
         return mIdTokenClaims;
     }
+
+    @NonNull
+    @Override
+    public String getUsername() {
+        String username = "";
+
+        if (null != getClaims()) {
+            final String usernameClaim = (String) getClaims().get(IDToken.PREFERRED_USERNAME);
+
+            if (null != usernameClaim) {
+                username = usernameClaim;
+            }
+        }
+
+        return username;
+    }
 }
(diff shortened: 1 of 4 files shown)


msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -97,4 +98,20 @@ String getEnvironment() {
     public Map<String, ?> getClaims() {
         return mIdTokenClaims;
     }
+
+    @NonNull
+    @Override
+    public String getUsername() {
```
rpdome: nit: should we add a comment stating that as of now, this field can be empty in B2C scenario?

msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -97,4 +98,20 @@ String getEnvironment() {
...
+        String username = "";
+
+        if (null != getClaims()) {
+            final String usernameClaim = (String) getClaims().get(IDToken.PREFERRED_USERNAME);
```
rpdome: Haven't tried it myself, but would this be converted to a string "null"?
https://stackoverflow.com/questions/18723596/no-exception-while-type-casting-with-a-null-in-java

rpdome: might be safer to do null check before casting.

shahzaibj: It looks like it will still be null 

https://repl.it/repls/FlakyEagerRar

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: Sure I'll add a comment

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: Good catch

iambmelt: Thanks @shahzaibj for checking - leaving as-is

msal/src/main/java/com/microsoft/identity/client/Account.java
iambmelt: Resolved in 07583c6db89fb6712643dd07845c05f4143777f3

iambmelt:
@rpdome Added comment as requested. @shahzaibj confirmed that casting a `null` String doesn't yield `"null"` but instead yields a _true_ `null`

PR #694, opened 2019-07-29
Mark authority as nonnull in silent call, move the validation code.

rpdome:


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1434,7 +1428,7 @@ boolean matches(@NonNull final String username,
     };
 
     protected IAuthenticationResult acquireTokenSilentSync(@NonNull final String[] scopes,
```
kreedula: Not related to this change but why is this `protected` ?

rpdome: This seems to be our internal function (i.e. the public facing interfaces IMulti/ISingle are invoking this, but we don't expose it directly).

PR #709, opened 2019-08-16
Clean up PCA

### LINKED ISSUE #703: [Back Compat] getDeviceMode() throws NPE when old broker installed
**Description**
getIntentForAuthService() returns null when trying to talk to older v1 broker
Which results with NPE when trying to connect the MicrosoftAuthService.

**Solution**
To fix this issue, we need to refactor the single account/FLW broker calls to align with ApiDispatcher and have the broker eligible checks as Multi accounts broker calls.

The telemetry events integration will also be benefit from this.

rpdome:
** Major changes **
- Extracts PCAConfigurationFactory class, and PCA is now taking PCAConfiguration as part of the parameter. 
    - This is because the factory method needs the configuration to determine if it should get shared device flag from Broker.
- If Broker is not enabled, MSAL will not get shared device flag and isSharedDevice will ALWAYS be set to false. (To fix #703 - we will not make any call to broker unless MSAL is eligible).
- getCurrentAccount now returns MSALException.
- Make Async SingleAccount/MultiAccount PCA to return with a callback interface that no longer requires casting. 

** Clean up PCA **
- Sort PCA factory methods and group them with #region (so that you can collapse them in AndroidStudio)
- Consolidate PCA constructor into one constructor (as one of them doesn't call the initialize function, figured it might be better do this to prevent this sort of issue in the future).
- Extract some helper functions to MsalUtils.

** Test cases update **
- Remove ignored E2E test cases in PublicClientApplication. (We might want to replace this with Robolectric at some point). @shahzaibj FYI.
- Merge constructorTest into PublicClientApplication test, as it requires Mockito to bypass certain stuffs in PCA constructor.


msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -348,35 +348,37 @@ public boolean removeAccount(@NonNull final OperationParameters parameters)
     /**
      * Get device mode from Broker.
      */
-    public void getBrokerDeviceMode(final Context appContext,
+    public void getBrokerDeviceMode(final PublicClientApplicationConfiguration configuration,
                                     final PublicClientApplication.BrokerDeviceModeCallback callback) {
+        final String methodName = ":getBrokerDeviceMode";
+
+        try {
```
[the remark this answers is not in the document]
rpdome: Then we make a request to broker and get the mode.
If the app doesn't specifically request Multiple account PCA, then we would use the mode to determine what PCA to return.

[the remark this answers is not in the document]
rpdome: It doesn't matter if the config file is set to single or multiple. If MSAL is broker-eligible, we make this call to get shared device flag.

Single Account PCA could be in 'shared device' and 'non-shared device'.
Multiple Account PCA, however, has to be in 'non-shared device' only.

But these will be determined in the PCA factory methods, with
1. the value this function returns
2. the mode set in config file

msal/src/androidTest/java/com/microsoft/identity/client/PublicClientApplicationTest.java
```diff
@@ -127,189 +103,177 @@ public void tearDown() {
...
+        }
+    }
+
+    @Test
+    public void testMultipleAccountAsyncConstructor() {
+        final Context context = new PublicClientApplicationTest.MockContext(mAppContext);
+        mockPackageManagerWithDefaultFlag(context);
+        mockHasCustomTabRedirect(context);
+
+        PublicClientApplication.createMultipleAccountPublicClientApplication(context, com.microsoft.identity.msal.test.R.raw.test_msal_config_multiple_account, new PublicClientApplication.IMultipleAccountApplicationCreatedListener() {
```
shahzaibj: I would suggest importing  `com.microsoft.identity.msal.test.R` at the top, and then in the method calls we can just do R.raw.filename. So this way our statements aren't too long and don't need to scroll horizontally. 

Same applies for all such occurrences in this file.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -632,71 +624,99 @@ public void onError(MsalException exception) {
...
     }
 
     private void initializeApplication() {
         final Context context = mPublicClientConfiguration.getAppContext();
+        setupTelemetry(context, mPublicClientConfiguration);
+
+        AzureActiveDirectory.setEnvironment(mPublicClientConfiguration.getEnvironment());
+        Authority.addKnownAuthorities(mPublicClientConfiguration.getAuthorities());
+
+        if (mPublicClientConfiguration.getOAuth2TokenCache() instanceof MsalOAuth2TokenCache) {
```
kreedula: Not part of this change but can we move this to helper method like `initializeTSL`

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfigurationFactory.java
```diff
@@ -0,0 +1,173 @@
...
+                TAG + methodName,
+                "Loading default configuration"
+        );
+        return loadConfiguration(context, R.raw.msal_default_config);
+    }
+
+    @VisibleForTesting
+    static PublicClientApplicationConfiguration loadConfiguration(@NonNull final Context context,
+                                                                  final int configResourceId) {
+        InputStream configStream = context.getResources().openRawResource(configResourceId);
```
kreedula: nit : final

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfigurationFactory.java
```diff
@@ -0,0 +1,173 @@
...
+    @VisibleForTesting
+    static PublicClientApplicationConfiguration loadConfiguration(@NonNull File configFile) {
+        try {
+            return loadConfiguration(new FileInputStream(configFile), false);
+        } catch (FileNotFoundException e) {
+            throw new IllegalArgumentException("Provided configuration file path=" + configFile.getPath() + " not found.");
+        }
+    }
+
+    private static PublicClientApplicationConfiguration loadConfiguration(InputStream configStream, boolean isDefaultConfiguration) {
```
kreedula: nit : `final` and `NonNull`

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -224,21 +184,6 @@ private void getCurrentAccountFromSharedDevice(@NonNull final CurrentAccountCall
         final String methodName = ":getCurrentAccountFromSharedDevice";
 
         //TODO: migrate to Command.
-        try {
```
kreedula: broker checks no longer needed ? are we doing it before?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -395,18 +443,47 @@ public void onError(MsalException exception) {
...
+     *                   {@link PublicClientApplication}. Cannot be null.
+     *                   <p>
+     *                   Note: The {@link Context} should be the application context instead of the running activity's context, which could potentially make the sdk hold a
+     *                   strong reference to the activity, thus preventing correct garbage collection and causing bugs.
+     *                   </p>
+     * @param configFile The file containing the JSON configuration for the PublicClientApplication
+     * @param listener   a callback to be invoked when the object is successfully created.
+     *                   <p>
+     *                   For more information on the schema of the MSAL config json please
+     * @param listener   a callback to be invoked when the object is successfully created.
```
iambmelt: duplicate `@param listener`

msal/src/androidTest/java/com/microsoft/identity/client/PublicClientApplicationTest.java
```diff
@@ -520,6 +360,22 @@ private void mockPackageManagerWithClientId(final Context context, final String
...
+            final ApplicationInfo applicationInfo = Mockito.mock(ApplicationInfo.class);
+            Mockito.when(context.getPackageManager().getApplicationInfo(mAppContext.getPackageName(), 0))
+                    .thenReturn(applicationInfo);
+
+            final PackageInfo packageInfo = Mockito.mock(PackageInfo.class);
+            Mockito.when(context.getPackageManager().getPackageInfo(mAppContext.getPackageName(), 0))
+                    .thenReturn(packageInfo);
+
+        } catch (Exception e) {
+            // Best effort.
```
iambmelt: Does this need a `fail()` call?

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
rpdome: I moved the check to getBrokerAccountMode instead, meaning that we can safely assume that whenever isSharedDevice = true, it can talk to broker.

If it can't talk to broker, the flag will always be set to false.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -632,71 +624,99 @@ public void onError(MsalException exception) {
...
-        AzureActiveDirectory.setEnvironment(mPublicClientConfiguration.getEnvironment());
-        Authority.addKnownAuthorities(mPublicClientConfiguration.getAuthorities());
+    private static void createMultipleAccountPublicClientApplication(@NonNull final PublicClientApplicationConfiguration configuration,
+                                                                     @NonNull final IMultipleAccountApplicationCreatedListener listener) {
+        create(configuration,
+                null,
+                null,
+                new ApplicationCreatedListener() {
+                    @Override
+                    public void onCreated(IPublicClientApplication application) {
```
iambmelt: Could consider adding `final` to callback params, here and elsewhere

rpdome:
Thanks @heidijinxujia @shahzaibj @kreedula @iambmelt !

Issue #704, opened 2019-08-07
MSAL crashes when the passed authority uses domain suffix instead of guid.


rpdome:
**Scenario**:
The customer is trying to pass the non-guid authority, (i.e. https://login.microsoftonline.com/testtenant.onmicrosoft.com) as the optional authority of the acquireTokenSilent call, and the app crashes.

**Root cause**:
In OperationParametersAdapter.createAcquireTokenSilentOperationParameters(), we have a check to decide if we want to use home or tenant profile. In this case, however, isAccountHomeTenant failed because that function was expecting 'tenantId' to be GUID, but was getting "testtenant.onmicrosoft.com" instead. Therefore, this function returns false and we're trying to use tenant profile, which was null and caused the crash.

iambmelt:
Note: Testing this fix must also verify B2C is working (incl. w/ `b2clogin.com` style urls)

/cc @negoe

iambmelt:
```
    --------- beginning of crash
2019-08-16 09:50:55.198 31997-31997/com.microsoft.identity.client.sample.local E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.microsoft.identity.client.sample.local, PID: 31997
    java.lang.NullPointerException: Attempt to invoke interface method 'java.util.Map com.microsoft.identity.client.ITenantProfile.getClaims()' on a null object reference
        at com.microsoft.identity.client.internal.controllers.OperationParametersAdapter.createAcquireTokenSilentOperationParameters(OperationParametersAdapter.java:292)
        at com.microsoft.identity.client.PublicClientApplication.acquireTokenSilentAsync(PublicClientApplication.java:1106)
        at com.microsoft.identity.client.testapp.MsalWrapper$3.onTaskCompleted(MsalWrapper.java:264)
        at com.microsoft.identity.client.MultipleAccountPublicClientApplication$4.onTaskCompleted(MultipleAccountPublicClientApplication.java:295)
        at com.microsoft.identity.client.MultipleAccountPublicClientApplication$4.onTaskCompleted(MultipleAccountPublicClientApplication.java:266)
        at com.microsoft.identity.common.internal.controllers.ApiDispatcher$1$2.run(ApiDispatcher.java:101)
        at android.os.Handler.handleCallback(Handler.java:790)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:164)
[...]

PR #710, opened 2019-08-16
Fixes resolution of AccountRecords used for AT/S calls when named tenants are used

iambmelt:
Exploring a fix for #704, this is cleanup related to that inquiry

# MSAL + TenantProfile Request Behavior

## Authority Types
- AAD
- B2C
---
### B2C Rules
- Always use the home account for this authority
---
### AAD Rules
- May use either the home or a `TenantProfile`
- Which profile is determined by the authorizing authority

### When to use home...
- If the authority is "home equivalent"
    * `/common`
    * `/consumers`
    * `/organizations`
- If the authority contains a GUID and that GUID is `==` to home account `tid`
- If the authority is named _and_ its GUID is `==` to home `tid`

### When to use a profile...
- If the authority contains a GUID and that GUID corresponds to a guest account
- If the authority is named _and_ its GUID is `==` to a guest `tid`
---
### Things you can take for granted...
- The `environment` of an `IAccount` is the same across profiles
- The `home_account_id` is the same across profiles

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -1165,7 +1165,7 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
 
             final boolean isUuid = isUuid(tenantIdNameOrAlias);
 
-            if (!isUuid) {
+            if (!isUuid && !isHomeTenantEquivalent(tenantIdNameOrAlias)) {
                 final OpenIdProviderConfiguration providerConfiguration =
                         loadOpenIdProviderConfigurationMetadata(requestAuthority);
 
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -27,10 +27,12 @@ import android.app.Activity;
 import android.content.Context;
 import android.content.pm.PackageManager;
 import android.os.Looper;
+
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 import androidx.annotation.WorkerThread;
+
 import android.text.TextUtils;
 import android.util.Pair;
 
@@ -859,25 +861,6 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
                 null, // authority
                 callback,
                 null, // loginHint
-                null // claimsRequest
-        );
-    }
-
-    @Override
-    public void acquireToken(@NonNull final Activity activity,
-                             @NonNull final String[] scopes,
-                             @Nullable final String loginHint,
-                             @NonNull final AuthenticationCallback callback) {
-        acquireToken(
-                activity,
-                scopes,
-                null, // account
-                null, // uiBehavior
-                null, // extraQueryParams
-                null, // extraScopes
-                null, // authority
-                callback,
-                loginHint,
                 null // claimsRequest
         );
     }
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -27,6 +27,7 @@ import android.app.Activity;
 import android.content.Context;
 import android.content.pm.PackageManager;
 import android.net.Uri;
+import android.os.Handler;
 import android.os.Looper;
 import android.text.TextUtils;
 import android.util.Pair;
@@ -92,6 +93,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 import static com.microsoft.identity.client.internal.controllers.OperationParametersAdapter.isAccountHomeTenant;
 import static com.microsoft.identity.client.internal.controllers.OperationParametersAdapter.isHomeTenantEquivalent;
@@ -147,10 +150,11 @@ import static com.microsoft.identity.common.exception.ClientException.TOKEN_SHAR
  * </p>
  */
 public class PublicClientApplication implements IPublicClientApplication, ITokenShare {
-    private static final String TAG = PublicClientApplication.class.getSimpleName();
 
+    private static final String TAG = PublicClientApplication.class.getSimpleName();
     private static final String INTERNET_PERMISSION = "android.permission.INTERNET";
     private static final String ACCESS_NETWORK_STATE_PERMISSION = "android.permission.ACCESS_NETWORK_STATE";
+    private static final ExecutorService sBackgroundExecutor = Executors.newCachedThreadPool();
 
     /**
      * Constant used to signal a home account's tenant id should be used when performing cache lookups
@@ -984,12 +988,11 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
 
     @Override
     public void acquireToken(@NonNull final AcquireTokenParameters acquireTokenParameters) {
-        final ILocalAuthenticationCallback localAuthenticationCallback =
-                getLocalAuthenticationCallback(acquireTokenParameters.getCallback());
-
-        new Thread(new Runnable() {
+        sBackgroundExecutor.submit(new Runnable() {
             @Override
             public void run() {
+                final ILocalAuthenticationCallback localAuthenticationCallback =
+                        getLocalAuthenticationCallback(acquireTokenParameters.getCallback());
                 try {
                     acquireTokenParameters.setAccountRecord(
                             selectAccountRecordForTokenRequest(
@@ -1019,10 +1022,17 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
 
                     ApiDispatcher.beginInteractive(command);
                 } catch (final BaseException exception) {
-                    localAuthenticationCallback.onError(exception);
+                    // If there is an Exception, post it to the main thread...
+                    final Handler handler = new Handler(Looper.getMainLooper());
+                    handler.post(new Runnable() {
+                        @Override
+                        public void run() {
+                            localAuthenticationCallback.onError(exception);
+                        }
+                    });
                 }
             }
-        }).start();
+        });
     }
 
     protected void acquireTokenSilent(@NonNull final String[] scopes,
@@ -1050,13 +1060,13 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
     @Override
     public void acquireTokenSilentAsync(
             @NonNull final AcquireTokenSilentParameters acquireTokenSilentParameters) {
-        final ILocalAuthenticationCallback callback = getLocalAuthenticationCallback(
-                acquireTokenSilentParameters.getCallback()
-        );
-
-        new Thread(new Runnable() {
+        sBackgroundExecutor.submit(new Runnable() {
             @Override
             public void run() {
+                final ILocalAuthenticationCallback callback = getLocalAuthenticationCallback(
+                        acquireTokenSilentParameters.getCallback()
+                );
+
                 try {
                     acquireTokenSilentParameters.setAccountRecord(
                             selectAccountRecordForTokenRequest(
@@ -1089,7 +1099,7 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
                     callback.onError(exception);
                 }
             }
-        }).start();
+        });
     }
 
 
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -1158,7 +1158,7 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
             // If the account is named like <tenant_name>.onmicrosoft.com we need to query the OpenId
             // Provider Configuration Metadata in order to get the tenant id. Once we have the
             // tenant id, we must then select the appropriate home or profile.
-            final String tenantIdNameOrAlias = aadAuthority.getAudience().getTenantId();
+            String tenantIdNameOrAlias = aadAuthority.getAudience().getTenantId();
 
             // The AccountRecord we'll use to request a token...
             final AccountRecord accountRecord = new AccountRecord();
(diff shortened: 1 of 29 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java
```diff
@@ -287,16 +310,57 @@ public static AcquireTokenSilentOperationParameters createAcquireTokenSilentOper
...
                     // Use that tenant's profile...
-                    final ITenantProfile tenantProfile = multiTenantAccount.getTenantProfiles().get(tenantId);
-
-                    validateClaimsExistForTenant(tenantId, tenantProfile.getClaims());
-
-                    requestAccountRecord.setUsername(
-                            SchemaUtil.getDisplayableId(tenantProfile.getClaims())
-                    );
-                    requestAccountRecord.setLocalAccountId(tenantProfile.getId());
+                    final ITenantProfile[] tenantProfile = {multiTenantAccount.getTenantProfiles().get(tenantId)};
```
iambmelt: Quality doesn't hit the bar here.

- Do not block UI thread
- Guidance should say "prefer using tenantId authority"

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -979,64 +988,51 @@ protected void validateAcquireTokenSilentParameters(AcquireTokenSilentParameters
...
-                        acquireTokenParameters.getAccount(),
-                        getRequestTenantId(
-                                mPublicClientConfiguration,
-                                acquireTokenParameters
-                        )
-                )
-        );
-
-        validateAcquireTokenParameters(acquireTokenParameters);
+        sBackgroundExecutor.submit(new Runnable() {
```
iambmelt: Didn't want to introduce this, but it's required for cases when an IAccount is passed-in and a specific `AccountRecord` needs to be authorized - when broker is used, we have to synthesize the account that is passed across the boundary

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1064,51 +1060,178 @@ protected void acquireTokenSilent(@NonNull final String[] scopes,
...
+            accountRecord.setHomeAccountId(multiTenantAccount.getHomeAccountId());
+
+            if (isHomeTenantEquivalent(tenantIdNameOrAlias)
+                    || isAccountHomeTenant(multiTenantAccount.getClaims(), tenantIdNameOrAlias)) {
+                // Use home...
+                validateClaimsExistForTenant(tenantIdNameOrAlias, multiTenantAccount);
+
+                accountRecord.setLocalAccountId(multiTenantAccount.getId());
+                accountRecord.setUsername(
+                        SchemaUtil.getDisplayableId(multiTenantAccount.getClaims())
```
kreedula: See if we can use `username` from Account

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1064,51 +1060,178 @@ protected void acquireTokenSilent(@NonNull final String[] scopes,
...
+                final ITenantProfile profileForRequest =
+                        multiTenantAccount
+                                .getTenantProfiles()
+                                .get(tenantIdNameOrAlias);
+
+                validateClaimsExistForTenant(tenantIdNameOrAlias, profileForRequest);
+
+                accountRecord.setLocalAccountId(profileForRequest.getId());
+                accountRecord.setUsername(
+                        SchemaUtil.getDisplayableId(profileForRequest.getClaims())
```
kreedula: Consider if we directly use `username`  directly from `ITenantProfile`

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1064,51 +1060,178 @@ protected void acquireTokenSilent(@NonNull final String[] scopes,
...
+                final List<String> paths = issuerUri.getPathSegments();
+                final String tenantPath = paths.get(0);
+
+                if (multiTenantAccount.getTenantId().equals(tenantPath)) {
+                    // use home...
+                    validateClaimsExistForTenant(tenantPath, multiTenantAccount);
+
+                    accountRecord.setLocalAccountId(multiTenantAccount.getId());
+                    accountRecord.setUsername(
+                            SchemaUtil.getDisplayableId(multiTenantAccount.getClaims())
```
kreedula: same here

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1064,51 +1060,178 @@ protected void acquireTokenSilent(@NonNull final String[] scopes,
...
+                    final ITenantProfile profileForRequest =
+                            multiTenantAccount
+                                    .getTenantProfiles()
+                                    .get(tenantPath);
+
+                    validateClaimsExistForTenant(tenantPath, profileForRequest);
+
+                    accountRecord.setLocalAccountId(profileForRequest.getId());
+                    accountRecord.setUsername(
+                            SchemaUtil.getDisplayableId(profileForRequest.getClaims())
```
kreedula: same here

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1064,51 +1060,178 @@ protected void acquireTokenSilent(@NonNull final String[] scopes,
...
+                    accountRecord.setUsername(
+                            SchemaUtil.getDisplayableId(multiTenantAccount.getClaims())
+                    );
+                } else {
+                    // Use the matching profile
+                    final ITenantProfile profileForRequest =
+                            multiTenantAccount
+                                    .getTenantProfiles()
+                                    .get(tenantPath);
+
```
kreedula: Let's return an appropriate `MsalClientException` if we don't have any matching tenant profiles as discussed for silent call. For interactive, we can ignore the account.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: Resolved in f9d215149496fc299f2f606a73aa81e82f51fe73

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: Resolved in f9d215149496fc299f2f606a73aa81e82f51fe73

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: Resolved in f9d215149496fc299f2f606a73aa81e82f51fe73

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: Resolved in f9d215149496fc299f2f606a73aa81e82f51fe73

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: Resolved in 432c763f3b1a97c9cf8e8ee9f143022f58624ff8

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1064,51 +1060,181 @@ protected void acquireTokenSilent(@NonNull final String[] scopes,
...
+            // Provider Configuration Metadata in order to get the tenant id. Once we have the
+            // tenant id, we must then select the appropriate home or profile.
+            final String tenantIdNameOrAlias = aadAuthority.getAudience().getTenantId();
+
+            // The AccountRecord we'll use to request a token...
+            final AccountRecord accountRecord = new AccountRecord();
+            accountRecord.setEnvironment(multiTenantAccount.getEnvironment());
+            accountRecord.setHomeAccountId(multiTenantAccount.getHomeAccountId());
+
+            if (isHomeTenantEquivalent(tenantIdNameOrAlias)
```
rpdome: If there's a way to determine that this is not guid (UUID.fromString and catch exception?), we could probably offloading the openID metadata query here and prevent duplicating the logic below.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -979,64 +988,51 @@ protected void validateAcquireTokenSilentParameters(AcquireTokenSilentParameters
...
-        );
-
-        validateAcquireTokenParameters(acquireTokenParameters);
-
-        final AcquireTokenOperationParameters params = OperationParametersAdapter.
-                createAcquireTokenOperationParameters(
-                        acquireTokenParameters,
-                        mPublicClientConfiguration
-                );
+        sBackgroundExecutor.submit(new Runnable() {
```
rpdome: Isn't beginInteractive already dispatching to a bg thread?

iambmelt: It does inside of the ApiDispatcher, but at this entry point we're still on the main thread. To support querying OpenId metadata I'm moving this to a bg thread _earlier_

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1064,51 +1060,181 @@ protected void acquireTokenSilent(@NonNull final String[] scopes,
...
+
+                validateClaimsExistForTenant(tenantIdNameOrAlias, profileForRequest);
+
+                accountRecord.setLocalAccountId(profileForRequest.getId());
+                accountRecord.setUsername(profileForRequest.getUsername());
+
+                return accountRecord;
+            } else {
+                // We need to query metadata...
+                final OpenIdProviderConfigurationClient client =
```
rpdome: nit: suggest extracting this into a separate function.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
rpdome: ooooh.. okay

rpdome: In that case can you add a comment here?

rpdome: alternatively you could pass this whole operation as a callback to selectAccountRecordForTokenRequest. not sure which one is uglier though :D

iambmelt: Hahah ok - I think I'll leave as is. I'll take your suggestion and add a comment 👍

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: Took your suggestion, applied in a0b7e4ce4c33b99313edb78bfcd9f7351d79822a

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: Resolved in 82cfd1150f97267050368f44cc3959204b98fe10

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: Resolved in 09c2c72fabfa2da4d71e79e22d402598d4db1410

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1064,51 +1064,197 @@ protected void acquireTokenSilent(@NonNull final String[] scopes,
...
+            final String tenantIdNameOrAlias = aadAuthority.getAudience().getTenantId();
+
+            // The AccountRecord we'll use to request a token...
+            final AccountRecord accountRecord = new AccountRecord();
+            accountRecord.setEnvironment(multiTenantAccount.getEnvironment());
+            accountRecord.setHomeAccountId(multiTenantAccount.getHomeAccountId());
+
+            final boolean isUuid = isUuid(tenantIdNameOrAlias);
+
+            if (isUuid) {
```
rpdome: I was actually thinking of something like this
if (!isUuid){
final OpenIdProviderConfiguration providerConfiguration =
                         loadOpenIdProviderConfigurationMetadata(requestAuthority);

                  final String issuer = providerConfiguration.getIssuer();
                 final Uri issuerUri = Uri.parse(issuer);
                 final List<String> paths = issuerUri.getPathSegments();
                 tenantIdNameOrAlias = paths.get(0);
}

then proceed with the current if else.

iambmelt: I gotcha - this is cleaner. I'll make that change

iambmelt: Resolved in a7e162282275bf8fea08721516ccc29e9671fe1c

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1064,51 +1064,174 @@ protected void acquireTokenSilent(@NonNull final String[] scopes,
...
+            final boolean isUuid = isUuid(tenantIdNameOrAlias);
+
+            if (!isUuid && !isHomeTenantEquivalent(tenantIdNameOrAlias)) {
+                final OpenIdProviderConfiguration providerConfiguration =
+                        loadOpenIdProviderConfigurationMetadata(requestAuthority);
+
+                final String issuer = providerConfiguration.getIssuer();
+                final Uri issuerUri = Uri.parse(issuer);
+                final List<String> paths = issuerUri.getPathSegments();
+                tenantIdNameOrAlias = paths.get(0);
```
shahzaibj: Should we check for size of paths list here?

iambmelt: This result will follow a standard format of `https://login.microsoftonline.com/{tenantid}/v2.0` - if it doesn't this would be problematic. I can add a check here, though there isn't anything we can do other than throw an Exception

iambmelt: Added a simple test in a472a57e92cc5edf72b52aa672d70faddff76c9c - not exhaustive (for instance, doesn't assert that the value is actually a `UUID`) but I think it suffices

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1064,51 +1064,187 @@ protected void acquireTokenSilent(@NonNull final String[] scopes,
...
+                                    mPublicClientConfiguration.getAppContext(),
+                                    params.getAuthority(),
+                                    mPublicClientConfiguration
+                            ),
+                            callback
+                    );
+
+                    ApiDispatcher.submitSilent(silentTokenCommand);
+                } catch (final BaseException exception) {
+                    callback.onError(exception);
```
shahzaibj: @iambmelt Why are we not posting the exception to the main thread here in silent as we do above in the interactive call?

iambmelt: Because we may not be on the main thread

shahzaibj: What happens if we are on the main thread? The behavior I'm seeing in my tests due to this is that after callback onError method is called, the tests keeps running even after that.

iambmelt: The current implementation will return the error on a background thread which in your case isn't going to work so sounds like we'll have to add the same logic here as we used in [`BrokerBaseStrategy`](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerBaseStrategy.java#L74-L80). I'll open a PR for it and you can test, thanks @shahzaibj

iambmelt: @shahzaibj A proposed change is on `iambmelt/ats-error-thread`

shahzaibj: @iambmelt Thanks!

PR #724, opened 2019-08-28
Add backward compability for 'preferred_username'

rpdome:
When an app is migrated from ADAL -> MSAL, the cached id_token won't have 'preferred_username'. This breaks UX and ATS as we're using this field as username in Broker token request.


rpdome:
Thanks @iambmelt!

PR #729, opened 2019-09-03
Change UIBehavior to Prompt

shoatman:


msal/src/main/java/com/microsoft/identity/client/Prompt.java
```diff
@@ -66,7 +71,7 @@ public OpenIdConnectPromptParameter toOpenIdConnectPromptParameter() {
         switch (this) {
             case SELECT_ACCOUNT:
                 return OpenIdConnectPromptParameter.SELECT_ACCOUNT;
-            case FORCE_LOGIN:
+            case LOGIN:
                 return OpenIdConnectPromptParameter.LOGIN;
             case CONSENT:
                 return OpenIdConnectPromptParameter.CONSENT;
```
kreedula: Shouldn't default value be `OpenIdConnectPromptParameter.NONE`

msal/src/main/java/com/microsoft/identity/client/Prompt.java
```diff
@@ -41,20 +41,25 @@
...
     /**
      * acquireToken will send prompt=consent to the authorize endpoint.  The user will be prompted to consent even if consent was granted before.
      */
-    CONSENT;
+    CONSENT,
+
+    /**
+     * acquireToken will not send the prompt parameter to the authorize endpoint.  The user may be prompted to login or to consent as required by the request.
+     */
+    WHEN_REQUIRED;
```
iambmelt: `toString()` doesn't appear to handle this case

iambmelt: same for `toOpenIdConnectPromptParameter()`

msal/src/main/java/com/microsoft/identity/client/Prompt.java
shoatman: It should be WhenRequired, but i'm not changing it in this PR.

msal/src/main/java/com/microsoft/identity/client/Prompt.java
shoatman: Yes... because it's not meant to be converted to string.  I felt that throwing illegal argument exception would make that very clear.

shoatman: It has no corresponding OpenIdConnectPromptParameter value

iambmelt: This should be an `UnsupportedOperationException` then, as there is no argument

iambmelt: Also, not meant to be converted to a string sounds suspicious. What about logging?

shoatman: Let me see what I can do to achieve my goal in a different way.

shoatman: Fixed.

msal/src/main/java/com/microsoft/identity/client/Prompt.java
```diff
@@ -41,35 +44,47 @@
...
             case SELECT_ACCOUNT:
                 return OpenIdConnectPromptParameter.SELECT_ACCOUNT;
-            case FORCE_LOGIN:
+            case LOGIN:
                 return OpenIdConnectPromptParameter.LOGIN;
             case CONSENT:
                 return OpenIdConnectPromptParameter.CONSENT;
+            case WHEN_REQUIRED:
+                String error = "WHEN_REQUIRED Does not have corresponding value in in the OIDC prompt enumeration.  It's meant to convey do not sent the prompt parameter.";
+                Logger.info(tag, error);
```
kreedula: Shouldn’t we return OpenIdConnectPromptParameter.None here ? If a developer passes in WhenRequired , wouldn’t this always throw an exception when converting to OpenIdConnectParameter for brokered auth ?

shoatman: WhenRequired means don't send a prompt parameter at all.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java
```diff
@@ -180,14 +180,14 @@ public static AcquireTokenOperationParameters createAcquireTokenOperationParamet
...
-        if (acquireTokenParameters.getUiBehavior() == null) {
+        if (acquireTokenParameters.getPrompt() == null) {
             acquireTokenOperationParameters.setOpenIdConnectPromptParameter(
                     OpenIdConnectPromptParameter.SELECT_ACCOUNT
             );
         } else {
             acquireTokenOperationParameters.setOpenIdConnectPromptParameter(
                     acquireTokenParameters
-                            .getUiBehavior()
+                            .getPrompt()
```
kreedula: Ok, so when Prompt is WhenRequired , may be we should not set it to operation parameters here right? As this conversion here would throw an exception ?

shoatman: Good catch.  I'll fix.  Thanks

PR #732, opened 2019-09-04
Add Robolectric tests for PublicClientApplication methods

shahzaibj:
Related Common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/629

In this PR, I've added Robolectric tests to MSAL. I also updated submodule to point to my common branch that includes classes for ROPC Test strategies so that my tests can run on this build. The travis file was updated to run unit tests. 

The tests that hit the network (AcquireTokenNetworkTest, B2CTest) have been disabled at this time as the AutomationRunner certificate is not available on travis at this time


msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenBaseTest.java
```diff
@@ -0,0 +1,54 @@
...
+import com.microsoft.identity.client.PublicClientApplication;
+import com.microsoft.identity.client.exception.MsalException;
+
+import org.mockito.Mockito;
+
+import java.io.File;
+
+public abstract class AcquireTokenBaseTest {
+
+    private static final String CONFIG_FILE_PATH = "src/test/res/raw/aad_test_config.json";
```
rpdome: Are you planning to cover test functions specific to SingleAccountPCA or MultipleAccountPCA? I assume those function will be used more by the majority of apps. (AcquireToken with parameter is sort of 'advanced') 

Doesn't have to be in this scope - just FYI - we can chat later.

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenMockTest.java
```diff
@@ -0,0 +1,577 @@
...
+                        .fromAuthority(AAD_MOCK_AUTHORITY)
+                        .callback(new AuthenticationCallback() {
+                            @Override
+                            public void onSuccess(IAuthenticationResult authenticationResult) {
+                                fail("Unexpected Success");
+                            }
+
+                            @Override
+                            public void onError(MsalException exception) {
+                                Assert.assertTrue(true);
```
rpdome: nit: double check that we're getting an expected MsalException?

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenMockTest.java
```diff
@@ -0,0 +1,577 @@
...
+                            @Override
+                            public void onCancel() {
+                                fail("User cancelled flow");
+                            }
+                        })
+                        .build();
+
+                publicClientApplication.acquireToken(parameters);
+                RoboTestUtils.flushScheduler();
+                publicClientApplication.acquireTokenSilentAsync(silentParameters);
```
rpdome: we might have to put this inside acquireToken's callback?

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenBaseTest.java
```diff
@@ -0,0 +1,54 @@
...
+                } catch (InterruptedException e) {
+                    e.printStackTrace();
+                }
+            }
+
+            @Override
+            public void onError(MsalException exception) {
+                exception.printStackTrace();
+            }
+        });
```
rpdome: wait? (flushScheduler)

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/RoboTestUtils.java
```diff
@@ -0,0 +1,113 @@
...
+        final Map<String, ?> cacheValues = sharedPreferences.getAll();
+        final String keyToRemove = getCacheKeyForAccessToken(cacheValues);
+        if (keyToRemove != null) {
+            SharedPreferences.Editor editor = sharedPreferences.edit();
+            editor.remove(keyToRemove);
+            editor.commit();
+        }
+    }
+
+    static void flushScheduler() {
```
rpdome: Throwing out an alternative here.

Would it be cleaner to use countDownLatch instead (and rename it to waitForResult() or something)
- AcquireTokenBase calls waitForResult(), which will block the thread.
- in every exit point, we invoke returnResultPass() or returnResultFail() which essentially will unblock the thread. 

In this case we just need to call waitToResult() once per test case, at the very last line.

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenMockTest.java
```diff
@@ -0,0 +1,577 @@
...
+                        .fromAuthority(AAD_MOCK_AUTHORITY)
+                        .callback(new AuthenticationCallback() {
+                            @Override
+                            public void onSuccess(IAuthenticationResult authenticationResult) {
+                                fail("Unexpected success");
+                            }
+
+                            @Override
+                            public void onError(MsalException exception) {
+                                Assert.assertTrue(exception instanceof MsalClientException);
```
rpdome: nit: verify that we're getting the expected error message. Here and every test case we're expecting an exception.

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenMockTest.java
```diff
@@ -0,0 +1,577 @@
...
+                RoboTestUtils.flushScheduler();
+                publicClientApplication.acquireTokenSilentAsync(silentParameters);
+                RoboTestUtils.flushScheduler();
+            }
+
+        }.performTest();
+    }
+
+    @Test
+    public void canAcquireSilentIfValidTokensAvailableInCache() {
```
rpdome: Where are we injecting the token into the cache?

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/RoboTestUtils.java
shahzaibj: I can rename the method for sure, but countDownLatch does not work well with Robolectric. In my experience, I found it much easier to work with flushing the scheduler than using the countDownLatch with Robolectric.

shahzaibj: Also if we use something like returnResultPass() or returnResultFail() then we would also have multiple calls to those then for every exit point, and since there would be multiple exit points that would also add up to a lot of method calls. I'll prefer sticking to the current approach.

msal/build.gradle
```diff
@@ -74,6 +80,18 @@ android {
...
+
+    testOptions {
+        unitTests.all {
+            exclude 'com/microsoft/identity/client/robolectric/tests/AcquireTokenNetworkTest.class'
+            exclude 'com/microsoft/identity/client/robolectric/tests/B2CTest.class'
+        }
+    }
+
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
```
iambmelt: We have to be careful changing source compatibility here - this can cause issues for downstreamers of the library who are still using 1.7.

At a _minimum_, if you need this change
- Run it by immediate partners (Company Portal, Authenticator, Teams)
- Apply this to `common`, `adal`, and `android-complete` too

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenBaseTest.java
shahzaibj: Yeah, we can add those in later

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenMockTest.java
shahzaibj: That would have been the ideal case but Robolectric's threading model does not allow us to do that. By doing it this way, we are basically treating these as synchronous methods.

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenMockTest.java
shahzaibj: In this test case we are simulating that a successful acquire token call was made some time ago, and since that call was successful, we should have valid token in the cache because tokens are saved to the cache at the end of a successful acquire token call. Now we try do a silent call and it should succeed because we have valid (unexpired) tokens in the cache.

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenMockTest.java
shahzaibj: Why do you think we should check for the message? I think this can cause problems because error messages may change from time to time and that would require us to refactor the test whenever our error message changes. I think checking for type of exception should be good enough. What do you say?

msal/src/test/java/com/microsoft/identity/client/robolectric/shadows/ShadowAuthority.java
```diff
@@ -0,0 +1,89 @@
...
+import com.microsoft.identity.common.internal.logging.Logger;
+
+import org.robolectric.annotation.Implements;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.List;
+
+@Implements(Authority.class)
+public class ShadowAuthority {
```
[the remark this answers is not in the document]
shahzaibj: Addressed

msal/src/test/java/com/microsoft/identity/client/robolectric/shadows/ShadowAuthority.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private static final String TAG = ShadowAuthority.class.getSimpleName();
+
+    private static final String AAD_MOCK_PATH_SEGMENT = "mock";
+    private static final String B2C_TEST_PATH_SEGMENT = "tfp";
+
+    /**
+     * Returns an Authority based on an authority url.  This method attempts to parse the URL and based on the contents of it
```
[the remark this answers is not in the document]
shahzaibj: Addressed

msal/src/test/java/com/microsoft/identity/client/robolectric/shadows/ShadowAuthority.java
```diff
@@ -0,0 +1,89 @@
...
+
+        String authorityType = pathSegments.get(0);
+
+        switch (authorityType.toLowerCase()) {
+            // For our test environment, authority could be a AAD, B2C or a mocked authority
+            // For AAD and B2C, we create a test version of that authority that supports ROPC
+            // more cases can be added here in the future
+            case AAD_MOCK_PATH_SEGMENT:
+                //Return new AAD MOCK Authority
+                Logger.verbose(
```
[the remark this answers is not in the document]
shahzaibj: Addressed

msal/src/test/java/com/microsoft/identity/client/robolectric/shadows/ShadowStrategyResultServerError.java
```diff
@@ -0,0 +1,27 @@
...
+import com.microsoft.identity.common.internal.net.ObjectMapper;
+import com.microsoft.identity.common.internal.providers.oauth2.TokenErrorResponse;
+import com.microsoft.identity.common.internal.providers.ropc.MockTestStrategy;
+
+import org.robolectric.annotation.Implements;
+
+@Implements(MockTestStrategy.class)
+public class ShadowStrategyResultServerError {
+
+    public HttpResponse makeHttpResponseFromResponseObject(final Object obj) {
```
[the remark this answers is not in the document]
shahzaibj: Addressed

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenMockTest.java
shahzaibj: Addressed

msal/build.gradle
shahzaibj: Addressed.

That change has been removed as it was not needed after refactor.

msal/src/test/java/com/microsoft/identity/client/robolectric/shadows/ShadowAuthority.java
```diff
@@ -0,0 +1,89 @@
+package com.microsoft.identity.client.robolectric.shadows;
```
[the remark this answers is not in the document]
shahzaibj: Addressed

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenMockTest.java
rpdome: Actually, error code / error message should not be changed. (It's a contract between us and developers, and any changes made will not be obvious to them, as they should still be able to build and might not be aware of it at all).

Making a check here will help us catch unexpected changes.

shahzaibj:
`getOAuth2TokenCache()` is now package-private within `PublicClientApplicationConfiguration.java`. Tests will need a refactor to keep working

shahzaibj:
> `getOAuth2TokenCache()` is now package-private within `PublicClientApplicationConfiguration.java`. Tests will need a refactor to keep working

A solution is to move the method (that's calling the getOAuth2TokenCache) to the same package as the package where the getOAuth2TokenCache method is declared. It would still be in the test directory, and in our test directories the top level package name is always the same as the top level package name in our java source folder. So everything will just work and we can still use this without any major refactors.

msal/src/test/java/com/microsoft/identity/client/robolectric/shadows/ShadowStrategyResultServerError.java
```diff
@@ -0,0 +1,53 @@
...
+        // create a tokenErrorResponse object
+        final TokenErrorResponse errorResponse = createFakeTokenErrorResponse();
+        final String httpResponseBody = ObjectMapper.serializeObjectToJsonString(errorResponse);
+        // create http response from error response, use 500 http code
+        HttpResponse httpResponse = new HttpResponse(500, httpResponseBody, null);
+        return httpResponse;
+    }
+
+    // create a fake token error response
+    private TokenErrorResponse createFakeTokenErrorResponse() {
```
kreedula: rename to `mock`

msal/src/test/java/com/microsoft/identity/client/robolectric/shadows/ShadowStrategyResultServerError.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.microsoft.identity.common.internal.net.HttpResponse;
+import com.microsoft.identity.common.internal.net.ObjectMapper;
+import com.microsoft.identity.common.internal.providers.oauth2.TokenErrorResponse;
+import com.microsoft.identity.internal.testutils.strategies.MockTestStrategy;
+
+import org.robolectric.annotation.Implements;
+
+@Implements(MockTestStrategy.class)
+public class ShadowStrategyResultServerError {
```
kreedula: Question : Do we need to create each individual Shadow for every server error if we want to test ?

 for example errors from server like  `BAD_TOKEN`, `Protection_Policy_Required` etc ?

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenBaseTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private static final String CONFIG_FILE_PATH = "src/test/res/raw/aad_test_config.json";
+
+    abstract void makeAcquireTokenCall(final IPublicClientApplication publicClientApplication,
+                                       final Activity activity) throws InterruptedException;
+
+
+    private Activity getActivity(final Context context) {
+        final Activity mockedActivity = Mockito.mock(Activity.class);
+        Mockito.when(mockedActivity.getApplicationContext()).thenReturn(context);
```
kreedula: Curious why mocking with `Mockito` here, I thought `Roboelectric` already provides API for that ?

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenBaseTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+        return mockedActivity;
+    }
+
+
+    public void performTest() {
```
kreedula: nit : rename method to be more meaningful

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenMockTest.java
```diff
@@ -0,0 +1,600 @@
...
+
+@RunWith(RobolectricTestRunner.class)
+@Config(shadows = {ShadowStorageHelper.class, ShadowAuthority.class, ShadowHttpRequest.class, ShadowMsalUtils.class})
+public final class AcquireTokenMockTest {
+
+    private static final String[] SCOPES = {"user.read"};
+    private static final String AAD_MOCK_AUTHORITY = "https://test.authority/mock";
+
+    @Test
+    public void canAcquireToken() {
```
kreedula: nit : rename something like acquireTokenSuccessTest` and all the tests below. Also a brief description of the test would be good

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenMockTest.java
kreedula: yes actual error code from the server should not change. So validating for error code would be good and should not cause any harm, instead it will validate server behavior

msal/build.gradle
```diff
@@ -74,6 +80,13 @@ android {
             outputFileName = fileName
         }
     }
+
+    testOptions {
+        unitTests.all {
+            exclude 'com/microsoft/identity/client/robolectric/tests/AcquireTokenNetworkTest.class'
```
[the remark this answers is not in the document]
shahzaibj: What do you mean?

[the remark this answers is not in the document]
shahzaibj: The tests were written so we can start running them on Travis, why do you not want to run them?

msal/src/test/java/com/microsoft/identity/client/robolectric/shadows/ShadowStrategyResultServerError.java
shahzaibj: With the current state of the tests, yes. There is WIP on moving away from that. Of course the developer will still need create a mock success or error response but we're moving away from doing that in the shadow and will provide some other mechanism to inject different mocked responses into the test instead of using the shadow

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenBaseTest.java
shahzaibj: Provides API for what?

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenMockTest.java
shahzaibj: Why append the word test in every test? I feel like it is unnecessary as the `@test` annotation already indicates it is a test

msal/build.gradle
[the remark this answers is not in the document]
kreedula: In that case, you should not have  both `unit` and `network` tests in the same package. My intention here is to convey that we can't just keep excluding classing everytime we add one for a network test. So we need to split these in two folders and exclude let's say `roboelectric/tests/network/`

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenBaseTest.java
kreedula: API to mock an activity/ context ? Surprised  Roboelectric is not already providing a basic shadow to mock an activity

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenMockTest.java
kreedula: It's a classic convention for readability to indicate tests,  even though we religiously didn't follow in our library in few tests, most of our tests follow this convention. 

For example, if I search for `acquireToken` in the project, i can safely ignore this method since it has `test` in it. 

Also the name here `canAcquireToken` on glance doesn't indicate to me about what we are trying to test and validate. Let's say we have 10 tests for acquire token, wouldn't this be too generic to comprehend?

msal/src/test/java/com/microsoft/identity/client/robolectric/tests/AcquireTokenBaseTest.java
shahzaibj: Robolectric gives us a real context (mocked context but mocked really well). Once we have the context, we can do pretty much anything. We can also use Robolectric's `setupActivity` method to setup a specific activity. I wouldn't think of that as mocking though as the Activity returned is like a real Activity (we can click button and stuff). 

Since we do not have any Activities in our library we did not need to use that. Specifically for these tests without UI Automation, we just need a mocked Activity as we don't need to do anything with it.

msal/src/test/java/com/microsoft/identity/client/RoboTestCacheHelper.java
```diff
@@ -0,0 +1,51 @@
...
+
+    // adding this method here in its own class as the getOAuth2TokenCache method has package-private
+    // access inside the PubliClientApplication class. Therefore, it is required to place this method
+    // in this class as part of the com.microsoft.identity.client package to be able to utilize it
+    public static ICacheRecord saveTokens(TokenResponse tokenResponse, IPublicClientApplication application) throws ClientException {
+        final OAuth2TokenCache tokenCache = application.getConfiguration().getOAuth2TokenCache();
+        final String clientId = application.getConfiguration().getClientId();
+        final Authority authority = new MockAuthority();
+        final OAuth2Strategy strategy = authority.createOAuth2Strategy();
+        final MicrosoftStsAuthorizationRequest fakeAuthRequest = Mockito.mock(MicrosoftStsAuthorizationRequest.class);
```
kreedula: nit: rename to `mockAuthRequest`

gradle/versions.gradle
```diff
@@ -12,6 +12,7 @@ ext {
     androidMavenGradlePluginVersion = "1.4.1"
 
     // Libraries
+    androidxTestCoreVersion = "1.2.0"
```
iambmelt: This constants will likely need to be added `android-complete` too -- since AFAIK these values are always resolved from the root project

shahzaibj: Already done and merged:
https://github.com/AzureAD/android-complete/pull/18

msal/src/test/java/com/microsoft/identity/client/robolectric/utils/RoboTestUtils.java
```diff
@@ -0,0 +1,144 @@
...
+            editor.commit();
+        }
+    }
+
+    public static void flushScheduler() {
+        final Scheduler scheduler = RuntimeEnvironment.getMasterScheduler();
+        while (!scheduler.advanceToLastPostedRunnable()) ;
+    }
+
+    public static Activity getActivity(final Context context) {
```
kreedula: nit : rename `getMockActivity`

msal/src/test/java/com/microsoft/identity/client/RoboTestCacheHelper.java
shahzaibj: Addressed

msal/src/test/java/com/microsoft/identity/client/robolectric/utils/RoboTestUtils.java
shahzaibj: Addressed

kreedula:
LGTM, thanks @shahzaibj

PR #736, opened 2019-09-05
Add more details to AndroidManifest error. Expose loginHint to SignIn()

rpdome:
also push a commit (for bumping version for the previous release).

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
@@ -212,21 +212,21 @@ public class SingleAccountPublicClientApplication extends PublicClientApplicatio
         callback.onAccountLoaded(newAccount);
     }
 
-
     @Override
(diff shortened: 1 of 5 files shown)


msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -212,21 +212,21 @@ private void checkCurrentAccountNotifyCallback(@NonNull final CurrentAccountCall
...
 
-
     @Override
     public void signIn(@NonNull final Activity activity,
+                       @NonNull final String loginHint,
                        @NonNull final String[] scopes,
                        @NonNull final AuthenticationCallback callback) {
         acquireToken(
                 activity,
-                new String[]{"user.read"},
```
kreedula: was this an accidental change before to default this scope ?

rpdome: Probably accidental. Just caught it too.

Issue #733, opened 2019-09-04
(Pre-GA) Unimplemented App Configuration Properties


iambmelt:
AppConfig properties which are not wired-up:
- ~`HttpConfiguration`~ (Resolved in #735)
- ~`LoggerConfiguration`~ (Resolved in #738)
- ~`getSharedDeviceModeSupported()`~ (Resolved in #739)

PR #738, opened 2019-09-06
Partially implements #733 - Adds support for LoggerConfiguration

iambmelt:
See #733


iambmelt:
This PR also requires updating documentation

PR #748, opened 2019-09-10
Declined Scope Exception changes

kreedula:
- Also changed type of `forceRefresh` on `AcquireTokenSilentParams`  to `boolean` to avoid null pointers.
- corresponding common pr : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/641

PR #755, opened 2019-09-14
Change ForceRefresh to boolean to avoid getting boolean from null

rpdome:


kreedula:
This was already fixed here : https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/748

PR #756, opened 2019-09-16
Disable shared device flag in config

rpdome:
Disable this for now, as we can't promise that MSAL would work as expected for apps that sets this to false. (i.e. we prevent addAccount flow from signing in with another account, for example).

 Need to think through what the right behavior should be.


iambmelt:
@rpdome Can you make sure this brings in the latest submodule changes too?

PR #759, opened 2019-09-17
Enable broker by default

rpdome:
Also, if the redirectURI in the config is set to Broker's, we verify the AppManifest.
Otherwise, let it go and set mUseBroker to false (for backward compatibility).

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
@@ -23,6 +23,7 @@
 package com.microsoft.identity.client;
 
 import android.content.Context;
+import android.util.Log;
 
 import androidx.annotation.NonNull;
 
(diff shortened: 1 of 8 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -116,7 +122,7 @@
     TelemetryConfiguration mTelemetryConfiguration;
 
     @SerializedName(SHARED_DEVICE_MODE_SUPPORTED)
-    boolean mSharedDeviceModeSupported;
+    Boolean mSharedDeviceModeSupported;
```
kreedula: why to `object` here instead of primitive? I think we should do the other way around to use only primitives for `boolean` to avoid null pointers. 

We should change these also to `boolean` if possible in this PR, it would be great

 @SerializedName(MULTIPLE_CLOUDS_SUPPORTED)
    Boolean mMultipleCloudsSupported;

    @SerializedName(USE_BROKER)
    Boolean mUseBroker;

rpdome: Having nullable values has its pros and cons:
pros: cleaner merge config file. Given that the value is nullable, we can only merge/overwrite the default config only with what was defined by the app developer.

cons: null pointer exception. Although if this happens, it means that we forgot to set the value in the default configuration file.

PR #767, opened 2019-09-24
Fixed logic that throws/return exception when SingleAccount APIs are not used correctly.

rpdome:


msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -220,6 +220,12 @@ public void signIn(@NonNull final Activity activity,
                        @NonNull final String loginHint,
                        @NonNull final String[] scopes,
                        @NonNull final AuthenticationCallback callback) {
+        final IAccount persistedAccount = getPersistedCurrentAccount();
```
kreedula: nit : Consider using a helper function for this function and call everywhere.

rpdome: This is tricky, because what we're doing with persistedAccount object are different case-by-case.

Issue #779, opened 2019-09-30
Can't use callback methods if callback object is null


shahzaibj:
If the callback object is null, then we cannot call onSuccess or onError on that object. When we do that we run into a NullPointerException. In fact, we validate non-null for callback in the validateAcquireTokenParameters method but then we try to post that exception via the onError method of the callback which is not possible as the callback is null.

shahzaibj:
Fixed by #780

PR #780, opened 2019-09-30
Fix MSAL issue #779

shahzaibj:
In this PR I've fixed MSAL issue #779 
See #779 

To fix this issue, I check if callback is null before we call any onSuccess or onError methods on it and then we throw an IllegalStateException if the callback is null

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -184,6 +184,8 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
         private static final String ACTIVITY = "activity";
         private static final String SCOPES = "scopes";
         private static final String ACCOUNT = "account";
+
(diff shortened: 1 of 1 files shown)


shahzaibj:
@iambmelt Thanks! Constant error string has been added.

iambmelt:
@shahzaibj :shipit:

shahzaibj:
> LGTM, just to make sure, these are only methods where we return the result to user right?

@kreedula 
Inside PublicClientApplication.java yes, however, there are other methods in MultipleAccountPublicClientApplication.java where we use callback to pass the result to the user such as the `getAccount` method. The `getAccount` method or other such methods inside MultipleAccountPCA class do not have any validate Non Null logic at this time. These methods don't accept that many parameters but I think it might still be good to validate non null on these. What do you say?

kreedula:
Thanks @shahzaibj . Lets get this in and take in a fresh change for the methods mentioned

PR #743, opened 2019-09-10
Fix MSAL Issue #741

shahzaibj:
See #741

PR #782, opened 2019-10-01
Use Argument Exception when validating non-null as we use it to parse…

shahzaibj:
… error code

Logic to validate non-null was added in #743 and we were throwing the MsalArgumentException if an argument was null, however, MsalArgumentException always provides us with the `unknown_error` error code because our existing exception adapter logic cannot parse it from an Msal Exception and can only parse it from a corresponding internal Exception that is located in common. For a complete understanding, please see [MsalExceptionAdapter.java](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/internal/controllers/MsalExceptionAdapter.java)

I've modified the validateNonNullArg method to instead throw an ArgumentException which can be parsed by the ExceptionAdapter.


shahzaibj:
@kreedula I will close this PR as I've added a better fix here in #784 

Any thoughts?

shahzaibj:
Closing this PR as moving forward with fix introduced in #784

PR #791, opened 2019-10-04
Add Server Side Telemetry to MSAL

shahzaibj:
Related Common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/671


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1390,6 +1403,7 @@ public void run() {
                             callback
                     );
 
+                    silentTokenCommand.setPublicApiId(PublicApiId.LOCAL_ACQUIRE_TOKEN_SILENT);
```
rpdome: As discussed, 
if you want to track what API is calling, you could rename this function to internal and take an API constant as another parameter, then pass that parameter to setPublicApiId. Should be a pretty straighforward and simple change.

You could do it in this PR or a separate one. That one i'll let you make a call.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1312,6 +1323,7 @@ public void run() {
                             localAuthenticationCallback
                     );
 
+                    command.setPublicApiId(PublicApiId.LOCAL_ACQUIRE_TOKEN_INTERACTIVE);
```
rpdome: Same as the comment below.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1064,6 +1067,14 @@ private void initializeTokenSharingLibrary() {
         }
     }
 
+    private void setupServerSideTelemetry(Context context) {
+        EstsTelemetry.initializeEstsTelemetryCache(context);
+        com.microsoft.identity.common.internal.logging.Logger.verbose(
+                TAG,
+                "Server side telemetry cache has been initialized properly."
```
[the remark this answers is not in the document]
shahzaibj: We aren't really doing much here other than calling constructors and creating objects. The context parameters is non-null, however someone can still pass null to that and in that case the cache won't be initialized. I can add an explicit null check inside the method and also move the log into the method and either log a successful message is cache is created or failure if context is null.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shahzaibj: Yeah, sounds good. I think I'll do that in a separate PR after this gets merged in.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shahzaibj: Replied to the comment above.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -334,6 +333,7 @@ public void onTaskCompleted(Boolean result) {
                     }
             );
 
+            command.setPublicApiId(PublicApiId.SIGN_OUT);
```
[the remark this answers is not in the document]
shahzaibj: Addressed

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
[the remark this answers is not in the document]
shahzaibj: I added the log as stated above.

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -517,10 +518,10 @@ public void acquireTokenSilentAsync(@NonNull final String[] scopes,
     @WorkerThread
     public IAuthenticationResult acquireTokenSilent(@NonNull final String[] scopes,
                                                     @NonNull final String authority) throws MsalException, InterruptedException {
-
         final IAccount persistedAccount = getPersistedCurrentAccount();
         if (persistedAccount == null) {
-            throw new MsalClientException(MsalClientException.NO_CURRENT_ACCOUNT);
+            MsalClientException exception = new MsalClientException(MsalClientException.NO_CURRENT_ACCOUNT);
```
kreedula: why a new variable ?

shahzaibj: I don't remember doing it. I might have accidentally clicked on the `introduce local variable` suggestion within Android Studio and that might have done it. I'll remove it.

shahzaibj: @kreedula Addressed

PR #795, opened 2019-10-05
Deleted Shadows from MSAL which are moved to common

kreedula:


kreedula:
Closing this PR as attempting to `common` and reusing is not working roboelectric

PR #800, opened 2019-10-10
Refactor E2E network tests to run individually, & also as a Suite

shahzaibj:
In this PR, I've refactored robolectric network tests to allow running as a suite as well as individually.

The acquire token network tests were originally written using the `ParameterizedRobolectricTestRunner` because it allowed us to get rid of duplicated tests as the tests for both AAD and B2C are the same and the only difference lied in the json configuration file (or the authority in that file) passed when creating the PublicClientApplication.

However, parameterized tests cannot be run individually, and this is a problem because it stops us from being able to debug a specific test. We would would have to run all tests in the class even if we wanted to debug / run a specific test.

To solve this problem, I've replaced the parameterized runner with regular `RobolectricTestRunner`, and I've introduced a couple more test classes using java inheritance that still allows us to avoid duplicating tests while still being able to run them individually as well as a suite.

### DIFF
--- a/msal/src/test/java/com/microsoft/identity/client/robolectric/tests/network/AcquireTokenNetworkAADTest.java
+++ b/msal/src/test/java/com/microsoft/identity/client/robolectric/tests/network/AcquireTokenNetworkAADTest.java
@@ -1,0 +1,9 @@
+package com.microsoft.identity.client.robolectric.tests.network;
+
+public class AcquireTokenNetworkAADTest extends AcquireTokenNetworkTest {
+
+    public AcquireTokenNetworkAADTest() {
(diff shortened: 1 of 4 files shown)


msal/src/test/java/com/microsoft/identity/client/robolectric/tests/network/AcquireTokenNetworkAADTest.java
```diff
@@ -0,0 +1,12 @@
+package com.microsoft.identity.client.robolectric.tests.network;
```
iambmelt: License info, here and elsewhere

shahzaibj: Addressed

PR #803, opened 2019-10-11
Refactor public client api overloads for ests telemetry

shahzaibj:
Related Common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/687

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
@@ -74,7 +74,7 @@ public class MultipleAccountPublicClientApplication extends PublicClientApplicat
 
     @Override
     public IAuthenticationResult acquireTokenSilent(@NonNull String[] scopes, @NonNull IAccount account, @NonNull String authority) throws MsalException, InterruptedException {
-        return acquireTokenSilentSyncInternal(scopes, authority, account, false);
+        return acquireTokenSilentSyncInternal(scopes, authority, account, false, PublicApiId.MULTIPLE_ACCOUNT_PCA_ACQUIRE_TOKEN_SILENT_WITH_SCOPES_ACCOUNT_AUTHORITY);
     }
 
     @Override
@@ -91,7 +91,7 @@ public class MultipleAccountPublicClientApplication extends PublicClientApplicat
                 callback
         );
 
-        acquireTokenSilentAsyncInternal(acquireTokenSilentParameters);
+        acquireTokenSilentAsyncInternal(acquireTokenSilentParameters, PublicApiId.MULTIPLE_ACCOUNT_PCA_ACQUIRE_TOKEN_SILENT_ASYNC_WITH_SCOPES_ACCOUNT_AUTHORITY_CALLBACK);
     }
 
     /**
@@ -101,7 +101,7 @@ public class MultipleAccountPublicClientApplication extends PublicClientApplicat
      */
     @Override
     public void getAccounts(@NonNull final LoadAccountsCallback callback) {
-        getAccountsInternal(callback);
+        getAccountsInternal(callback, PublicApiId.MULTIPLE_ACCOUNT_PCA_GET_ACCOUNTS_WITH_CALLBACK);
     }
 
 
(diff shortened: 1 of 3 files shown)


shahzaibj:
> Suggest running the code formatter, looks like there might be some extra spaces here and there inside of `MultipleAccountPublicClientApplication`, other files too maybe.
> 
> Approving this PR with comments -- these are learnings from ADAL:
> 
> * I see us already taking steps towards plumbing API event ids into the params of methods -- in this PR it all looks unavoidable however as a general statement I will say that I'd like for us to minimize the amount of _required_ parameters to methods that aren't directly related to their _core function_. Without getting too prescriptive, any bit of state that snakes through the library's flow of control is essentially a future barrier to refactoring. It becomes a challenge to refactor the guts of the library without affecting the order or contents of the telemetry emitted.

Agreed @iambmelt , I will see how I can refactor it in a future PR. My initial though is that we need something similar to a DiagnosticContext that needs to be initialized earlier (prior to entering the ApiDispatcher) and that context will need to be aware of the actual overload that is being used.

Thanks for the feedback!

msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
```diff
@@ -207,7 +217,7 @@ public void onTaskCompleted(List<IAccount> result) {
             public void onError(MsalException exception) {
                 future.setResult(new AsyncResult<List<IAccount>>(null, exception));
             }
-        });
+        }, PublicApiId.MULTIPLE_ACCOUNT_PCA_GET_ACCOUNTS);
```
rpdome: should we differentiate between async and sync overloads?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1211,18 +1211,20 @@ public void acquireToken(@NonNull final Activity activity,
...
-                                @NonNull final String[] scopes,
-                                @Nullable final IAccount account,
-                                @Nullable final Prompt uiBehavior,
-                                @Nullable final List<Pair<String, String>> extraQueryParameters,
-                                @Nullable final String[] extraScopesToConsent,
-                                @Nullable final String authority,
-                                @NonNull final AuthenticationCallback callback,
-                                @Nullable final String loginHint,
-                                @Nullable final ClaimsRequest claimsRequest) {
+    AcquireTokenParameters buildAcquireTokenParameters(@NonNull final Activity activity,
```
rpdome: nit: format.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1557,6 +1568,14 @@ private static boolean isUuid(@NonNull final String tenantIdNameOrAlias) {
     public IAuthenticationResult acquireTokenSilent(
             @NonNull final AcquireTokenSilentParameters acquireTokenSilentParameters)
             throws InterruptedException, MsalException {
+        return acquireTokenSilentInternal(acquireTokenSilentParameters, PublicApiId.PCA_ACQUIRE_TOKEN_SILENT_WITH_PARAMETERS);
```
rpdome: We might want to know if this is invoked by MultipleAccountPCA or SingleAccountPCA.

i.e. It might be hard to tell which acquireToken/acquireTokenSilent overload in MutlipleAccountPCA is 'popular', since we can't tell how many % of the number we get from this PCA_ACQUIRE_TOKEN_SILENT_WITH_PARAMETERS is from MultipleAccountPCA.

rpdome: This means we would need to override this in the subclasses, and mark this as 'not implemented'.
(Might require some change in robolectric, though.)

shahzaibj: @rpdome Based on what we discussed, there are about four different methods for which we will have to do this. So that would result in creating 8 new methods (4 in each Single Account PCA & Multiple Account PCA) that all do the same thing. I prefer not to take this change at the moment. We can do this later once we find a better way to achieve this.

PR #805, opened 2019-10-14
Throw IllegalStateException if callback is null for single accout pca

shahzaibj:
This was already done for the base class `PublicClientApplication`, but we also need to do it for `SingleAccountPublicClientApplication` as it overrides the `getCommandCallback` method.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
@@ -251,6 +251,10 @@ public class SingleAccountPublicClientApplication extends PublicClientApplicatio
 
             @Override
             public void onTaskCompleted(@NonNull final ILocalAuthenticationResult localAuthenticationResult) {
+                if (authenticationCallback == null) {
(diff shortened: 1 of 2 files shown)


msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -251,6 +251,10 @@ public void signIn(@NonNull final Activity activity,
 
             @Override
             public void onTaskCompleted(@NonNull final ILocalAuthenticationResult localAuthenticationResult) {
+                if (authenticationCallback == null) {
```
kreedula: we already check for callback `non null` before invoking the request itself right ? How are we running into this

shahzaibj: we check for non-null and if it is NULL, then we post the error using the `onError` of the callback itself, but since the callback itself is null, hence we can't actually post that error and run into a `NullPointerException`. I've added this check to instead throw an `IllegalStateException` which is what happens in the case of Multiple Account as well.

PR #809, opened 2019-10-22
Use Bound service first

rpdome:


shahzaibj:
Why is this needed?

shahzaibj:
Test is failing

com.microsoft.identity.client.AuthenticationActivityTest > testIntentWithNoRequestUri[test(AVD) - 5.0.2] FAILED 
	junit.framework.AssertionFailedError
	at junit.framework.Assert.fail(Assert.java:48)

I will restart the build and then lets see

PR #812, opened 2019-10-24
Docs first commit

shoatman:
Home.md and placeholder pages for additional documentation.


docs/home.md
```diff
@@ -0,0 +1,42 @@
...
+## Android Libraries
+
+MSAL consists of 2 Android Libraries/Packages:
+
+- com.microsoft.identity.client.msal
+  - github: [https://github.com/AzureAD/microsoft-authentication-library-for-android](https://github.com/AzureAD/microsoft-authentication-library-for-android)
+- com.microsoft.identity.common
+  - github: [https://github.com/AzureAD/microsoft-authentication-library-common-for-android](https://github.com/AzureAD/microsoft-authentication-library-common-for-android)
+
+Where MSAL depends on common for definition of abstractions and implementation identity provider specific implementation.
```
shahzaibj: This sentence seems to have a typo or mistake. May be re-write to:

`Where MSAL depends on common for definition of abstractions and identity provider specific implementation.`

docs/home.md
```diff
@@ -0,0 +1,42 @@
...
+- com.microsoft.identity.client.msal
+  - github: [https://github.com/AzureAD/microsoft-authentication-library-for-android](https://github.com/AzureAD/microsoft-authentication-library-for-android)
+- com.microsoft.identity.common
+  - github: [https://github.com/AzureAD/microsoft-authentication-library-common-for-android](https://github.com/AzureAD/microsoft-authentication-library-common-for-android)
+
+Where MSAL depends on common for definition of abstractions and implementation identity provider specific implementation.  
+
+## Component Overview Diagram
+
+The following component diagram illustrates the components of the library and the dependencies between them.  Abstract components are marked as with the "abstract" annotation.  Each components and groups of components will be covered in more details in separate documents linked to below.
```
shahzaibj: Change `Each components and groups of components will be covered in more details in separate documents linked to below.` to as follows:

`Each component and group of components will be covered in more detail in separate documents linked below.`

shahzaibj: What do you mean by `abstract` components? The are implemented with `abstract` classes?

shahzaibj:
In the Diagram, I see a component called as `OAuth2StrategyConfiguration`, may be we rename to `OAuth2Configuration` as that will match the actual class name in our code

docs/home.md
```diff
@@ -0,0 +1,43 @@
...
+
+![MSAL Component Diagram](./images/MSAL_Components.png)
```
rpdome: Single account mode also caches current account in shared pref. Not sure if we have/should address this in the diagram.

rpdome: BrokerClientStrategy should be BrokerBaseStrategy instead.

rpdome: I think we might want to add the inherited classes in the file instead of abstract classes. I personally find it sort of hard to read.

i.e. In the case of Controller..
1. "BrokerBaseStrategy" is only invoked by a variant of controller (BrokerMsalController), not all of them.
2. There are also controllers in common that only exists in Broker, but not MSAL.

docs/publicclientapplication.md
```diff
@@ -0,0 +1,45 @@
+# Public Client Applications and Factory Methods
+
+## Introduction
+
+The primary interface for using the Microsoft Authentication Library (MSAL) is the PublicClientApplication.  The name comes directly from the OAuth2 specification and refers to clients that cannot securely keep secrets.  All mobile applications are considered public clients from an oAuth perspective.  
+
+MSAL Android divides public client application into 2 sub-types:
+
+- Single Account Public Client Application - Which is intended to be used by applications that only expect one account to be in use at a time
+- Multiple Account Public Client Applications - Which is intended to be used by applications that expect there user to have multiple accounts and to want to use multiple accounts at the same time.
```
rpdome: nit: typo (there user)

docs/home.md
shoatman: Fixed.

docs/home.md
shoatman: Could be an abstract class could be an interface.  I would argue that both of these are abstractions.

docs/home.md
shoatman: Fixed BrokerBaseStrategy

shoatman: This diagram is probably more art than executable.  I want to keep the MSAL docs free of broker implementation details.... we should cover that in the broker docs.

docs/publicclientapplication.md
shoatman: fixed.

docs/home.md
shoatman: Fixed.  thanks

docs/home.md
```diff
@@ -0,0 +1,43 @@
...
+- com.microsoft.identity.common
+  - github: [https://github.com/AzureAD/microsoft-authentication-library-common-for-android](https://github.com/AzureAD/microsoft-authentication-library-common-for-android)
+
+Where MSAL depends on common for definition of abstractions and of the identity provider specific implementation.
+
+## Component Overview Diagram
+
+The following component diagram illustrates the components of the library and the dependencies between them.  Abstract components are marked as with the "abstract" annotation.  Each component and group of components will be covered in more detail in separate documents linked below.
+
+![MSAL Component Diagram](https://www.lucidchart.com/publicSegments/view/88a1bca3-18ab-4eac-aa4f-ba85d4d214f6/image.png)
```
kreedula: Could be for future, but what do you think of adding components to indicate that the controller calls `MicrosoftTokenRequest` or `TokenRequest` after it gets the `auth code ` from `Authorization Strategy` ?

shoatman: I was thinking that we'll likely need a separate document just on the controllers....Effectively documenting the business logic of the controllers.

PR #813, opened 2019-10-25
Enable MSAL to talk to Broker via AccountManager without requiring GET_ACCOUNTS permission

rpdome:


msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerBaseStrategy.java
```diff
@@ -79,53 +94,16 @@ Handler getPreferredHandler() {
...
-        );
-
-        requestBundle.putInt(
-                AuthenticationConstants.Broker.CALLER_INFO_UID,
-                Binder.getCallingUid()
+                new Gson().toJson(
+                        mRequestAdapter.brokerRequestFromAcquireTokenParameters(parameters),
+                        BrokerRequest.class)
         );
+        interactiveRequestIntent.putExtra(AuthenticationConstants.Broker.ACCOUNT_NAME, parameters.getLoginHint());
```
kreedula: is this new ? why are we passing this seperately, wouldn't this be part of the `BrokerRequest`?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAccountManagerStrategy.java
```diff
@@ -56,397 +48,306 @@
...
-                                .putErrorDescription(errorMsg)
-                );
-
-                throw new ClientException(
-                        ErrorStrings.BROKER_REQUEST_CANCELLED,
-                        errorMsg,
-                        e
-                );
-            }
+        if (!BrokerMsalController.isAccountManagerPermissionsGranted(parameters.getAppContext())) {
```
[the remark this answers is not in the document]
rpdome: We still do (for pre API 22)

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerBaseStrategy.java
rpdome: Seems like the only place that uses this (apart from logging) is BrokerAccountService - meaning this is not needed.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAccountManagerStrategy.java
```diff
@@ -56,397 +48,305 @@
...
-                throw new ClientException(
-                        ErrorStrings.BROKER_REQUEST_CANCELLED,
-                        errorMsg,
-                        e
-                );
-            }
+        return result;
+    }
+
+    private static BaseException getBaseExceptionFromException(final Exception e) {
```
shahzaibj: We already have a method like this in `ExceptionAdapter.java` within common, why not use that?

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -384,7 +385,7 @@ private void persistCurrentAccount(@Nullable final List<ICacheRecord> cacheRecor
             return;
         }
 
-        String currentAccountJsonString = MsalBrokerResultAdapter.getJsonStringFromICacheRecordList(cacheRecords);
+        String currentAccountJsonString = JsonExtensions.getJsonStringFromICacheRecordList(cacheRecords);
```
shahzaibj: nit: final

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAccountManagerStrategy.java
```diff
@@ -56,397 +48,305 @@
...
             }
 
-            Telemetry.emit(
-                    new BrokerEndEvent()
-                            .putAction(methodName)
-                            .isSuccessful(true)
-            );
+            @Override
+            public String getMethodName() {
+                return ":helloWithAccountManager";
```
shahzaibj: This does not match the actual method name, is that intended ?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAccountManagerStrategy.java
```diff
@@ -56,397 +48,305 @@
...
+            }
 
-            return cacheRecords;
-        }
+            @Override
+            public Boolean getResultFromBundle(Bundle bundle) throws BaseException {
```
shahzaibj: nit: final

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAccountManagerStrategy.java
```diff
@@ -56,397 +48,305 @@
...
+                    public Bundle getRequestBundle(AcquireTokenOperationParameters parameters) {
+                        final Bundle requestBundle = new Bundle();
+                        requestBundle.putString(AuthenticationConstants.Broker.BROKER_ACCOUNT_MANAGER_OPERATION_KEY,
+                                AuthenticationConstants.BrokerAccountManagerOperation.GET_INTENT_FOR_INTERACTIVE_REQUEST);
+                        return requestBundle;
+                    }
+
+                    @Override
+                    public String getMethodName() {
+                        return ":acquireTokenSilentWithAccountManager";
```
shahzaibj: This is wrong

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAccountManagerStrategy.java
rpdome: Yes.

shahzaibj:
@rpdome Left few comments

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAccountManagerStrategy.java
rpdome: It's not exactly same. (i.e. this returns an error specific to this class if the exception can't be matched).

shahzaibj: May be we can add that as a case to the method in common? 

Leaving it here is fine too. I'll let you pick.

rpdome: We can't. This generic error message is specific to account manager.

Issue #806, opened 2019-10-15
RedirectURI should be consistent and consumed ideally from one location


titorodi:
**Is your feature request related to a problem? Please describe.**
It's inconvenient to get the brokerRedirectUri, which from the getBrokerRedirectAPI is not URL encoded, cannot be URL encoded in the Android manifest, but must be URL encoded in the JSON config, and the make sure I copy it to the AndroidManifest.xml and tweak it as appropriate as I debug with one signature or make an official build with another signature.

Also, the error logged is incorrect JSON (see quoting ' vs " on path:)

2019-10-15 15:16:41.633 30515-30515/com.microsoft.identity.client.sample.dist E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.microsoft.identity.client.sample.dist, PID: 30515
    java.lang.IllegalStateException: Intent filter for: BrowserTabActivity is missing.  Please make sure you have the following activity in your AndroidManifest.xml 
    
    <activity android:name="com.microsoft.identity.client.BrowserTabActivity">
    	<intent-filter>
    		<action android:name="android.intent.action.VIEW" />
    		<category android:name="android.intent.category.DEFAULT" />
    		<category android:name="android.intent.category.BROWSABLE" />
    		<data
    			android:host="com.microsoft.identity.client.sample.dist"
    			android:path='/4Zbe4iISKyw2VEycARmSsWI9gfc="
    			android:scheme="msauth" />
    	</intent-filter>
    </activity>
    
        at com.microsoft.identity.client.PublicClientApplicationConfiguration.checkIntentFilterAddedToAppManifestForBrokerFlow(PublicClientApplicationConfiguration.java:466)
        at com.microsoft.identity.client.PublicClientApplication.initializeApplication(PublicClientApplication.java:994)
        at com.microsoft.identity.client.PublicClientApplication.<init>(PublicClientApplication.java:977)
        at com.microsoft.identity.client.MultipleAccountPublicClientApplication.<init>(MultipleAccountPublicClientApplication.java:71)
        at com.microsoft.identity.client.PublicClientApplication$2.onGetMode(PublicClientApplication.java:863)
        at com.microsoft.identity.client.internal.controllers.BrokerMsalController$1$1.run(BrokerMsalController.java:390)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at android.os.Looper.loop(Looper.java:214)
[...]

PR #824, opened 2019-10-31
Fix RedirectURI path error message

rpdome:
#806


shahzaibj:
So this is a partial fix for issue #806 ?

rpdome:
@shahzaibj yes

PR #830, opened 2019-11-11
Snapshot update of common dependancy + submodule

kreedula:


msal/build.gradle
```diff
@@ -140,7 +140,7 @@ dependencies {
         transitive = false
     }
 
-    snapshotApi(group: 'com.microsoft.identity', name: 'common', version: '1.0.8-SNAPSHOT', changing: true)
+    snapshotApi(group: 'com.microsoft.identity', name: 'common', version: '1.0.9-SNAPSHOT', changing: true)
```
rpdome: Let's skip 1.0.9 and 1.0.10 (those are for the unreleased broker).

kreedula: but common dev is at 1.0.9 right ? we wouldn't get changes from dev of common otherwise

rpdome: ahh.. sorry.. 1.0.9 belongs to 3.1.5 (released with the memleak fix). So yes. you were right. Please ignore my previous comment.

kreedula:
Few unstable tests are failing, these are known and not related to PR, @shahzaibj is planning to make changes to ignore these tests in future builds

PR #837, opened 2019-11-14
AT/PoP Implementation

iambmelt:
Work in progress.

Expands:
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/733/


msal/src/main/java/com/microsoft/identity/client/HttpMethod.java
```diff
@@ -0,0 +1,75 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.client;
+
+/**
+ * HTTP Request Methods as defined in <a href="https://tools.ietf.org/html/rfc7231#section-4.3">RFC-7231/§4.3</a>
+ * and <a href="https://tools.ietf.org/html/rfc5789">RFC-5789</a>.
+ */
+public enum HttpMethod {
```
iambmelt: Remove this, use a String instead.

iambmelt: Actually, keep it, but don't pass it to `common`

iambmelt:
TODO: Fix serialization of AuthenticationScheme...

See `BrokerAuthServiceStrategy#acquireTokenSilent()` -> `perform()`

Faulty impl in: `MsalBrokerRequestAdapter#getRequestBundleForAcquireTokenSilent()`

`mParams` are not serializing, suspect customer serializer is needed in `Gson`? See instance `sRequestAdapterGsonInstance`

msal/src/androidTest/java/com/microsoft/identity/client/AuthorityTest.java
```diff
@@ -119,7 +121,7 @@ public URL getAuthorityURL() {
         }
 
         @Override
-        public OAuth2Strategy createOAuth2Strategy() {
+        public OAuth2Strategy createOAuth2Strategy(OAuth2StrategyOptions options) {
```
iambmelt: `final`

iambmelt: Resolved in 1ef4aff7a55b9af604534bb8dae949cf47b1955a

iambmelt:
> TODO: Fix serialization of AuthenticationScheme...
> 
> See `BrokerAuthServiceStrategy#acquireTokenSilent()` -> `perform()`
> 
> Faulty impl in: `MsalBrokerRequestAdapter#getRequestBundleForAcquireTokenSilent()`
> 
> `mParams` are not serializing, suspect customer serializer is needed in `Gson`? See instance `sRequestAdapterGsonInstance`

#Resolved

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalWrapper.java
```diff
@@ -148,14 +152,33 @@ public void acquireToken(final Activity activity,
...
+                                        .toLowerCase()
+                                        .split(" ")
+                        )
+                )
+                .withCallback(getAuthenticationCallback(notifyCallback));
+
+        if (USE_POP) {
+            try {
+                builder.withProofOfPossession(
+                        new URL("https://signedhttprequest.azurewebsites.net/api/validateSHR"),
```
iambmelt: Should either remove this or add new controls to testapp to support PoP testing....

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
```diff
@@ -32,29 +32,52 @@
...
         mAccessToken = mostRecentlyAuthorized.getAccessToken();
         mTenantId = mostRecentlyAuthorized.getAccount().getRealm();
         mAccount = AccountAdapter.adapt(cacheRecords).get(0);
+        mAuthorizationHeader = authHeader;
     }
 
     @Override
     @NonNull
     public String getAccessToken() {
+        if (mAuthorizationHeader.startsWith(SCHEME_POP + SCHEME_DELIMITER)) {
```
rpdome: why do we have to do this for PoP? 

I'd expect mAccessToken.getSecret() to return the same value. Isn't that the case?

rpdome: I'm thinking if we should mint the AT in AuthenticationResultAdapter.adapt(), or even here. (instead of minting it as part of mAuthorizationHeader and pass it all the way up from the lower layer).

i.e. if we want to do here, we could probably do something like
if (mAccessToken.getAccessTokenType() == SCHEME_POP){
    return (IPopAuthenticationSchemeInternal) scheme.mintAccessToken(mAccessToken.getSecret());
}

return mAccessToken.getSecret();

If we do this in adapt(), then there is no need to pass the ITokenAuthenticationSchemeInternal object here.
also this means there is no need to add mAccessToken as part of the scheme object.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -188,6 +188,7 @@
         static final String ACTIVITY = "activity";
         static final String SCOPES = "scopes";
         static final String ACCOUNT = "account";
+        static final String AUTH_SCHEME = "authentication_scheme";
```
rpdome: is this used?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
```diff
@@ -111,8 +114,15 @@ public AcquireTokenResult acquireToken(@NonNull final AcquireTokenOperationParam
...
 
+        // Build up params for Strategy construction
+        final OAuth2StrategyParameters strategyOptions = new OAuth2StrategyParameters();
+        strategyOptions.setContext(parameters.getAppContext());
+
         //1) Get oAuth2Strategy for Authority Type
-        final OAuth2Strategy oAuth2Strategy = parameters.getAuthority().createOAuth2Strategy();
+        final OAuth2Strategy oAuth2Strategy = parameters
+                .getAuthority()
+                .createOAuth2Strategy(strategyOptions);
```
rpdome: I'm more inclined to just pass parameter.getAppContext() here. (I assume that you introduce this for future-proofing) 

Your call.

iambmelt: I went back-and-forth on this quite a bit....and like you I'm not totally satisfied with it. I ultimately went with this approach for two reasons:
1. Future proofiness (as you call out)
2. Not all the places we construct a `Strategy` do we have a `Context` available -- we could annotate this as a `@Nullable` parameter but I thought this might actually be weirder and less flexible

Does that strike you as a reasonable enough reason to leave-in this idea of `StrategyOptions`?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: Removed in d6c1d6ca7a10b33fe77dd2ad871842b49d34ee1c

msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
```diff
@@ -316,7 +337,12 @@ public AcquireTokenResult acquireTokenSilent(
                     new LocalAuthenticationResult(
                             fullCacheRecord,
                             cacheRecords,
-                            SdkType.MSAL
+                            SdkType.MSAL,
+                            strategy.getAuthorizationHeader(
+                                    parameters.getAuthenticationScheme(),
+                                    new TokenResponse() {{
+                                        setAccessToken(fullCacheRecord.getAccessToken().getSecret());
+                                    }}) // TODO this is a workaround
```
kreedula: TODO here, i see a fake token response is created, may be add an api with takes in just the AT and  Auth scheme and return the `authorization header` ?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
rpdome: I'm fine with that.

msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
```diff
@@ -40,14 +41,33 @@
...
-    protected TokenParameters(final TokenParameters.Builder builder) {
+    protected TokenParameters(@NonNull final TokenParameters.Builder builder) {
         mAccount = builder.mAccount;
         mAuthority = builder.mAuthority;
         mClaimsRequest = builder.mClaimsRequest;
         mScopes = builder.mScopes;
+
+        // In the future, if more scheme are added, some conditional logic will need to be
+        // introduced here.
+        if (null != builder.mAuthenticationScheme) {
```
shahzaibj: I think this logic can move to the `OperationParametersAdapter` class. I think TokenParameters should just contain the value that was set by the user (which could be `null` if the user didn't provide anything), and then when creating `OperationParameters` we can set the appropriate value (or the fallback value) based on this logic.

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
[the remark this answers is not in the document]
iambmelt: Hah! So this was originally implemented this way, but unfortunately doesn't work for broker scenarios.... (which I realized late and had to refactor 😅)

The `DevicePoPManager` singleton uses different keys depending on whether or not we're in the broker or a client app using MSAL, because of this we can't late-bind the signing because MSAL apps won't have access to the same signing keys used by the broker.

AKA

We can't sign the result once we get it back from broker, it instead needs to be signed _inside_ the broker and passed up/back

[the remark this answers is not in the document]
iambmelt: Grab me or post more feedback if you have more to add here - in the meantime, I'll mark resolved and keep chugging

msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
iambmelt: Resolved in 36b08f67f5b0eafe99733a7cd3031728888e0640

msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
iambmelt: I'll take this feedback, I am going to update the `AuthenticationSchemeFactory` in `common` (which is called by `OperationParametersAdapter`)

iambmelt: Resolved in 757354f0202ec40b61377c76386ad0948fd0ded1

iambmelt:
@rpdome @shahzaibj @kreedula Feedback resolved/responded -- ready for another round. Thanks!

PR #839, opened 2019-11-19
Broker BrowserSupport for Cobo

kreedula:
- Introduced new class `IntuneAcquireTokenParameters` in the internal package (only intended to be used by COBO app) to set Broker Browser support


msal/src/main/java/com/microsoft/identity/client/internal/IntuneAcquireTokenParameters.java
```diff
@@ -0,0 +1,55 @@
...
+import com.microsoft.identity.client.AcquireTokenParameters;
+
+/**
+ * Internal class specifically for the Microsoft Intune app (COBO) to app to enable System Browser support on Broker.
+ */
+public class IntuneAcquireTokenParameters extends AcquireTokenParameters {
+
+    private boolean mBrokerBrowserSupportEnabled;
+
+    public IntuneAcquireTokenParameters(final Builder builder) {
```
iambmelt: Given this class has a builder, should the ctor be `public`? Slightly confused by this API - I would expect the `Builder` in this class to call a `private` ctor

msal/src/main/java/com/microsoft/identity/client/internal/IntuneAcquireTokenParameters.java
```diff
@@ -0,0 +1,55 @@
...
+    public static class Builder extends AcquireTokenParameters.Builder {
+
+        private boolean mBrokerBrowserSupportEnabled;
+
+        public IntuneAcquireTokenParameters.Builder brokerBrowserSupportEnabled(boolean brokerBrowserSupportEnabled){
+            mBrokerBrowserSupportEnabled = brokerBrowserSupportEnabled;
+            return this;
+        }
+
+
```
iambmelt: No `@Override` for `build()`?

PR #840, opened 2019-11-20
Add prod config for testing prod MSAL. Add debug keystore.

rpdome:


testapps/testapp/build.gradle
```diff
@@ -55,12 +63,19 @@ android {
...
+            resValue("string", "application_name", "msal-prod")
+        }
     }
 }
 
 dependencies {
     // Compile Dependency
-    implementation project(':msal')
+    localImplementation project(':msal')
+    distImplementation project(':msal')
```
iambmelt: How does dist differ from prod? Could be an internal (or different) artifact stream?

rpdome: There might be better names for these, but basically there should be 3 flavors.

1. local: local MSAL using local common. (to debug your branch with debug Broker).
2. dist: local MSAL using published common. (to debug your branch with Prod Broker).
3. prod: published MSAL. (specifically for bug bash - we're pulling this from internal stream).

I added #3 to simplify the bugbash process - as of now I have to change gradle script manually to make sure it's pulling the bug-bashed MSAL.

rpdome: Here are what they do.

1. local: local MSAL using local common.
2. dist: local MSAL using published common.
3. prod: published MSAL.

after giving it some more thought, we might not need the current 'dist', and I could make 'prod' to be the new 'dist'.

rpdome: gently ping @iambmelt

iambmelt: @rpdome I might go the other way; it seems that `dist` can be easily pointed towards production by updating a constant. As developers would have to verify that the `prod` pointer goes to the latest build anyway, it might be my preference to just have `dist` which, when it's checked it, points to prod.

All that said, I am not terribly fixated on the naming, `dist` seems fine to me unless you have a compelling reason to want to call it `prod`

rpdome: I'll stick with dist then. Thanks!

PR #842, opened 2019-11-22
Adding a null check on account claims to ensure we don't assign account which is not signed in

kreedula:


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1508,14 +1508,18 @@ private AccountRecord selectAccountRecordForTokenRequest(
...
+                                authority.getAuthorityURL().toString()
+                        );
             }
 
             IAccount accountForRequest;
 
             if (isHomeTenantAlias(tenantId)
                     || isAccountHomeTenant(multiTenantAccount.getClaims(), tenantId)) {
-                accountForRequest = multiTenantAccount;
+                accountForRequest = (multiTenantAccount.getClaims() != null) ? multiTenantAccount : null;
```
rpdome: if claims is null, then isAccountHomeTenant will return false.

kreedula: yes but it's an `or` check , so if  `isHomeTenantAlias` is true we end up setting an account with no claims

PR #843, opened 2019-11-25
Add authority to account

shahzaibj:
Allows developers/users to call `getAuthority` on Account object so they can get that authority and use it for making silent calls.

Fixed for the case of AAD by just grabbing from the issuer claim inside the ID Token. The authority return by `getAuthority` for the case of B2C will be wrong, and that will be fixed in a future PR.


msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -146,4 +147,17 @@ public String getUsername() {
 
         return MISSING_FROM_THE_TOKEN_RESPONSE;
     }
+
+    @Override
+    @NonNull
+    public String getAuthority() {
+        if (null != getClaims()) {
```
rpdome: nit: please add TODO here.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/AcquireTokenNetworkB2CTest.java
```diff
@@ -33,6 +34,11 @@
  */
 public class AcquireTokenNetworkB2CTest extends AcquireTokenNetworkTest {
 
+    @Override
+    public String getAuthority() {
+        return mApplication.getConfiguration().getDefaultAuthority().getAuthorityURL().toString();
```
rpdome: nit: also add TODO here.

msal/src/main/java/com/microsoft/identity/client/Account.java
shahzaibj: Why?

Are you referring to the case of B2C?

rpdome: yes. we should add TODO for any in-progress/partially finished work.

shahzaibj: Addressed

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/AcquireTokenNetworkB2CTest.java
shahzaibj: Addressed

msal/src/test/java/com/microsoft/identity/client/e2e/shadows/ShadowAuthority.java
```diff
@@ -99,7 +100,8 @@ public static Authority getAuthorityFromAuthorityUrl(String authorityUrl) {
             case AAD_MOCK_DELAYED_PATH_SEGMENT:
                 authority = new MockDelayedResponseAuthority();
                 break;
-            case B2C_TEST_PATH_SEGMENT:
+            case B2C_PATH_SEGMENT:
+            case B2C_PATH_SEGMENT_ALT:
```
iambmelt: 💪

Issue #850, opened 2019-12-05
no id_token in claims object


ApostolisApo:
**Describe the bug**
Have the same problem both in my app and in _msalandroidkotlinapp_ example app. I am filling in my Azure details and when handling _AuthenticationCallback_ in both:
 * the single fragment (_mSingleAccountApp!!.signIn()_ )
 * multiple account fragment (_mMultipleAccountApp!!.acquireToken()_ )

the _claims_ object does not contain _id_token_. The access token and everything else is acquired properly

**Smartphone**
 - Device: Samsung S8
 - Android Version: Android 8
 - Browser Chrome
 - MSAL 1.0

**To Reproduce**
Replace the details in  .json file with your Azure App details and run the _msalandroidkotlinapp_ example app.  Choose to sign in with in the single application fragment. If you inspect the url of the login page you can see that in the URL parameter there is this: _response_type=code_. Should it be like this? Shouldn't it be _response_type=token id_token_ ?
 

**Expected behavior**
An id_token should be present in the claims object

**Actual Behavior**
The id_token is missing from the claims object
The claims object is missing id_token even if I call IMultipleAccountPublicClientApplication.getAccount(id)

iambmelt:
Hi @ApostolisApo 

The `claims` object _is_ the Id Token --  it is a `Map` of the parsed JWT claims.

Is there a specific field within the IdToken you are trying to access? Please note, the raw token string is not currently exposed externally; if you have a good use case for needing the unparsed token we can evaluate adding this piece of API surface

ApostolisApo:
@iambmelt Yes actually the service I am building against is using the bare token, I cannot use the broken down claims. To reconstruct them to a string is quite hard right? 
Anyway, in your example app you do exactly this, you access the bare string value and your app crashes because of that. And that was the reason I tried so much in vain to access it in mine.
It would be awesome if you exposed it because currently I resorted to another, a little more low-level library :(

iambmelt:
@ApostolisApo No, unfortunately you cannot reconstitute the token from the claims as the jwt signature won't match unless it is signed with the keys from STS.

I am tagging this issue as a Feature Request and will follow up with the team re: offering an API to access the raw IdToken in a future release.

TaugyStyle:
Hello :) 
Having the same request here , need the id_token string to validate signature against a back-end service. Since the ios MSAL library returns  a json containing the string id_token and access_token it would be good to have more or less the same behavior.
[...]

PR #870, opened 2019-12-17
Exposes id_token via AuthenticationResult

iambmelt:
Re: #850

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/IClaimable.java
+++ b/msal/src/main/java/com/microsoft/identity/client/IClaimable.java
@@ -30,10 +30,13 @@ import java.util.Map;
 public interface IClaimable {
 
     /**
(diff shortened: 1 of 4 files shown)


msal/src/main/java/com/microsoft/identity/client/IClaimable.java
```diff
@@ -29,6 +29,15 @@
 
 public interface IClaimable {
 
+    /**
+     * Gets the IdToken associated with this IClaimable.
+     *
+     * Note: MSAL does not validate the JWT token
+     * @return
```
iambmelt: Fix these docs

iambmelt: 27148bc456f4e94aa8750e54fa649de174156572

Issue #869, opened 2019-12-17
[LocalMSALControler is not eligible to use the broker]


thangld8:
**Describe the bug**
From LoginActivity, After I get Account so I try to pass IAccount as a Params to my JobServices:
```java
   b2cApp.getAccounts(new IPublicClientApplication.LoadAccountsCallback() {
            @Override
            public void onTaskCompleted(final List<IAccount> result) {
                // parse IAccount to String
                result.get(0);
                Gson gson = new Gson();
                String resultJson = gson.toJson(result.get(0));
                // Save IAccount to Preference
                EncryptPreference preference = EncryptPreference.getInstance(getContext());
                preference.setB2cAccountInfo(resultJson);

                // create PersistableBundle and JobInfo
                PersistableBundle bundle = new PersistableBundle();
                bundle.putString("listAccount", resultJson);
                b2CUsers = B2CUser.getB2CUsersFromAccountList(result);
                ComponentName refreshAccessTokenService = new ComponentName(getContext(), MerchantRefreshAccessTokenServices.class);
                JobInfo jobInfoRefreshToken = new JobInfo.Builder(1, refreshAccessTokenService)
                        .setPeriodic(5000)
                        .setExtras(bundle)
                        .setPersisted(true)
                        .build();

                JobScheduler jobScheduler = (JobScheduler)getSystemService(JOB_SCHEDULER_SERVICE);
                int resultCode = jobScheduler.schedule(jobInfoRefreshToken);
                if (resultCode == JobScheduler.RESULT_SUCCESS) {
                    Log.d(TAG, "Job scheduled!");
                } else {
                    Log.d(TAG, "Job not scheduled");
                }
            }

            @Override
            public void onError(MsalException exception) {
                displayError(exception);
            }
        });
```
After that, I am trying to call acquireTokenSilentAsync from JobService of Android:
```java
String resultJson = jobParameters.getExtras().getString("listAccount");
            try {
                Gson gson = new Gson();
                IAccount iAccount = gson.fromJson(resultJson, IAccount.class);
                AcquireTokenSilentParameters parameters = new AcquireTokenSilentParameters.Builder()
                        .fromAuthority(B2CConfiguration.getAuthorityFromPolicyName(Constants.B2C_SINGIN_POLICY[0]))
                        .withScopes(B2CConfiguration.getScopes())
                        .forAccount(iAccount)
                        .withCallback(getAuthSilentCallback())
                        .build();

[...]

PR #874, opened 2019-12-18
Change sharedDevice message log level, update test app

### LINKED ISSUE #873: getDeviceMode log message should be at level: warning
Regarding [this error log message](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java#L397-L398)

This should probably be a `warning` (or `info`, at your discretion) and not an `error` -- additionally, there is a minor spelling error; message should read:

"**LocalMSALController** is not eligible to use the broker. Do not check sharedDevice mode and return false immediately."

Related:
#869

rpdome:
Fix #873 + #869

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
@@ -394,7 +394,7 @@ public class LocalMSALController extends BaseController {
     public boolean getDeviceMode(OperationParameters parameters) throws Exception {
         final String methodName = ":getDeviceMode";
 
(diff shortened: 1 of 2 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
```diff
@@ -395,7 +395,7 @@ public boolean getDeviceMode(OperationParameters parameters) throws Exception {
         final String methodName = ":getDeviceMode";
 
         final String errorMessage = "LocalMSALControler is not eligible to use the broker. Do not check sharedDevice mode and return false immediately.";
```
iambmelt: "Controller" is spelled incorrectly here -- let's include that fix too

PR #881, opened 2020-01-01
Introduce AuthorizationFragment

rpdome:
Requirement: 
- There's an ask to display MSAL AuthorizationActivity in a certain area on a screen. 

Proposed change:
- Extracts AuthorizationFragment from AuthorizationActivity
- Expose an API on MSAL to take in a fragment.
- If a fragment is provided, replace that fragment with AuthorizationFragment (instead of popping up an activity).

Limitation:
- This will only work with MSAL-only (no broker), WebView flow. (Which satisfies the ask).
- onBackPressed will not work when MSAL is in fragment mode. 

Verified with the sample app.
Also removed unused AuthenticationActivity and related tests.

Related PR in common: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/763


msal/src/main/java/com/microsoft/identity/client/IMultipleAccountPublicClientApplication.java
```diff
@@ -115,6 +116,31 @@ void acquireToken(@NonNull final Activity activity,
...
+     *                 MSAL always sends the scopes 'openid profile offline_access'.  Do not include any of these scopes in the scope parameter.
+     * @param callback The {@link AuthenticationCallback} to receive the result back.
+     *                 1) If user cancels the flow by pressing the device back button, the result will be sent
+     *                 back via {@link AuthenticationCallback#onCancel()}.
+     *                 2) If the sdk successfully receives the token back, result will be sent back via
+     *                 {@link AuthenticationCallback#onSuccess(IAuthenticationResult)}
+     *                 3) All the other errors will be sent back via
+     *                 {@link AuthenticationCallback#onError(MsalException)}.
+     */
+    void acquireToken(@NonNull final Fragment fragment,
```
kreedula: Why add new overloads ? I personally prefer to  overloads only if it's must else too many methods would effect the readability of public api's . We spent a lot of time cleaning up the overloads already and we introduced `AcquireTokenParameters` for this reason . Since we expose `Fragment` through `AcquireTokenParameters` the developers should be able to use that.

rpdome: Ok. Make sense since this is experimental feature also.

rpdome: removed.

shahzaibj:
I believe you had to make changes to the testapp to test this out. Can you commit those too? It might be helpful to use those during bug bashes.

rpdome:
> I believe you had to make changes to the testapp to test this out. Can you commit those too? It might be helpful to use those during bug bashes.

No. I had to modify the Sample app to verify this change. I could push my branch there for testing purpose. (I don't want to merge the change into dev/master - as this 'fragment' support is still 'experimental').

Our test app isn't using fragment.

PR #887, opened 2020-01-08
Fix checking for custom tab redirect activity + Update submodule + Update Readme

shahzaibj:
We have a bug in `checkIntentFilterAddedToAppManifestForBrokerFlow` method where if we detect that we do not have to use the broker and then we just return from the method and do not validate if we have the intent filter for BrowserTabActivity.


README.md
```diff
@@ -73,6 +73,8 @@ It's simplest to create your configuration file as a "raw" resource file in your
 }
 ```
 
+>NOTE: In the `redirect_uri`, the part `<YOUR_PACKAGE_NAME>` refers to the package name returned by the `context.getPackageName()` method. This package name is the same as the [`application_id`](https://developer.android.com/studio/build/application-id) defined in your `build.gradle` file.
```
iambmelt: 👍

PR #889, opened 2020-01-08
Fix null pointer exception in BrokerMsalStrategies

rpdome:
Related PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/772

This fixes AzureAD/microsoft-authentication-library-common-for-android#770

Root cause:
- verifyHelloFromResultBundle returned boolean.
- In BrokerMsalController, since it returned boolean, exception is never thrown for that strategy.
- if hello failed twice - for BindService and AccountManager, then we'll end up with result = null.

Fix:
1. verifyHelloFromResultBundle() now throw an exception, if it gets an unexpected result bundle (i.e. Broker doesn't support hello in the calling strategy, which could be BindService or AccountManager).
 
2. Also introduce BrokerCommunicationException, which is thrown when
 - We ran into bind service errors in BrokerAuthServiceStrategy
 - We ran into AccountManager error in BrokerAccountManagerStrategy

3. In BrokerMsalController
 - When we loop through strategies, if BrokerCommunicationException is thrown, we continue, otherwise we throw an exception right away.
 - After the loop is done, if result = null, it means we got BrokerCommunicationException for EVERY strategies we iterated through. In this case, we throw BrokerCommunicationException.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
@@ -155,7 +155,7 @@ public class BrokerMsalController extends BaseController {
          * If this value returns null, no telemetry event will be emitted.
          */
         @Nullable
-        String getTelemetryApiName();
+        String getTelemetryApiId();
 
         /**
          * A method that will be invoked before the success event is emitted.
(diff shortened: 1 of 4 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -188,33 +186,42 @@ public AcquireTokenResult acquireToken(AcquireTokenOperationParameters parameter
...
+        }
+
+        // This means that we've tried every strategies...
+        if (result == null) {
+            final BrokerCommunicationException exception = new BrokerCommunicationException("MSAL failed to communicate to Broker.");
+            if (strategyTask.getTelemetryApiName() != null) {
+                Telemetry.emit(
+                        new ApiEndEvent()
+                                .putException(exception)
+                                .putApiId(strategyTask.getTelemetryApiName())
```
shahzaibj: is telemetry api name the same as api id?

rpdome: renamed to getTelemetryApiId() for consistency.

PR #891, opened 2020-01-10
Merge the provided authority/clientid with the config object. Add redirectUri as part of PCA factory overload.

### LINKED ISSUE #890: Hard crash when clientId is null
When calling PublicClientApplication.create(context, clientid, listener) w/o going through the typical example of providing a config resource file, the app has a hard crash because the OperationalParameters are not being updated with the provided clientId, so line 252 in OperationParameters.java (getClientId().hashCode()) is bombing due to a null exception.  Ideally, at least from my perspective, in the main create() in PublicClientApplication.java after (or during) line 854 (params = OperationParamatersAdapter.createOperationParameters) the clientId and authority should be replaced with any value provided to the method...in addition to sanity checks in OperationParameters.

rpdome:
Fix #890 

Two issues here.
1. PCA is not honoring authority/clientID before it makes a call to get device mode. Since no configuration file is passed. clientID will be null and exception is thrown.
2. redirectURI is also not there. Given that there are more than one default redirectURI format (broker Redirect URI was introduced as part of MSAL 1.0), we decided that this should be passed explicitly (instead of trying to prefill it).

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -183,6 +183,7 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
         static final String CALLBACK = "callback";
         static final String CLIENT_ID = "client_id";
         static final String AUTHORITY = "authority";
+        static final String REDIRECT_URI = "redirect_uri";
         static final String CONFIG_FILE = "config_file";
         static final String ACTIVITY = "activity";
         static final String SCOPES = "scopes";
@@ -233,8 +234,7 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
      * @param listener             a callback to be invoked when the object is successfully created.
      *                             Cannot be null.
      * @see PublicClientApplication#create(Context, File, ApplicationCreatedListener)
-     * @see PublicClientApplication#create(Context, String, ApplicationCreatedListener)
-     * @see PublicClientApplication#create(Context, String, String, ApplicationCreatedListener)
+     * @see PublicClientApplication#create(Context, String, String, String, ApplicationCreatedListener)
      * @see PublicClientApplication#create(Context, int)
      */
     public static void create(@NonNull final Context context,
@@ -247,6 +247,7 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
                 initializeConfiguration(context, configFileResourceId),
                 null, // client id
                 null, // authority
+                null,
                 listener
         );
     }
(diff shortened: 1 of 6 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -247,6 +247,7 @@ public static void create(@NonNull final Context context,
                 initializeConfiguration(context, configFileResourceId),
                 null, // client id
                 null, // authority
+                null,
```
kreedula: nit: add `// redirectUri` as a comment for readability like above

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -285,44 +285,8 @@ public static void create(@NonNull final Context context,
...
-        validateNonNullArgument(context, NONNULL_CONSTANTS.CONTEXT);
-        validateNonNullArgument(clientId, NONNULL_CONSTANTS.CLIENT_ID);
-        validateNonNullArgument(listener, NONNULL_CONSTANTS.LISTENER);
-
-        create(
-                initializeConfiguration(context),
-                clientId,
-                null, // authority
+                null, // authority,
+                null,
```
kreedula: same

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -849,8 +815,23 @@ private static ISingleAccountPublicClientApplication createSingleAccountPublicCl
     private static void create(@NonNull final PublicClientApplicationConfiguration config,
                                @Nullable final String clientId,
                                @Nullable final String authority,
+                               @Nullable final String redirectUri,
                                @NonNull final ApplicationCreatedListener listener) {
+        if (clientId != null) {
```
kreedula: why `null` checks here, since we are already validating in public api's above.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -768,6 +733,7 @@ private static IPublicClientApplication create(
         final ResultFuture<AsyncResult<IPublicClientApplication>> future = new ResultFuture<>();
         create(configuration,
                 null, // client id
+                null,
```
kreedula: same

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
rpdome: Some API are not passing this value (i.e. the one that only takes config file).

rpdome: Basically, this function is an 'end point' of all PCA factory API.

kreedula: shouldn't we throw here if it's null then?

kreedula: you mean we will get from config if it's null here ?

rpdome: yes

rpdome: for this function, config is @Nonnull. The caller is responsible for generating the default config file if it's not passed by the app.

PR #895, opened 2020-01-14
Fix snapshot api for common

shahzaibj:
The nightly snapshot builds are failing as I didn't append `-SNAPSHOT` to common snapshot api post msal 1.1.0-hf1 deployment.

Please note that we also had a bug in our versionSnapshot task and hence this PR depends on[ common pr 775](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/775) to be merged in prior to merging this.


Om83:
@shahzaibj , Why do we need this change?

shahzaibj:
> @shahzaibj , Why do we need this change?

@Om83 After every release, when we merge the release branch back to `dev`, we need to update the version of the snapshot api and append "-SNAPSHOT" at the end. This is needed for nightly snapshot builds to work properly as the common snapshot artifact produced has the suffix "-SNAPSHOT" at the end. I forgot to make this change post `msal 1.1.0-hf1 release` and hence the nightly builds were failing for the last few days.

Please note that this PR depends on [ common pr 775](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/775) to be merged in first as we did have a bug in our `snapshotVersion` task.

PR #900, opened 2020-01-22
Do not throw any exception until we've tried all strategies

rpdome:
I found out today that we cannot be so sure what kind of exception would be thrown when bind service fails (i.e. in a variant of Samsung S8, NullPointerException is thrown).

For that particular device, AccountManager method would work (forced through android debugger), but the current code will never trigger that as an exception will be thrown before it tries the next strategy.

The fix here is to try every strategy, and then only throw if all strategies fail.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAccountManagerStrategy.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAccountManagerStrategy.java
@@ -33,9 +33,9 @@ import android.os.Bundle;
 import androidx.annotation.NonNull;
 import androidx.annotation.WorkerThread;
 
+import com.microsoft.identity.client.exception.BrokerCommunicationException;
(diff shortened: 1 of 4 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAccountManagerStrategy.java
```diff
@@ -135,7 +134,7 @@ void hello(@NonNull final OperationParameters parameters)
             throws BaseException {
 
         if (!BrokerMsalController.isAccountManagerPermissionsGranted(parameters.getAppContext())) {
-            throw new BrokerCommunicationException("Account manager permissions are not granted");
+            throw new BrokerCommunicationException("Account manager permissions are not granted", null);
```
iambmelt: Minor nit regarding comments: flips between `AccountManager` and `Account manager`

PR #907, opened 2020-01-30
First Draft, AT PoP Docs (Preview)

iambmelt:
Uploads 1-Pager for AT/PoP integration -- this is a preview feature

### DIFF
--- a/docs/access_token-pop.md
+++ b/docs/access_token-pop.md
@@ -12,6 +12,8 @@ By default, Bearer access tokens issued by AAD have a 1 hour validity. The valid
 
 When an SHR is created by the client, a timestamp (`ts`) claim is embedded in the JWT. The resource middleware will, upon receipt of the token, inspect its signature and timestamp to ensure integrity (anti-tamper protection) and validity (non-expiry). By default, the SAL middleware honors a 5 minute validity period meaning that once an SHR been signed by the client it may be used for 5 minutes before the resource will require a newly signed access token.
 
+>**Note:**<br/>`AuthenticationResult#getExpiresOn()` does _not_ indicate the expiry/validity period configured on the resource middleware for Signed HTTP Requests. Instead, it reflects the configured expiry period for access tokens emitted by AAD. As such, this value cannot be used to determine whether or not an SHR will be accepted by the middleware. Because of this, `getExpiresOn()` cannot be used to signal cache freshness/eviction criteria; therefore it is recommended that users of MSAL call `acquireTokenSilent()` each time a new SHR is required.
+
 ## Can I continue to use Bearer flows in my application?
 Yes! PoP and Bearer flows may be used interchangeably with MSAL and with supported broker versions **so long as the targeted resource supports it**.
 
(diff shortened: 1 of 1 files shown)


docs/access_token-pop.md
```diff
@@ -0,0 +1,99 @@
...
+    // ex: PoP eyJhbGciOiJSU(...)
+    final String authorizationHeader = authenticationResult.getAuthorizationHeader();
+}
+```
+
+#### Get the SHR
+```java
+@Override
+public void onSuccess(final IAuthenticationResult authenticationResult) {
+    // Uses the existing MSAL method, this method returns the SHR minus and scheme-prefix.
```
iambmelt: `minus any`

docs/access_token-pop.md
```diff
@@ -0,0 +1,99 @@
...
+}
+```
+
+#### Get the auth scheme
+```java
+@Override
+public void onSuccess(final IAuthenticationResult authenticationResult) {
+    // Returns "Bearer", "pop", or whichever scheme was used to acquire this token
+    // Note that authentication scheme prefixes are case insensitive
+    final String shr = authenticationResult.getAuthenticationScheme();
```
iambmelt: variable name is wrong (thx @shahzaibj)

docs/access_token-pop.md
iambmelt: Resolved in ffbce7315819aca55e3af0e362cca84445d4163f

docs/access_token-pop.md
iambmelt: Resolved in ffbce7315819aca55e3af0e362cca84445d4163f

docs/access_token-pop.md
```diff
@@ -0,0 +1,99 @@
...
+
+Proof of Possession (PoP) increases the security posture of these tokens embedding them inside of a JWT envelope and signing (binding) that JWT with RSA key material. The key material is generated on the device which was originally issued the tokens and never leaves it. The resulting JWT is called the Signed HTTP Request (SHR). This binding renders the access token unusable without a recent proof for the target resource endpoint.
+
+## How long is a PoP/AT valid?
+By default, Bearer access tokens issued by AAD have a 1 hour validity. The validity of the Signed HTTP Request may be shorter, depending on the resource middleware configuration.
+
+When an SHR is created by the client, a timestamp (`ts`) claim is embedded in the JWT. The resource middleware will, upon receipt of the token, inspect its signature and timestamp to ensure integrity (anti-tamper protection) and validity (non-expiry). By default, the SAL middleware honors a 5 minute validity period meaning that once an SHR been signed by the client it may be used for 5 minutes before the resource will require a newly signed access token.
```
rpdome: Do you think if we should call out that the one we expose in AcquireTokenResult is not going to match this 5 minute value (and they shouldn't rely on that - instead, they should always be calling MSAL to get a newly-minted token instead of caching it)

iambmelt: Good call out -- I'll add a little blurb on this

iambmelt: Blurb added in afbe5c921f19b9900eba49a4b6e8cd7553b368f2

PR #932, opened 2020-02-12
[TestApp] Use same redirect uri for both AAD and B2C msal config

shahzaibj:
The word "microsoft" is not allowed in a redirect_uri for an app that supports personal MSA accounts.

### DIFF
--- a/testapps/testapp/src/main/res/raw/msal_config.json
+++ b/testapps/testapp/src/main/res/raw/msal_config.json
@@ -8,7 +8,7 @@
     {
       "type": "AAD",
       "audience": {
-        "type": "AzureADandPersonalMicrosoftAccount"
+        "type": "AzureADMultipleOrgs"
       }
     }
   ],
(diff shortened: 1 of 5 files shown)


testapps/testapp/src/main/res/raw/msal_config.json
```diff
@@ -1,14 +1,14 @@
...
   "authorization_user_agent" : "DEFAULT",
-  "redirect_uri" : "msauth://com.microsoft.identity.client.sample.local/1wIqXSqBj7w%2Bh11ZifsnqwgyKrY%3D",
+  "redirect_uri" : "msauth://com.msft.identity.client.sample.local/1wIqXSqBj7w%2Bh11ZifsnqwgyKrY%3D",
   "multiple_clouds_supported":true,
   "client_capabilities": "CP1",
   "authorities" : [
     {
       "type": "AAD",
       "audience": {
-        "type": "AzureADandPersonalMicrosoftAccount"
```
shahzaibj: This was not correct, this app didn't support personal account. I've added a new (app) config below that supports both organization and personal accounts.

PR #944, opened 2020-02-24
Update test app

rpdome:
**UI Changes**
- Replace "Environment" and "User Agent" with a picker that allows you to set a "config file".
- Add configs for sovereign clouds.
   - I've created new app registrations in each cloud with the IDLAB account for temporary use. Will need your help here to get the actual ones @shahzaibj.
- Add "Auth Scheme" spinner.
- Add "Authority" text field.
- In the "Account" section.
   - Add "-- No Account Selected --" as the last option in the list.
   - If an account is selected
       - the login_hint text field will be grayed out.
       - the authority text field will display a hint that if this value is not provided, the account's authority will be used.
- Add buttons for copying logs & acquireTokenResult to clipboard.

**Other Fixes**
- Clean up the codes. 
   - Move unnecessary code in MainActivity back to AcquireTokenFragment.
   - MsalWrapper now only init and load accounts only when necessary.
   - Extract Single Account and Multiple Account logic out.
- Fix crashes that could happen during navigation.
- Fix the issue where default browser is not displayed.

Screen Shot 2020-02-23 at 5 20 14 PM


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/ResultFragment.java
```diff
@@ -69,6 +74,18 @@ public View onCreateView(final LayoutInflater inflater, final ViewGroup containe
...
         previousAccessToken = accessToken;
+
+        mCopyAuthResultButton = (Button) view.findViewById(R.id.btn_copyAuthResult);
+        mCopyAuthResultButton.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                ClipboardManager clipboard = (ClipboardManager)
+                        getActivity().getSystemService(Context.CLIPBOARD_SERVICE);
+                ClipData clip = ClipData.newPlainText("Auth Result", mTextView.getText());
+                clipboard.setPrimaryClip(clip);
```
kreedula: This is great 👍

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/RequestOptions.java
```diff
@@ -0,0 +1,80 @@
+package com.microsoft.identity.client.testapp;
```
kreedula: Copyright

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MultipleAccountModeWrapper.java
```diff
@@ -0,0 +1,85 @@
+package com.microsoft.identity.client.testapp;
```
kreedula: Copyright here and everywhere else

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
```diff
@@ -90,103 +99,128 @@ public View onCreateView(final LayoutInflater inflater, final ViewGroup containe
...
+                    mLoginhint.setEnabled(true);
+
+                    mAuthority.setHint("");
+                } else {
+                    mLoginhint.setText("");
+                    mLoginhint.setHint("Disabled");
+                    mLoginhint.setCursorVisible(false);
+                    mLoginhint.setEnabled(false);
+
+                    mAuthority.setHint("For ATS: Selected account's authority will be used if nothing is provided here");
```
kreedula: nit : trim to "default: account's authority"

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalWrapper.java
```diff
@@ -166,9 +85,13 @@ public void acquireToken(final Activity activity,
...
 
-        if (USE_POP) {
+        if (requestOptions.getAuthority() != null && !requestOptions.getAuthority().isEmpty()) {
+            builder.fromAuthority(requestOptions.getAuthority());
+        }
+
+        if (requestOptions.getAuthScheme() == Constants.AuthScheme.POP) {
             try {
                 builder.withAuthenticationScheme(
                         PoPAuthenticationScheme.builder()
```
iambmelt: Can we add another `Spinner` and `EditText` to accept the `HttpMethod` and `Resource URL`? These could _optionally_ be shown only when the Auth Scheme `Spinner` indicates `PoP`

rpdome: sure np

PR #949, opened 2020-02-26
Add msa only config file

shahzaibj:


testapps/testapp/src/main/res/raw/msal_config_msa_only.json
```diff
@@ -0,0 +1,13 @@
+{
+  "client_id" : "9668f2bd-6103-4292-9024-84fa2d1b6fb2",
+  "authorization_user_agent" : "DEFAULT",
+  "redirect_uri" : "msauth://com.msft.identity.client.sample.local/1wIqXSqBj7w%2Bh11ZifsnqwgyKrY%3D",
+  "authorities" : [
+    {
+      "type": "AAD",
```
rpdome: this is confusing lol. AAD but MSA only.

shahzaibj: I couldn't think of what to call it. lol

It is indeed technically AAD + MSA but we restrict authority to MSA only. Which could then be overridden by the developer in AT parameters. Let me know if you think of a better name for this config file

rpdome: I don't think it's a big deal :) (hence I signed off).

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/Constants.java
```diff
@@ -39,7 +39,8 @@
         INSTANCE_AWARE_COMMON,
         INSTANCE_AWARE_ORGANIZATION,
         B2C,
-        MSA
+        MSA,
```
Om83: We should consider this making more dynamic through selection in the app and then generating the json on the fly. That would allow us to generate new combinations on need basis rather than having to checkin a file for a specific case.

shahzaibj: I'll investigate later what are the options for this case. Going to merge this for now.

PR #951, opened 2020-02-27
Add browser skip broker config file

rpdome:


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/Constants.java
```diff
@@ -47,9 +49,15 @@ public static int getResourceIdFromConfigFile(ConfigFile configFile){
...
 
+            case BROWSER_SKIP_BROKER:
+                return R.raw.msal_config_browser_skip_broker;
+
             case WEBVIEW:
                 return R.raw.msal_config_webview;
 
+            case WEBVIEW_SKIP_BROKER:
+                return R.raw.msal_config_webview_skip_broker;
+
```
Om83: This will result in numerous jsons to be created based on various combinations.
For a more flexible approach we should split the options in its own category e.g. whether or not use Broker . 
This can then be combined with other configuration values and generate the json on fly based on selection.

Om83: I am fine for now. But for future we should consider this change.

rpdome:
Probably won't need this now. Closing.

PR #962, opened 2020-03-05
Logging sample

shahzaibj:


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalSampleApp.java
```diff
@@ -40,27 +46,39 @@
...
             @Override
             public void log(String tag, Logger.LogLevel logLevel, String message, boolean containsPII) {
-                // contains PII indicates that if the log message contains PII information. If Pii logging is
-                // disabled, the sdk never returns back logs with Pii.
-                mLogSize = mLogs.toString().getBytes().length;
-                if (mLogSize + message.getBytes().length >= LOG_SIZE) {
-                    clearLogs();
-                }
-                mLogs.append(message).append('\n');
+                writeToLogFile(getApplicationContext(), "TAG: " + message);
```
shoatman: Looks like we didn't log the tag value.

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalSampleApp.java
```diff
@@ -40,27 +46,39 @@
...
+    private synchronized void writeToLogFile(Context ctx, String msg) {
+        File directory = ctx.getDir(ctx.getPackageName(), Context.MODE_PRIVATE);
+        File logFile = new File(directory, "msallogs.txt");
+        FileOutputStream outputStream = null;
+        try {
+            outputStream = new FileOutputStream(logFile, true);
+            OutputStreamWriter osw = new OutputStreamWriter(outputStream);
+            osw.write(msg);
+            osw.flush();
+            osw.close();
```
shoatman: Should we have a finally block for closing the stream.

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalSampleApp.java
```diff
@@ -40,27 +46,39 @@
...
-    private static final int LOG_SIZE = 1024 * 1024;
     private StringBuilder mLogs;
-    private int mLogSize;
+
+    private synchronized void writeToLogFile(Context ctx, String msg) {
+        File directory = ctx.getDir(ctx.getPackageName(), Context.MODE_PRIVATE);
+        File logFile = new File(directory, "msallogs.txt");
+        FileOutputStream outputStream = null;
+        try {
+            outputStream = new FileOutputStream(logFile, true);
```
shoatman: Are going to rollover this file when it becomes too large?  I'd be concerned about someone taking this code as is and copying it in to their application.

shahzaibj:
@shoatman This is just a quick sample I made for a customer that wanted to know how to use logger and callbacks. The primary motive here was to demonstrate how the callback works and I am not concerned about the specifics of writing to file. I am not actually planning on taking this change and going to close this PR. The test app already has a basic implementation of logger but if you think we need a more robust sample then I think we can create one on the README page or perhaps in the Android Azure Sample app. Let me know your thoughts.

PR #976, opened 2020-03-18
Fix #967

### LINKED ISSUE #967: Cache Migration not performed in below scenarios. See Description.
- Single Account
- Multiple Account - getAccount(id)
- Multiple Account - getAccounts - if an interactive call was already made post adoption of MSAL

shahzaibj:
Fix #967


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1821,4 +1832,41 @@ public void onError(MsalException exception) {
             throw result.getException();
         }
     }
+
+    void performMigration(@NonNull final TokenMigrationCallback callback) {
+        // Create the SharedPreferencesFileManager for the legacy accounts/credentials
```
rpdome: nit: move these inside "else"?

shahzaibj: Addressed

PR #979, opened 2020-03-23
Picks up Common w/ PoP Clock-Skew Sync

iambmelt:
Expands:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/864
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/868

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java
@@ -97,6 +97,7 @@ public class OperationParametersAdapter {
 
         acquireTokenOperationParameters.setAuthenticationScheme(
                 AuthenticationSchemeFactory.createScheme(
+                        acquireTokenParameters.getActivity(),
                         acquireTokenParameters.getAuthenticationScheme()
                 )
         );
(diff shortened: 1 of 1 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/controllers/OperationParametersAdapter.java
```diff
@@ -97,6 +97,7 @@ public static AcquireTokenOperationParameters createAcquireTokenOperationParamet
 
         acquireTokenOperationParameters.setAuthenticationScheme(
                 AuthenticationSchemeFactory.createScheme(
+                        acquireTokenParameters.getActivity(),
```
shahzaibj: Is there a `createScheme` method that takes an `Activity`? I can't find it.

iambmelt: `Activity` is a subclass of `Context`

iambmelt: Btw I am going to clean up this constructor a little bit to remove the hard reference to the Context and instead include a reference to the `IClockSkewManger` -- will be pushed up momentarily

iambmelt: Cleaned up here:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/868/commits/dff13326e369b599f0ea4aa483c401431b56076b

Issue #982, opened 2020-03-26
Proguard issue: Authentication fails with release/minified version


slowcar:
I have implemented the MSAL library in our Android application.
It works properly when debugging, but i have issues getting the release version to work.
When i disable proguard (minifyEnabled false) the authentication is successful in the release version, otherwise not.

I have added the lines mentioned in https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/513#issuecomment-543064962 and this resolved the app crashing at launch, but i still get an error after authentication in the browser:

![error](https://user-images.githubusercontent.com/28760605/77642311-bc4c9a80-6f5d-11ea-825e-6e8ad5563d63.jpg)

Is there an official guide regarding proguard settings? Any idea how to resolve this?

shahzaibj:
@slowcar It seems like the client did not send the `grant_type` parameter in the request to the server. Looking it the server side logs, it is not immediately clear why the client might not have sent this parameter as part of the request. We would need client side logs for complete investigation. Can you share MSAL logs with me? Here are the instructions for capturing logs in MSAL: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-logging?tabs=android

iambmelt:
> @slowcar It seems like the client did not send the `grant_type` parameter in the request to the server. Looking it the server side logs, it is not immediately clear why the client might not have sent this parameter as part of the request. We would need client side logs for complete investigation. Can you share MSAL logs with me? Here are the instructions for capturing logs in MSAL: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-logging?tabs=android

@shahzaibj This property was likely minified away by ProGuard and so Gson serialization did not send it -- the easiest way to test this is to remove proguard from your release build and test it. Similar issues have been reported [here ](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/513#issuecomment-543064962)and [here](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/613)

@slowcar Perhaps [this is the guidance](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/613#issuecomment-494557367) you're looking for? (Sorry, I haven't personally tested this)

slowcar:
Thanks for the quick reply. I think @iambmelt is correct that the property is most likely minified away. I already tested the release without proguard and that works.
[...]

Issue #613, opened 2019-05-21
com.microsoft.identity.client.exception.MsalServiceException: AADSTS900144: The request body must contain the following parameter: 'grant_type'.


NLLAPPS:
Hi, I have suddenly started to receive the following error at PublicClientApplication.acquireToken 
 as soon as interactive authentication login completed.
I am using 0.3.1-alpha

`com.microsoft.identity.client.exception.MsalServiceException: AADSTS900144: The request body must contain the following parameter: 'grant_type'.`

No code changed. Was there some sort of server side change?

shoatman:
Not that I'm aware of.  If you're logging... possible to provide the correlation id of the request.  I can look it up on the server.

NLLAPPS:
It might be related to R8. I have minification enabled and using Latest Android Studio Beta. I am able to authenticate fine without pro guard.
I'll try to find what I need to exclude from minification

NLLAPPS:
-keep class com.microsoft.** { *; }  solves the issue with R8/minification but it is too broad. Any ideas about what to keep?

shoatman:
We'll take a look.  Thanks for raising this.

NLLAPPS:
Thank you. I have narrowed it down to 
`com.microsoft.identity.** { *; }`
It is enough for me for now

NLLAPPS:
I was able to drill up to
`-keep class com.microsoft.identity.common.internal.providers.microsoft.** { *; }`

RDLemos:
Hello,
I got the exact same problem... this is the error I receive on the release version of my app. In the Debug version, it works fine:
05-30 18:11:56.205 22064 22064 D circuit : [MSConnectorManager] Authentication failed: com.microsoft.identity.client.exception.MsalServiceException: AADSTS900144: The request body must contain the following parameter: 'grant_type'.

iambmelt:
Tagging as an enhancement to provide a ProGuard config / documentation thereabout

iambmelt:
Closed by #987

Issue #513, opened 2019-01-29
Proguard not supported by default


guperrot:
When using proguard + msal, the following rules are required to make the client work:

```
# This enum is accessed via reflection in common msal lib thus proguard removes it.
-keep public enum com.microsoft.identity.common.internal.ui.AuthorizationAgent  {
    *;
}

# Classes for below package is used by MSAL library. Ignore warning coming from the dependency.
-dontwarn com.nimbusds.jose.**
```

This rules can be packaged directly in the msal library and common library instead of asking developers to do so.

mosess:
Another missing rule that was causing some problems to me:
```
-keep class com.microsoft.identity.common.** {
  *;
}
```

guperrot:
But it would keep every single class from common with this global rule, defeating the use of Proguard altogether. The fix should be more targeted, what was the issue exactly @mosess?.

mosess:
You're right, it should be more targeted.
I thought that maybe the recommendation of ADAL might be relevant for this library as well, as maybe more issues may arise in other functionalities when making the specific problem work.

The specific error was:
`class com.microsoft.identity.common.internal.providers.a.b.j declares multiple JSON   fields named c`

guperrot:
Quick fix is likely target to `class com.microsoft.identity.common.internal.providers.**` which is narrower scope yet not ideal.

If you can look at the proguard mapping of your application and see what is a.b.j (or with the retrace tool) maybe we can figure out better rules.

mosess:
I agree, that's not ideal.
The ideal solution would be Microsoft mapping the exact parts of the library that might break when using Proguard and define it similar to the ADAL library.
For now, I used the general solution just to make it work.

joshfriend:
New versions of proguard and all versions of R8 will allow you to ship proguard rules inside of an AAR package. [Here's an example of how that is accomplished](https://github.com/JakeWharton/ThreeTenABP/pull/12)

guperrot:
We can already do it in proguard, the Microsoft AppCenter SDK uses this already.

guperrot:
This is what came up when fixing the rules when used in AppCenter:

https://github.com/microsoft/AppCenter-SDK-Android/pull/1149/files#diff-54abb45c98c994c8e22a56dc46f07a0b

This covers basic AAD B2C scenarios but you might want to apply that keep rules to every model or enum used via Gson.

guillaume-tgl:
[...]

PR #987, opened 2020-03-28
Adds basic proguard documentation

iambmelt:
We've had a lot of asks about how to do this:
- https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/982
- https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/513
- https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/613

This PR publishes the exclusions developers need to include in their project if they don't want to hit serialization errors and related issues due to obfuscation/minification

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,48 @@ if(account != null){
 
 ```
 
+## ProGuard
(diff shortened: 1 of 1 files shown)


guillaume-tgl:
Is there a reason why you don't include a Proguard file directly in your SDK like all modern Android libraries rather than documentation?
See for example https://proandroiddev.com/handling-proguard-as-library-developer-or-in-a-multi-module-android-application-2d738c37890
Or the official doc: https://developer.android.com/studio/build/shrink-code (Search for Library Dependency in the table)

PR #989, opened 2020-03-30
Adding default proguard config to project repo, relocating from readme

iambmelt:
Based on feedback in #987 - moving this config out of readme and including by default

related:
- #982 
- #513 
- #613


README.md
```diff
@@ -205,45 +205,6 @@ if(account != null){
 ## ProGuard
 MSAL uses reflection and generic type information stored in `.class` files at runtime to support various persistence and serialization related functionalities. Accordingly, library support for minification and obfuscation is limited. For reliability, we recommend developers exclude the Microsoft package from ProGuard.
```
shahzaibj: Do we still need this bit in README?

iambmelt: Only if you (or others) think it provides useful information as to why the proguard config is the way it is.

I have no strong feelings about including this if it's just noise

shahzaibj: I think we should remove it (or at least remove the last statement) because it seems like we are still asking developers to exclude MSAL classes within their ProGaurd file, but I believe developers no longer need to that since we have included our own ProGaurd file, right?

iambmelt: Updated in 79d2152a68852c70c8cb7a4cf72dd6377d32524f

How does that sound now?

shahzaibj: Yeah that sounds good!

iambmelt:
@rpdome 

Is this expected?
```
* What went wrong:
Could not resolve all files for configuration ':common:debugCompileClasspath'.
> Could not resolve com.microsoft.device.display:display-mask:0.3.0.
  Required by:
      project :common
   > Could not resolve com.microsoft.device.display:display-mask:0.3.0.
      > Could not get resource 'https://identitydivision.pkgs.visualstudio.com/_packaging/AndroidADAL/maven/v1/com/microsoft/device/display/display-mask/0.3.0/display-mask-0.3.0.pom'.
         > java.lang.NullPointerException (no error message)
   > Could not resolve com.microsoft.device.display:display-mask:0.3.0.
      > Could not get resource 'https://identitydivision.pkgs.visualstudio.com/_packaging/AndroidADAL/maven/v1/com/microsoft/device/display/display-mask/0.3.0/display-mask-0.3.0.pom'.
         > java.lang.NullPointerException (no error message)
```

msal/proguard-rules.pro
```diff
@@ -15,3 +15,36 @@
 #-keepclassmembers class fqcn.of.javascript.interface.for.webview {
 #   public *;
 #}
+
+##---------------Begin: proguard configuration for MSAL  --------
+-keep class com.microsoft.** { *; }
```
guillaume-tgl: It's not really the idea behind a Proguard configuration file to keep *all* files of the library. You're supposed to only keep classes that are accessed through JNI or reflection: 
https://developer.android.com/studio/build/shrink-code#keep-code

Otherwise, it removes the purpose of shrinking and obfuscation.

README.md
```diff
@@ -203,46 +203,7 @@ if(account != null){
...
-
-# Prevent R8 from leaving Data object members always null
--keepclassmembers,allowobfuscation class * {
-  @com.google.gson.annotations.SerializedName <fields>;
-}
-
-##---------------Begin: proguard configuration for Nimbus  ----------
--keep class com.nimbusds.** { *; }
-```
+MSAL uses reflection and generic type information stored in `.class` files at runtime to support various persistence and serialization related functionalities. Accordingly, library support for minification and obfuscation is limited. A default configuration is shipped with this library; please [file an issue](https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/new/choose) if you find any issues.
```
joshfriend: In order for proguard/R8 to be able to pick up and use these default rules you have to reference your `.pro` file(s) from the `build.gradle` file. Specifying `consumerProguardFiles` for the `release` build type is what does this, see this example:

https://github.com/JakeWharton/ThreeTenABP/pull/12/files#diff-91f078d04c55297c16043badd6a8b960R33-R38

joshfriend: another example from the appcenter sdk: https://github.com/microsoft/appcenter-sdk-android/blob/f80899948cc9769cb4f4909e357bbef64c5ffc1e/sdk/build.gradle#L38

msal/proguard-rules.pro
iambmelt: Sure - unfortunately I don't have the test resources to verify that all components in the `com.microsoft.*` namespace can be shrunk and still behave normally, as we have some dependencies in this namespace that are external to our team. A full test pass takes a few days and a few people

Maybe at some point in the future this can be done in an exact way, but this is today's point-in-time approach

README.md
iambmelt: Great -- I've incorporated this feedback in 7437a3f552c020183d1070be30407c83b1b8c39c

Thanks for taking a look!

iambmelt: I'll leave this PR open until tomorrow before I merge it up -- if you have any other feedback happy to incorporate it

joshfriend:
❤️

Issue #991, opened 2020-04-01
Single Account PCA - calling signIn twice does not prompt for credentials both times


shoatman:
**Describe the bug**
App is in development.  We are calling signIn on single account public client application in order to force the user to login in again when the app re-opens.

**Smartphone (please complete the following information):**
Not a device specific issue.  

**Stacktrace**
None

**To Reproduce**
Call signIn method twice.  The user is not prompted to enter their credentials twice.

**Expected behavior**
When signIn method is called twice.  The user is always expected to be prompted.

**Actual Behavior**
The user is not prompted to enter their credentials again.  (NOTE: from Shane: This is likely due to cookies in the browser/embedded web view.

iambmelt:
There's an additional issue in the method signature: the `login_hint` optional argument is listed as `@NonNull`

https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java#L84

Will also fix in addressing this issue

iambmelt:
Confirmed that an Exception is thrown when `signIn` is called twice.

An empty (no detail message) `MsalClientException` is thrown with the error_code `invalid_parameter`

iambmelt:
>Expected behavior
When signIn method is called twice. The user is always expected to be prompted.

Doesn't this break our browser SSO story? What about apps that have already granted consent? First parties?

iambmelt:
Digging into this deeper, it looks possible to bypass adding just 1 account in SingleAccountMode if the broker is being used:
- Install broker
- Install app
- Sign-in to app (credentials added to broker)
- Uninstall app **without** calling `removeAccount()`
- Reinstall app
- Call `signIn()` again

Chomui:
I have the same bug. I added a parameter to a JSON file:
 "prompt" : "LOGIN"
But if try to login for second time - i receive an exception "invalid_parameter"
If i logout before login, i can pick an account instead of prompt to enter credentials again.

The only way - CookieManager.getInstance().removeAllCookies(null)
But it's not a solution, because i don't want to clear all WebView cookies

Chomui:
As i can see there is already pull request. I hope i can use it soon

iambmelt:
> I have the same bug. I added a parameter to a JSON file:
> "prompt" : "LOGIN"
> But if try to login for second time - i receive an exception "invalid_parameter"
[...]

PR #992, opened 2020-04-02
SA-PCA: New overloads, javadoc cleanup

iambmelt:
See #991 

## Changes in this PR:
- Changes a `@NonNull` annot to `@Nullable` to match 'Optional' notes in javadoc
- Adds `final` to some params, instance vars
- Tweaks some formatting to remove
- Changes 'nil' to 'null' in javadocs
- Adds new `signIn()` overload which adds a `prompt` param
- Adds new `reauthorize()` method which lets your set a `prompt` for the already-signed-in account

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
@@ -20,7 +20,6 @@
 //  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 //  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 //  THE SOFTWARE.
-
 package com.microsoft.identity.client;
 
 import android.app.Activity;
@@ -81,9 +80,10 @@ public interface ISingleAccountPublicClientApplication extends IPublicClientAppl
      * @link AuthenticationCallback#onError(MsalException)}.
      */
     void signIn(@NonNull final Activity activity,
-                @NonNull final String loginHint,
+                @Nullable final String loginHint,
                 @NonNull final String[] scopes,
-                @NonNull final AuthenticationCallback callback);
+                @NonNull final AuthenticationCallback callback
+    );
 
     /**
      * Signs out the current the Account and Credentials (tokens).
@@ -117,7 +117,8 @@ public interface ISingleAccountPublicClientApplication extends IPublicClientAppl
      */
     void acquireTokenSilentAsync(@NonNull final String[] scopes,
                                  @NonNull final String authority,
-                                 @NonNull final SilentAuthenticationCallback callback);
+                                 @NonNull final SilentAuthenticationCallback callback
+    );
 
     /**
      * Perform acquire token silent call. If there is a valid access token in the cache, the sdk will return the access token; If
@@ -139,7 +140,7 @@ public interface ISingleAccountPublicClientApplication extends IPublicClientAppl
         /**
          * Invoked when the account is loaded.
          *
-         * @param activeAccount the signed-in account. This could be nil.
+         * @param activeAccount the signed-in account. This could be null.
          */
         void onAccountLoaded(@Nullable final IAccount activeAccount);
 
@@ -147,8 +148,8 @@ public interface ISingleAccountPublicClientApplication extends IPublicClientAppl
          * Invoked when signed-in account is changed after the application resumes, or prior to running a scheduled background operation.
          * The calling app is responsible for keeping track of this account and cleaning its states if the account changes.
          *
-         * @param priorAccount   the previous signed-in account. This could be nil.
-         * @param currentAccount the current signed-in account. This could be nil.
+         * @param priorAccount   the previous signed-in account. This could be null.
+         * @param currentAccount the current signed-in account. This could be null.
          */
         void onAccountChanged(@Nullable final IAccount priorAccount, @Nullable final IAccount currentAccount);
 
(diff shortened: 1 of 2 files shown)


msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
```diff
@@ -64,7 +63,11 @@
     ICurrentAccountResult getCurrentAccount() throws InterruptedException, MsalException;
 
     /**
-     * Allows a user to sign in to your application with one of their accounts.
+     * Allows a user to sign in to your application with one of their accounts. This method may only
```
iambmelt, line 66: @shoatman I've added this to javadoc to clarify what the behavior of this method is

msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
```diff
@@ -81,9 +84,65 @@
...
+     *                  which will have the UPN pre-populated.
+     * @param scopes    The non-null array of scopes to be consented to during sign in.
+     *                  MSAL always sends the scopes 'openid profile offline_access'.  Do not include any of these scopes in the scope parameter.
+     *                  The access token returned is for MS Graph and will allow you to query for additional information about the signed in account.
+     * @param callback  {@link AuthenticationCallback} that is used to send the result back. The success result will be
+     *                  sent back via {@link AuthenticationCallback#onSuccess(IAuthenticationResult)}.
+     *                  Failure case will be sent back via {
+     * @link AuthenticationCallback#onError(MsalException)}.
+     */
+    void signIn(@NonNull final Activity activity,
```
iambmelt, line 113: @shoatman I have introduced this overload for apps who wish to use SA-PCA, but _also_ wish to specify a `prompt` for the "first run"

msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
```diff
@@ -81,9 +84,65 @@
...
+     *                 token returned is for MS Graph and will allow you to query for
+     *                 additional information about the signed in account.
+     * @param prompt   Nullable. Indicates the type of user interaction that is required.
+     *                 If no argument is supplied the default behavior will be used.
+     * @param callback {@link AuthenticationCallback} that is used to send the result back.
+     *                 The success result will be sent back via
+     *                 {@link AuthenticationCallback#onSuccess(IAuthenticationResult)}.
+     *                 Failure case will be sent back via {@link AuthenticationCallback#onError(MsalException)}.
+     */
+    void reauthorize(@NonNull final Activity activity,
```
iambmelt, line 141: @shoatman And I introduced this _new_ method for "reauth" -- hopefully this use of this is clearer

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -211,7 +247,71 @@ public void signIn(@NonNull final Activity activity,
...
+                       @Nullable final String loginHint,
+                       @NonNull final String[] scopes,
+                       @Nullable final Prompt prompt,
+                       @NonNull final AuthenticationCallback callback) {
+        final IAccount persistedAccount = getPersistedCurrentAccount();
+
+        if (persistedAccount != null) {
+            callback.onError(
+                    new MsalClientException(
+                            MsalClientException.INVALID_PARAMETER,
```
rpdome, line 264: Actually, I think INVALID_PARAMETER could be misleading. Let's create an opposite of MsalClientException.NO_CURRENT_ACCOUNT. 

Maybe ALREADY_SIGNED_IN?

shahzaibj: +1

msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
shoatman: I'm struggling with naming here.  I think updateSignIn?  or signInAgain or reauthenticate... i'm also wondering if we want...

msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
shoatman: I think that's fine... if we do this... if the second method actually required still?

msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
iambmelt: I'll take `signInAgain` -- seems the most discoverable name since `signIn()` already exists

msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
iambmelt: @rpdome @shahzaibj Just so we're clear, you are both suggesting a breaking change. See current implementation of Single Account

Do you really want this change?

shahzaibj: We can keep it as is

msal/src/main/java/com/microsoft/identity/client/ISingleAccountPublicClientApplication.java
iambmelt: @shoatman Technically, the second method was _never_ required since `acquireToken(params)` can do all the same things.... That said....

I see this overload of `signIn` as having a different purpose from the _original `signIn()` and the newly added `reauthorize()` (soon to be renamed, `signInAgain()`
- `signIn` _without_ a prompt param will allow for SSO state pickup from the browser, if the user already has state
- `signIn` _with_ a prompt param allows the developer to tweak the prompt behavior for the _first run_, but just like the other `signIn()` method, also disalllows multiple invocations of `signIn()`
- `signInAgain()` accepts no account _at all_, and can only be used once a user has already signed in -- it's purpose is only if you must reprompt the user for creds, as a banking app might upon relaunch

iambmelt:
> I have questions on naming of the method... but other than that looks good. Thanks!

Method name changed in 3deb523159e3d00e4068aa3e1bafcf3ac397cdba

Issue #995, opened 2020-04-06
acquireTokenSilent returns null for some calls when multiple threads try to access token


fatihsokmen:
**UPDATE**: Title seems to be misleading. It should read as acquireTokenSilent throws MsalClientException with NO_CURRENT_ACCOUNT message which ends up with null token.

**Describe the bug**
I have an android client api which is called by image loading library to load many photos onto recyclerview viewholder. Bearer token needs be added to api header to request photo file. While many threads try to access token in parallel, some of them get null. 

I did some investigation in sdk source code and realised that following function [persistCurrentAccount] could be the reason why it gets null. It is quite possible that while some threads clear out _sharedPreferencesFileManager_ , and the others can read from it. So it ends up with getting null. 

I've synchronised my function call to fix this issue at the moment but I believe this should be done by sdk and access to **acquireTokenSilent** or **persistCurrentAccount** should be limited one thread at a time.

<img width="839" alt="Screenshot 2020-04-06 at 15 24 08" src="https://user-images.githubusercontent.com/2427299/78571053-ad979a80-781d-11ea-8f20-1f52675c6860.png">

**Smartphone (please complete the following information):**
 - Device: Android Emulator
 - Android Version: 28
 - Browser: Chrome
 - MSAL Version : 1.0.0 to 1.3.0

**Stacktrace**
N/A

**To Reproduce**
1 - Create a recycler view with photo on each row.
2 - Scroll recyclerview.
3 - Photo download should use access token to retrieve data.  

**Expected behavior**
If user logged in, it should always return a non-null token or an appropriate exception.

**Actual Behavior**
_acquireTokenSilent_ sometime returns null.

**Screenshots**
NA

**Additional context**
NA

iambmelt:
@rpdome I've assigned this to you; IIRC you were making some tweaks around this current account query/persistence. Let me know if you need to load balance this

iambmelt:
Thanks for your report @fatihsokmen - we'll look into this threading issue

For your specific scenario: are you accessing the same endpont/resource to load these images? If yes, you can reuse your `Bearer` token across calls up to the lifetime of the token (usually 1 hour)

fatihsokmen:
Hi @iambmelt, thanks for getting back.

Yes, it is the same resource across calls. As a quick fix now, I've synchronised access to resource which works fine now. 

[...]

PR #999, opened 2020-04-15
Make persistedAccount operations thread safe

rpdome:
- Make getPersistedCurrentAccount() and persistCurrentAccount() thread safe. (#995)
- Add error checks for function that takes in acquireToken/silent parameters (as we're moving toward making those parameter immutable)
    - This means we'll be returning errors when these API is being misused (instead of trying to 'correct' the values). Technically, this could be considered a breaking change.
- Add error messages.


msal/src/main/java/com/microsoft/identity/client/SingleAccountPublicClientApplication.java
```diff
@@ -132,6 +133,7 @@ public void onTaskCompleted(final List<ICacheRecord> result) {
                                 // We do not support switching from MULTIPLE to SINGLE.
                                 // See getAccountFromICacheRecordList() for more details.
                                 checkCurrentAccountNotifyCallback(callback, result);
+                                persistCurrentAccount(result);
```
shahzaibj, line 136: I believe we are persisting the same account that was already persisted?

rpdome: Yes, given that there might be any changes on the account object. 

An alternative would be to check everything inside the stored account and see if anything changes, which, IMHO, I think it's an overkill and might not be easily sustainable.

PR #1005, opened 2020-04-16
Added validation in getAccount with identifier API to avoid Null pointers

kreedula:
Issue reported by Teams : https://identitydivision.visualstudio.com/Engineering/_workitems/edit/982121

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
@@ -29,6 +29,7 @@ import android.os.Looper;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
+import com.microsoft.identity.client.exception.MsalArgumentException;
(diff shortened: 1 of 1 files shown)


msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
```diff
@@ -189,6 +191,13 @@ public void getAccount(@NonNull final String identifier,
     private void getAccountInternal(@NonNull final String identifier,
                                     @NonNull final GetAccountCallback callback,
                                     @NonNull final String publicApiId) {
+        try {
+            validateNonNullArg(identifier, "identifier");
+            validateNonNullArg(identifier, "callback");
```
iambmelt, line 196: Should be `callback` as first param

shahzaibj: That probably won't make a difference if the `callback` is `null`. Need to throw `IllegalStateException` instead for `callback` validation

PR #1006, opened 2020-04-16
Fix to add callback instead of identifier for null check

kreedula:
Related PR : https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1005

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
@@ -193,7 +193,7 @@ public class MultipleAccountPublicClientApplication extends PublicClientApplicat
                                     @NonNull final String publicApiId) {
         try {
             validateNonNullArg(identifier, "identifier");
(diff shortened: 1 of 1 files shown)


msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
```diff
@@ -193,7 +193,7 @@ private void getAccountInternal(@NonNull final String identifier,
                                     @NonNull final String publicApiId) {
         try {
             validateNonNullArg(identifier, "identifier");
-            validateNonNullArg(identifier, "callback");
+            validateNonNullArg(callback, "callback");
         } catch (MsalArgumentException e) {
             callback.onError(e);
```
iambmelt, line 198: If the callback is null, this catch block will throw an NPE

kreedula: so I think we can't do much here , than to throw a runtime exception like `IllegalArgumentException`

PR #1015, opened 2020-04-23
Power optimization flag changes

kreedula:
common PR : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/894

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
@@ -29,8 +29,6 @@ import android.content.pm.Signature;
 import android.net.Uri;
 import android.util.Base64;
 
-import androidx.annotation.NonNull;
-
 import com.google.gson.annotations.SerializedName;
 import com.microsoft.identity.client.configuration.AccountMode;
 import com.microsoft.identity.client.configuration.HttpConfiguration;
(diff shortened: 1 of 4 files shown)


shahzaibj:
@kreedula Builds are failing. I believe it is because https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1003 needs to merge in.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -421,6 +433,7 @@ void mergeConfiguration(PublicClientApplicationConfiguration config) {
         this.mLoggerConfiguration = config.mLoggerConfiguration == null ? this.mLoggerConfiguration : config.mLoggerConfiguration;
         this.webViewZoomControlsEnabled = config.webViewZoomControlsEnabled == null || config.webViewZoomControlsEnabled;
         this.webViewZoomEnabled = config.webViewZoomEnabled == null || config.webViewZoomEnabled;
+        this.powerOptCheckEnabled = config.powerOptCheckEnabled == null || config.powerOptCheckEnabled;
```
shahzaibj, line 436: So it seems like if they don't declare anything in config then the power opt check is enabled by default. Do we want that? Or should it be disabled by default?

kreedula: yes, given we removed `UsageStats.isAppInactive` check , I think we would be safer not to enable this by default, we can also some data from Teams to make a decision.

shahzaibj: @kreedula So you are going to update the code? It seems enabled by default at the moment

kreedula: @shahzaibj Apologies, edited the comment above

Issue #1027, opened 2020-04-30
1.5.0 Failed to resolve: com.microsoft.device.display:display-mask:0.3.0


pequots34:
**Describe the bug**
ERROR: Failed to resolve: com.microsoft.device.display:display-mask:0.3.0

dabarnard:
I also started experiencing this error today (30 April 2020). Most of the documentation recommends using

`implementation 'com.microsoft.identity.client:msal:1.+'`

for the Gradle dependancy. I had to manually set my version number down to 

`implementation 'com.microsoft.identity.client:msal:1.4.0'`

for the issue to be resolved in the interim.

fatihsokmen:
Same here. I am trying to upgrade to latest (1.5.0) but getting following error.


`Could not determine the dependencies of task ':app:compileLocalDebugJavaWithJavac'.
  Could not resolve all task dependencies for configuration ':app:localDebugCompileClasspath'.
     Could not find com.microsoft.device.display:display-mask:0.3.0.
     Required by:
         project :app > com.microsoft.identity:common:2.0.7`

iambmelt:
Please consult the [changelog for version `1.5.0`](https://github.com/AzureAD/microsoft-authentication-library-for-android/releases/tag/v1.5.0)

[A new Maven artifact stream must be added to resolve this dependency](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/build.gradle#L32-L35).

Closing this issue; please comment/reopen if this doesn't resolve your issue.

iambmelt:
@shahzaibj Can we also update the README to capture this?

shahzaibj:
@iambmelt I've opened up a PR to update README here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1031

joshfriend:
> add the following lines to your repositories section
> 'https://pkgs.dev.azure.com/MicrosoftDeviceSDK/DuoSDK-Public/_packaging/Duo-SDK-Feed%40Local/maven/v1'
> Please exclude classes under "com.microsoft.device.display" in your proguard file.

There's a bunch of reasons why I'm not a fan of this situation:

* Breaking dependency resolution of an artifact in a minor version bump
* Asking developers to add a strange package repo URL to their repo lists
* Forcing consumers to manually add proguard rules when that can be handled by the library. (If I ignore this will my app crash? I don't know)
[...]

PR #1031, opened 2020-05-01
Update readme for duo sdk instructions

shahzaibj:
Since MSAL release 1.5.0 developers need to add this for the Duo SDK. Also updating readme with instructions.

See https://github.com/AzureAD/microsoft-authentication-library-for-android/releases/tag/v1.5.0
Also see https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/1027

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,14 @@ dependencies {
 }
 ```
 
+Please also add the following lines to your repositories section in your gradle script:
+
+```gradle
+maven { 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,6 +70,14 @@ dependencies {
 }
 ```
 
+Please also add the following lines to your repositories section in your gradle script:
```
Om83, line 73: Does this also need to go in our samples and docs? cc: @hamiltonha

shahzaibj: I believe it will need to go in the sample for sure...I don't think the existing sample will work without the change.

PR #1040, opened 2020-05-11
Update consumer-rules.pro

MariusVolkhart:
GSON's TypeAdapter is an abstract class, not an interface.


iambmelt:
The proguard config that we ship has the verbatim guidance from google's repo:
https://github.com/google/gson/blob/master/examples/android-proguard-example/proguard.cfg#L18

If you believe there's an issue with this config, I'd try to submitting an issue over there -- if that merges, we can take it here

MariusVolkhart:
Done. See https://github.com/google/gson/pull/1693

iambmelt:
@MariusVolkhart Brilliant. Thanks

I'll take the liberty of updating this branch with the latest changes for a clean merge. Thank you for your contribution!

PR #1048, opened 2020-05-20
[WIP] Expose serviced from cache on Authentication Result + Update Tests to Assert for that

shahzaibj:
See BackLog Item (CORPNET REQUIRED): https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/1022804

We are exposing whether the result (Authentication Result) was serviced from cache or not. I've also updated the tests to assert for this value.

This PR also depends on common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/917 for some changes to test code as a couple of tests (silent requests) were failing after this change due to running into cache misses.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
@@ -44,10 +44,8 @@ public final class AuthenticationResult implements IAuthenticationResult {
     private final IAccount mAccount;
     private boolean mServicedFromCache;
 
-    AuthenticationResult(
-            @NonNull final List<ICacheRecord> cacheRecords,
-            final boolean isServicedFromCache
-    ) {
+    AuthenticationResult(@NonNull final List<ICacheRecord> cacheRecords,
+                         final boolean isServicedFromCache) {
         final ICacheRecord mostRecentlyAuthorized = cacheRecords.get(0);
         mAccessToken = mostRecentlyAuthorized.getAccessToken();
         mTenantId = mostRecentlyAuthorized.getAccount().getRealm();
--- a/msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
@@ -113,7 +113,7 @@ public final class AuthenticationResult implements IAuthenticationResult {
     }
 
     @Override
-    public boolean isServicedFromCache() {
+    public boolean wasServicedFromCache() {
         return mServicedFromCache;
     }
 }
--- a/msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
@@ -42,12 +42,17 @@ public final class AuthenticationResult implements IAuthenticationResult {
     private final String mTenantId;
     private final AccessTokenRecord mAccessToken;
     private final IAccount mAccount;
+    private boolean mServicedFromCache;
 
-    AuthenticationResult(@NonNull final List<ICacheRecord> cacheRecords) {
+    AuthenticationResult(
+            @NonNull final List<ICacheRecord> cacheRecords,
+            final boolean isServicedFromCache
+    ) {
         final ICacheRecord mostRecentlyAuthorized = cacheRecords.get(0);
         mAccessToken = mostRecentlyAuthorized.getAccessToken();
         mTenantId = mostRecentlyAuthorized.getAccount().getRealm();
         mAccount = AccountAdapter.adapt(cacheRecords).get(0);
+        mServicedFromCache = isServicedFromCache;
     }
 
     @Override
@@ -107,5 +112,10 @@ public final class AuthenticationResult implements IAuthenticationResult {
     @NonNull
     public String[] getScope() {
         return mAccessToken.getTarget().split("\\s");
+    }
+
+    @Override
+    public boolean isServicedFromCache() {
+        return mServicedFromCache;
     }
 }
--- a/msal/src/test/java/com/microsoft/identity/client/e2e/tests/AcquireTokenAbstractTest.java
+++ b/msal/src/test/java/com/microsoft/identity/client/e2e/tests/AcquireTokenAbstractTest.java
@@ -82,6 +82,12 @@ public abstract class AcquireTokenAbstractTest extends PublicClientApplicationAb
         flushScheduler();
     }
 
+    // This method is used to perform a silent acquire token call with the provided scopes.
+    // The expectingResultFromCache parameters is used to determine whether we expect the result
+    // of this silent request to be served from cache or not. This value will then be compared
+    // against the actual value of isServicedFromCache in Authentication Result to determine
+    // if the test behaved as expected. If the value does not match with what we expect, then the
+    // test will fail.
     public void performSilentAcquireTokenCall(final String[] scopes,
                                               final boolean expectingResultFromCache) {
         final AcquireTokenSilentParameters silentParameters = new AcquireTokenSilentParameters.Builder()
@@ -96,6 +102,12 @@ public abstract class AcquireTokenAbstractTest extends PublicClientApplicationAb
         flushScheduler();
     }
 
+    // This method is used to perform a silent acquire token call with the provided account.
+    // The expectingResultFromCache parameters is used to determine whether we expect the result
+    // of this silent request to be served from cache or not. This value will then be compared
+    // against the actual value of isServicedFromCache in Authentication Result to determine
+    // if the test behaved as expected. If the value does not match with what we expect, then the
+    // test will fail.
     public void performSilentAcquireTokenCall(final IAccount account,
                                               final boolean expectingResultFromCache) {
         final AcquireTokenSilentParameters silentParameters = new AcquireTokenSilentParameters.Builder()
@@ -110,6 +122,12 @@ public abstract class AcquireTokenAbstractTest extends PublicClientApplicationAb
         flushScheduler();
     }
 
+    // This method is used to perform a silent acquire token call with the provided account and
+    // authority. The expectingResultFromCache parameters is used to determine whether we expect the
+    // result of this silent request to be served from cache or not. This value will then be
+    // compared against the actual value of isServicedFromCache in Authentication Result to
+    // determine if the test behaved as expected. If the value does not match with what we expect,
+    // then the test will fail.
     public void performSilentAcquireTokenCall(final IAccount account,
                                               final String authority,
                                               final boolean expectingResultFromCache) {
--- a/msal/src/test/java/com/microsoft/identity/client/e2e/tests/AcquireTokenAbstractTest.java
+++ b/msal/src/test/java/com/microsoft/identity/client/e2e/tests/AcquireTokenAbstractTest.java
@@ -82,39 +82,43 @@ public abstract class AcquireTokenAbstractTest extends PublicClientApplicationAb
         flushScheduler();
     }
 
-    public void performSilentAcquireTokenCall(final String[] scopes) {
+    public void performSilentAcquireTokenCall(final String[] scopes,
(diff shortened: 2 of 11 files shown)


msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
```diff
@@ -42,12 +42,17 @@
     private final String mTenantId;
     private final AccessTokenRecord mAccessToken;
     private final IAccount mAccount;
+    private boolean mServicedFromCache;
 
-    AuthenticationResult(@NonNull final List<ICacheRecord> cacheRecords) {
+    AuthenticationResult(
+            @NonNull final List<ICacheRecord> cacheRecords,
+            final boolean isServicedFromCache
+    ) {
```
iambmelt, line 50: As a formatting nit: could bring these braces up to end of the previous line

msal/src/test/java/com/microsoft/identity/client/e2e/tests/AcquireTokenAbstractTest.java
```diff
@@ -82,39 +82,43 @@ public void performInteractiveAcquireTokenCall(final String username, final Stri
         flushScheduler();
     }
 
-    public void performSilentAcquireTokenCall(final String[] scopes) {
+    public void performSilentAcquireTokenCall(final String[] scopes,
+                                              final boolean expectingResultFromCache) {
```
iambmelt, line 86: From the name, it's not immediately obvious what `expectingResultFromCache` is, or when it's `true`/`false`

Can we add some javadoc?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/AcquireTokenAbstractTest.java
```diff
@@ -82,39 +82,43 @@ public void performInteractiveAcquireTokenCall(final String username, final Stri
...
+                .withCallback(successfulSilentCallback(expectingResultFromCache))
                 .build();
 
         mApplication.acquireTokenSilentAsync(silentParameters);
         flushScheduler();
     }
 
-    public void performSilentAcquireTokenCall(final IAccount account) {
+    public void performSilentAcquireTokenCall(final IAccount account,
+                                              final boolean expectingResultFromCache) {
```
iambmelt, line 100: Same here

msal/src/test/java/com/microsoft/identity/client/e2e/tests/AcquireTokenAbstractTest.java
```diff
@@ -82,39 +82,43 @@ public void performInteractiveAcquireTokenCall(final String username, final Stri
...
                 .build();
 
         mApplication.acquireTokenSilentAsync(silentParameters);
         flushScheduler();
     }
 
-    public void performSilentAcquireTokenCall(final IAccount account, final String authority) {
+    public void performSilentAcquireTokenCall(final IAccount account,
+                                              final String authority,
+                                              final boolean expectingResultFromCache) {
```
iambmelt, line 115: same here

msal/src/test/java/com/microsoft/identity/client/e2e/tests/AcquireTokenAbstractTest.java
iambmelt: Or just comment/explainer for this flag. Don't necessarily need to javadoc tests

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
```diff
@@ -42,12 +42,17 @@
     private final String mTenantId;
     private final AccessTokenRecord mAccessToken;
     private final IAccount mAccount;
+    private boolean mServicedFromCache;
 
-    AuthenticationResult(@NonNull final List<ICacheRecord> cacheRecords) {
+    AuthenticationResult(
+            @NonNull final List<ICacheRecord> cacheRecords,
+            final boolean isServicedFromCache
```
AdamBJohnsonx, line 49: This is just about whether this is served from the local MSAL cache, I think.  Are we going to be similarly curious if this was served by an auth broker?

shahzaibj: Yes, this is the original work item: https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/875729

I split it into two legs: served from cache, served from broker.

This the work item I created for implementing the served from broker: https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/1022834

Going to be working on that next!

msal/src/test/java/com/microsoft/identity/client/e2e/tests/AcquireTokenAbstractTest.java
iambmelt: (Didn't check file name before commenting; realize now that this is a test file)

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
```diff
@@ -108,4 +113,9 @@ public IAccount getAccount() {
     public String[] getScope() {
         return mAccessToken.getTarget().split("\\s");
     }
+
+    @Override
+    public boolean isServicedFromCache() {
```
iambmelt, line 118: Not about `is` vs `was` here....

Grammatically, I think `was` works better

```java
if (result.wasServicedFromCache()) {
    // ...
}
```

There may be other method names we could choose from....

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
AdamBJohnsonx: OK.  There's not a lot of detail in these work items in terms of the actual desire here.  This makes me wonder: should this be a boolean or should it be some kind of object that could contain more information.  Because the next questions would be "which broker", and "how long has it been in this cache", "was it cached by the broker involved", etc.  I think we (should) get the age of the decision from the auth token itself, I think.

shahzaibj: I think we should definitely do "which broker". I think if the broker was being used, then it is implied that when the result was serviced from cache then it was serviced from the cache of the broker. For specifying how long it's been there would be a bit of an overkill IMO but open to suggestions on that.

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1048/commits/0f34d4cb4448efedf0ee4b691126cad86662a7a9

msal/src/test/java/com/microsoft/identity/client/e2e/tests/AcquireTokenAbstractTest.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1048/commits/f09a7afc15134b05fd6c23615308bfa4ea53dbdf

msal/src/test/java/com/microsoft/identity/client/e2e/tests/AcquireTokenAbstractTest.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1048/commits/f09a7afc15134b05fd6c23615308bfa4ea53dbdf

msal/src/test/java/com/microsoft/identity/client/e2e/tests/AcquireTokenAbstractTest.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1048/commits/f09a7afc15134b05fd6c23615308bfa4ea53dbdf

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
shahzaibj: Done here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1048/commits/70be3c0419d2ad18ad565fdea69411540c038908

We have the same thing on the internal `ILocalAuthenticationResult` here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/result/ILocalAuthenticationResult.java#L129 I'll change that too, albeit in a separate PR in common

shahzaibj:
@iambmelt Thanks for the review! All of your comments have been addressed!

msal/src/main/java/com/microsoft/identity/client/AuthenticationResult.java
AdamBJohnsonx: Right, I keep forgetting that the MSAL cache isn't a first tier before the broker process, though some form of sharing might be desirable in the longer term.  Generally, though, knowing the age of objects in a cache is useful for monitoring its effectiveness (and finding bugs in age-related code), though I think we don't do capacity-based cache eviction.

The discussion of record age aside, if you just had a source object with a few different potential kinds of value, maybe

local
broker(identity-of-broker)
authority(identity-of-authority?)

that would let you put all of this information into a single method and save us trouble in the future if we needed to add more of them, unless there's some reason to strongly prefer a primitive type in the return.

shahzaibj: @AdamBJohnsonx We could create object called as `ResultMetadata` and put these values like whether the result was served from cache or broker, which broker was used etc. inside that object and then expose that to the developers. 

For now we would have 3 fields: `wasServicedFromCache`, `wasServicedFromBroker`, `brokerUsed`, but as you said, we may want to add more fields here going forward so I think having a metadata object makes sense. Since this public API, I am holding off on merging this, let's think more about and let me know if you have any suggestions.

AdamBJohnsonx: This is actually a separate question - this API is about more than just Android, so if we're already committed to a common API where these are booleans and strings then my entire discussion is moot. ;-)

shahzaibj: This API is specifically for Android. AFAIK other platforms have not implemented this. This was something that was asked by the CSS support team to enable in Android. @iambmelt would have more context on that.

I think if other platforms choose to implement such a meta data, then they need to use the return types applicable to that platform/language.

AdamBJohnsonx: Then in this case, having a containing type makes the most sense for us.  It does for other MSALs as well, honestly, since it's a stumbling block for anybody working on or using an auth client, and hopefully the response shape is something that we can align on with other clients.  Developers on android are occasionally the same people working on iOS, etc. and it's frustrating when I can't use the library for one system in the same way I can the second.

shahzaibj: @AdamBJohnsonx If we are going to doing that then we should probably do that across all public apis in Android and iOS. That is not the case right now. For example, the errors returned by iOS and Android via their public api are completely different..they are even completely different types. Android exposes errors as exception objects and error codes as Strings whereas iOS does so by using enums: https://github.com/AzureAD/microsoft-authentication-library-for-objc/blob/master/MSAL/src/public/MSALError.h#L128

TBH, I don't even know what types we may be able to use that are the same across platforms. IMHO, when creating public API and selecting types, we should be picking types that offer the best experience on a given platform.

AdamBJohnsonx: If we haven't done it it isn't worth limiting ourselves with it.  I don't think we're disagreeing, mind, I was more making a note that this feature is surely going to be desired elsewhere.

shahzaibj:
Closing for now as going to be writing an API Review doc for this.

PR #1051, opened 2020-05-20
Use an existing utility method to check for email claim, if preferred user name is absent when adapting to AccountRecord

kreedula:
- Passthru accounts doesn't seem to have `preferred_username` claim, use `email` instead.
- Bug reported by teams  where broker can't return the token due to `name` mismatch : https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1023276

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/Account.java
+++ b/msal/src/main/java/com/microsoft/identity/client/Account.java
@@ -26,6 +26,7 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import com.microsoft.identity.client.exception.MsalClientException;
+import com.microsoft.identity.common.internal.cache.SchemaUtil;
 import com.microsoft.identity.common.internal.logging.Logger;
 import com.microsoft.identity.common.internal.providers.microsoft.MicrosoftIdToken;
 import com.microsoft.identity.common.internal.providers.microsoft.azureactivedirectory.AzureActiveDirectoryIdToken;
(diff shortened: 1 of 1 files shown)


shahzaibj:
@kreedula  Not clear what that ADO query is supposed to point to. May be provide link to a specific work item?

shahzaibj:
@kreedula What about `upn` claim? Is that not there too? Looks the existing logic fallbacks to upn if preferred username is not there

kreedula:
> @kreedula What about `upn` claim? Is that not there too? Looks the existing logic fallbacks to upn if preferred username is not there

yes v2 returns `preferred_username` , v1 returns `upn` , in passthru case we don't get any, so we need to rely on `email` :(

iambmelt:
@kreedula Did we get any input from EVO w.r.t to using `email` as fallback? Will that fallback work in all cases? Any special passthru account detection we need to do apart from this?

shahzaibj:
@kreedula How has passthru worked in teams all this time? My understanding is that with MSA accounts, teams always does an initial sign in of the user into the Microsoft Services Tenant using MSA Passthrough, seems like that didn't break yet. Is this passthrough user different from that?

kreedula:
> @kreedula Did we get any input from EVO w.r.t to using `email` as fallback? Will that fallback work in all cases? Any special passthru account detection we need to do apart from this?

Not that I am aware Brian, I remember you were testing some passthru behavior and did some documentation if I am not wrong. Also we have this logic in the library for all scenarios not only passthru accounts where if we couldn't get `preferred_username` we fallback to `email`. We missed to do it here.

kreedula:
> @kreedula How has passthru worked in teams all this time? My understanding is that with MSA accounts, teams always does an initial sign in of the user into the Microsoft Services Tenant using MSA Passthrough, seems like that didn't break yet. Is this passthrough user different from that?

With local MSAL it works fine, this an upgrade scenerio with ADAL to MSAL, where MSA guest account is stored in Broker and when we go to Broker, the name doesn't match the AccountManager account name due to the missing claim. The VSTS bug has the exact repro steps.

PR #1053, opened 2020-05-27
Changes related to add Content Provider Strategy to MSAL

kreedula:
Common PR : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/918
Broker PR : https://github.com/AzureAD/ad-accounts-for-android/pull/1298
Change Doc : https://identitydivision.visualstudio.com/DefaultCollection/DevEx/_git/AuthLibrariesApiReview/pullrequest/1554


msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerContentProviderStrategy.java
```diff
@@ -0,0 +1,487 @@
...
+            throws BaseException {
+        final String methodName = "acquireTokenSilentWithContentProvider";
+
+        return performContentProviderOperation(
+                parameters,
+                new ContentProviderOperation<SilentTokenCommandParameters, AcquireTokenResult>() {
+                    @Nullable
+                    @Override
+                    public String getRequestString(SilentTokenCommandParameters parameters) {
+                        final BrokerRequest brokerRequest = mRequestAdapter.
```
shahzaibj, line 251: since we have specific broker request classes now for other operations such as hello, remove account etc. I would suggest renaming `BrokerRequest.java` to something like `BrokerSilentTokenRequest.java` as I believe this will now be only used for silent token requests?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerContentProviderStrategy.java
```diff
@@ -0,0 +1,487 @@
...
+                          @Nullable final String negotiatedBrokerProtocolVersion) throws BaseException {
+        final String methodName = "getDeviceModeWithContentProvider";
+        return performContentProviderOperation(
+                parameters,
+                new ContentProviderOperation<CommandParameters, Boolean>() {
+
+            @Nullable
+            @Override
+            public String getRequestString(CommandParameters parameters) {
+                return null; // broker returns a boolean to indicate the device mode, no request string needed.
```
shahzaibj, line 374: Having `null` seems confusing and seems like nothing is being sent to the broker. I believe this is NULL because the only thing we are sending is basically telling the broker that we need to get the device mode? and due to the nature of the get device mode command, it doesn't need any additional parameters to be sent along unlike other operations such as remove account?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerContentProviderStrategy.java
kreedula: `BrokerRequest` will be used to pass all the request parameters for both silent and interactive requests. Given we need to support back compat with old MSAL and old brokers. I am not inclined  to split and create two new classes for `silent` and `interactive` , so leaving as - is

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerContentProviderStrategy.java
kreedula: Yes, `requestString` is different from `requestOperation` in itself. There could be operations to broker which doesn't need any additional request data to be passed , like `getDeviceMode` here. But the interface is common for all the operations , hence I tagged this with `Nullable`

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerContentProviderStrategy.java
```diff
@@ -0,0 +1,487 @@
...
+            }
+
+            @Override
+            public String getUriPath() {
+                return ACQUIRE_TOKEN_INTERACTIVE_PATH;
+            }
+
+            @Override
+            public Intent getResultFromBundle(Bundle resultBundle) throws BaseException {
+                final Bundle requestBundle = mRequestAdapter.getRequestBundleForAcquireTokenInteractive(
```
rpdome, line 218: call BrokerBaseStrategy.completeInteractiveRequestIntent()?

rpdome: (and we might want to move this function to MsalBrokerResultAdapter too)

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerContentProviderStrategy.java
```diff
@@ -0,0 +1,487 @@
...
+
+    private static final String TAG = BrokerContentProviderStrategy.class.getName();
+
+    public interface ContentProviderOperation<T extends CommandParameters, U> {
+
+        /**
+         * Constructs and returns a data string for the request to the broker from the CommandParameters.
+         */
+        @Nullable
+        String getRequestString(T parameters);
```
rpdome, line 77: nit: if possible to serialize/marshall the request bundle (as suggested in ad-accounts), I suggest renaming this to something like 'get Marshalled/Serialized RequestBundle'.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerContentProviderStrategy.java
```diff
@@ -0,0 +1,487 @@
...
+            @Nullable
+            @Override
+            public String getRequestString(RemoveAccountCommandParameters parameters) {
+                return mRequestAdapter.getRequestStringForRemoveAccount(parameters);
+            }
+
+            @Nullable
+            @Override
+            public String[] getOptions() {
+                return new String[]{negotiatedBrokerProtocolVersion};
```
rpdome, line 339: not particularly related to this PR, but if we manage to be able to reuse the request bundle here, can you move negotiatedBrokerProtocolVersion into MsalBrokerRequestAdapter as well? (there will be some change to clean up redundant codes in the other two strategy).

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerContentProviderStrategy.java
kreedula: moved it to `MsalBrokerResultAdapter`, we don't need to call `completeInteractiveRequestIntent` , as we construct the intent here using the bundle

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerContentProviderStrategy.java
kreedula: resolved via https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1053/commits/9020493baf4c990129af7f4d303efd5a03c85c22

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerContentProviderStrategy.java
kreedula: renamed to `getRequestBundle` as appropriately

kreedula:
@rpdome Addressed your comments and added `marshalling`

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -246,6 +246,7 @@ public AcquireTokenResult acquireToken(InteractiveTokenCommandParameters paramet
     // The order matters.
     private List<BrokerBaseStrategy> getStrategies() {
         final List<BrokerBaseStrategy> strategies = new ArrayList<>();
+        strategies.add(new BrokerContentProviderStrategy());
```
rpdome, line 249: Should we try authService first for a couple of release? Once it's tried and tested (i.e. with Teams), then we can switch it to the first one in the list.

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerContentProviderStrategy.java
```diff
@@ -0,0 +1,446 @@
...
+                    }
+
+                    @Override
+                    public String getUriPath() {
+                        return ACQUIRE_TOKEN_INTERACTIVE_PATH;
+                    }
+
+                    @Override
+                    public Intent getResultFromBundle(Bundle resultBundle) throws BaseException {
+                        return mRequestAdapter.getRequestIntentForAcquireTokenInteractive(
```
rpdome, line 207: Is this doing the same thing as completeInteractiveRequestIntent()?

(Other strategies are using that, let's refactor in such a way that they're doing the same thing - either using that function or keep using this).

kreedula: Nope, it's not exactly doing the same thing. We can get an `Intent` as a result using `ContentProvider` . I mean technically we don't need an `Intent` here for other strategies as well, all we need is a `bundle` or `data` to launch the `AccountChooserActivity`. So I consolidated constructing the `bundle` logic in `broker` , but it returns an `intent` for `auth` and `accountmanager` strategies and `bundle` here , so we need to construct the `intent` and launch the activity

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
kreedula: Teams needs to test this on prod brokers right, if we add `AuthService` first, they will still run into this issue and `ContentProvider` fallback might not work. `ContentProvider` does the data transfer using android shared memory but it still uses `Binder` to communicate, and if the process is dead due to the exception before, this can' communicate as well.

Also, Since we always fallback to other strategies, it should work like before

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerAccountManagerStrategy.java
```diff
@@ -212,11 +212,9 @@ public Bundle getRequestBundle(SilentTokenCommandParameters parameters) {
                                 parameters,
                                 negotiatedBrokerProtocolVersion
                         );
-                        requestBundle.putString(AuthenticationConstants.Broker.BROKER_ACCOUNT_MANAGER_OPERATION_KEY,
-                                AuthenticationConstants.BrokerAccountManagerOperation.ACQUIRE_TOKEN_SILENT);
                         requestBundle.putString(
-                                AuthenticationConstants.Broker.NEGOTIATED_BP_VERSION_KEY,
```
shahzaibj, line 218: Are we no longer sending this?

kreedula: We do, but i moved the logic to `MsalBrokerRequestAdapter` where the `bundle` is constructed to avoid duplicating in all strategies

msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java
```diff
@@ -254,14 +256,26 @@ public AcquireTokenResult acquireToken(InteractiveTokenCommandParameters paramet
...
+            sb.append("AuthServiceStrategy, ");
             strategies.add(new BrokerAuthServiceStrategy());
         }
+
         if (AccountManagerUtil.canUseAccountManagerOperation(mApplicationContext)){
+            sb.append("AccountManagerStrategy.");
             strategies.add(new BrokerAccountManagerStrategy());
         }
 
+        Logger.info(TAG, sb.toString());
```
shahzaibj, line 277: 👍

PR #1056, opened 2020-05-29
Make brokerEligible generic to all strategies. Make BrokerMsalControler throws MsalClientException.

rpdome:
brokerEligible will no longer try to check if MSAL can bind. BrokerMsalController will do the work.
@kreedula this would probably affect your change.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.java
@@ -69,7 +69,7 @@ public class MSALControllerFactory {
                                                       @NonNull final PublicClientApplicationConfiguration applicationConfiguration)
             throws MsalClientException {
         if (brokerEligible(applicationContext, authority, applicationConfiguration)) {
-            return new BrokerMsalController();
+            return new BrokerMsalController(applicationContext);
         } else {
             return new LocalMSALController();
         }
@@ -98,7 +98,7 @@ public class MSALControllerFactory {
         List<BaseController> controllers = new ArrayList<>();
         controllers.add(new LocalMSALController());
         if (brokerEligible(applicationContext, authority, applicationConfiguration)) {
-            controllers.add(new BrokerMsalController());
+            controllers.add(new BrokerMsalController(applicationContext));
         }
 
         return controllers;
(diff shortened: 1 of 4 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.java
```diff
@@ -121,49 +121,36 @@ public static boolean brokerEligible(@NonNull final Context applicationContext,
...
         // Check if broker installed
         if (!brokerInstalled(applicationContext)) {
-            Logger.verbose(TAG + methodName, logBrokerEligibleFalse);
-            Logger.verbose(TAG + methodName, "Broker application is not installed.");
+            Logger.verbose(TAG + methodName, logBrokerEligibleFalse +
+                    "Broker application is not installed.");
             return false;
         }
 
-        // Check if MicrosoftAuthService supported or AccountManager permission granted
```
kreedula, line 149: With this change, if in any case, `AuthService`, `AccountManager` and now `ContentProvider` , we will not try local right ?

rpdome: yup it'll always return true (= try broker). at this stage we don't care if any of them are available (or not). We know that broker app is installed and the app is configured to use broker. that's it.

error checking will be done when we actually are getting strategies.

kreedula: Is this the desired behavior though, with the previous check, if we can't connect to either of the strategies, we can atleast try local authentication. But with this change, we don't have the fallback of local authentication right?

rpdome: The previous behavior would throw an exception, not return false. (and IMO it doesn't make sense to me to try local auth if we fail to communicate to broker).

rpdome: (i.e. it should fail so that the user turns on power optimization, not unpredictably switching between local and broker)

rpdome: Actually, for silent call, throwing an exception here means that it won't be using local cache...

In this case this is a behavior change. The question is what should be the desired one? Given that local is always called first, maybe the code was bugged all along?

kreedula: We use `MSALControllerFactory.getAllControllers` for all silent/get/remove accounts right. It will add `LocalMsalController` first and then checks eligilibilty for `BrokerMsalController` and adds it. So, no behavior change here.

For interactive requests we use `MSALControllerFactory.getDefaultController` right, where if broker is eligible, we use `BrokerMsalController` and else `LocalMsalController` , so that would be a behavior change as validation checks for these strategies are removed, we pick `BrokerMSalController` and then if validation fails later , we wont do local authentication.

But again, are these checks to validate `AuthService` and `AccountManager` can ever be false ?

rpdome: with the old logic, if any of the checks failed it would throw exception right away. Given that this function is invoked in getAllController(), it means that it won't proceed with the LocalMsalController that was previously added to the list.

isMicrosoftAuthServiceSupported() checks if the bind service exists in broker, so the only scenario it would fail would be when the user uses BrokerV1. 

account manager looks for API version, permission, etc, but given that the is an or clause, this should never fail if BrokerV2 is installed.

kreedula: Wow I was under the impression, it returns a `boolean`, so it's busted all along 😆

PR #1064, opened 2020-06-12
Changes to allow applications to set OIDC scopes

kreedula:


msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -363,6 +368,14 @@ public void setPowerOptCheckEnabled(Boolean powerOptCheckEnabled) {
         this.powerOptCheckEnabled = powerOptCheckEnabled;
     }
 
+    /**
+     * Gets the List of OIDC scopes for the PublicClientApplication.
+     * @return List of OIDC scopes
```
shahzaibj, line 373: nit: run formatter

msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java
```diff
@@ -283,7 +285,7 @@ public AcquireTokenResult acquireTokenSilent(
 
         final List<ICacheRecord> cacheRecords = tokenCache.loadWithAggregatedAccountData(
                 parametersWithScopes.getClientId(),
-                TextUtils.join(" ", parametersWithScopes.getScopes()),
+                TextUtils.join(" ", requestScopes),// use only the requested scopes when retrieving access token.
```
shahzaibj, line 288: Not clear to me why this is being done? Looks we won't get a new ID Token now?

shahzaibj: Okay - this is not token request. Just cache lookup

kreedula: This is when we are retrieving the `access token` - we match if the scopes exists in the stored access token, so we should not include the `default` or the `oidc` scopes.

We don't match `idtoken` on scopes when retrieving

shahzaibj: But do we need the default scopes for lookup? Looks like we used to do so

shahzaibj: Okay we omit default scopes at the lower cache lookup layer

shahzaibj: We may want to keep it that way, your call

kreedula: We always omit `default scopes` while [comparing here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/cache/AbstractAccountCredentialCache.java#L245). In fact , we never save default scopes for accesstoken too , see [here ](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common/src/main/java/com/microsoft/identity/common/internal/cache/MicrosoftStsAccountCredentialAdapter.java#L142). So it's essentially a no-op, so we better only compare with requested scopes as only they will be present in the AT.

shahzaibj:
Can we add an E2E (network) test with a config file that requests specific oidc scopes and we can assert for getting those back in ID Token? and follow that up with a silent request to make sure everything works.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -150,6 +152,9 @@
     @SerializedName(POWER_OPT_CHECK_FOR_NETWORK_REQUEST_ENABLED)
     private Boolean powerOptCheckEnabled;
 
+    @SerializedName(OIDC_SCOPES)
+    private List<String> oidcScopes;
```
AdamBJohnsonx, line 156: Elsewhere, this is a set.  Should we make it a set?

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -138,7 +138,11 @@ public static InteractiveTokenCommandParameters createInteractiveTokenCommandPar
                 .isWebViewZoomControlsEnabled(configuration.isWebViewZoomControlsEnabled())
                 .isWebViewZoomEnabled(configuration.isWebViewZoomEnabled())
                 .powerOptCheckEnabled(configuration.isPowerOptCheckForEnabled())
-                .build();
+                .oidcScopes(
+                        configuration.getOidcScopes() != null
+                                ? new HashSet<>(configuration.getOidcScopes())
+                                : null
+                ).build();
```
AdamBJohnsonx, line 145: nit: if we can't just make it a set, could we DRY this out a little?

iambmelt:
Can we add a description to PR covering what's being added, how/why?

kreedula:
Closing this as this is no longer required.

PR #1065, opened 2020-06-12
Make the broker uri check case-insensitive

rpdome:
Found this during the debugging issue. The customer was registering part of their package name in the portal with capital letters (and MSAL would just perform local auth (without any obvious signal that would make the developer aware)... because the backcompat redirect url support logic [here](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/44b9414f3be233053f441435440db7fbcee312b0/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java#L554))

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
@@ -481,8 +481,8 @@ public class PublicClientApplicationConfiguration {
 
     private boolean isBrokerRedirectUri() {
         final String BROKER_REDIRECT_URI_REGEX = "msauth://" + mAppContext.getPackageName() + "/.*";
-        final Pattern pairRegex = Pattern.compile(BROKER_REDIRECT_URI_REGEX);
-        final Matcher matcher = pairRegex.matcher(mRedirectUri);
+        final Pattern pairRegex = Pattern.compile(BROKER_REDIRECT_URI_REGEX.toLowerCase());
+        final Matcher matcher = pairRegex.matcher(mRedirectUri.toLowerCase());
         return matcher.matches();
     }
 


shahzaibj:
> Found this during the debugging issue. The customer was registering part of their package name in the portal with capital letters (and MSAL would silently just go to broker... because the backcompat redirect url support logic [here](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/44b9414f3be233053f441435440db7fbcee312b0/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java#L554))

MSAL would silently go to broker? Not clear what that means? I believe if the case was mismatching then we weren't going to broker, right?

rpdome:
> > Found this during the debugging issue. The customer was registering part of their package name in the portal with capital letters (and MSAL would silently just go to broker... because the backcompat redirect url support logic [here](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/44b9414f3be233053f441435440db7fbcee312b0/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java#L554))
> 
> MSAL would silently go to broker? Not clear what that means? I believe if the case was mismatching then we weren't going to broker, right?

sorry. typo. it would silently NOT going to broker. Will edit that.

shahzaibj:
> > > Found this during the debugging issue. The customer was registering part of their package name in the portal with capital letters (and MSAL would silently just go to broker... because the backcompat redirect url support logic [here](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/44b9414f3be233053f441435440db7fbcee312b0/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java#L554))
> > 
> > 
> > MSAL would silently go to broker? Not clear what that means? I believe if the case was mismatching then we weren't going to broker, right?
> 
> sorry. typo. it would silently NOT going to broker. Will edit that.

Cool. I'm actually still not clear on the "silently" part..even interactive was going through local auth, right?

rpdome:
> > > > Found this during the debugging issue. The customer was registering part of their package name in the portal with capital letters (and MSAL would silently just go to broker... because the backcompat redirect url support logic [here](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/44b9414f3be233053f441435440db7fbcee312b0/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java#L554))
> > > 
> > > 
> > > MSAL would silently go to broker? Not clear what that means? I believe if the case was mismatching then we weren't going to broker, right?
> > 
> > 
> > sorry. typo. it would silently NOT going to broker. Will edit that.
> 
> Cool. I'm actually still not clear on the "silently" part..even interactive was going through local auth, right? 

It is interactive. By "silently", I  meant developers might not even be aware that it's not using Broker, especially if they use webview.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -481,8 +481,8 @@ private void nullConfigurationCheck(String configKey, String configValue) {
 
     private boolean isBrokerRedirectUri() {
         final String BROKER_REDIRECT_URI_REGEX = "msauth://" + mAppContext.getPackageName() + "/.*";
-        final Pattern pairRegex = Pattern.compile(BROKER_REDIRECT_URI_REGEX);
-        final Matcher matcher = pairRegex.matcher(mRedirectUri);
+        final Pattern pairRegex = Pattern.compile(BROKER_REDIRECT_URI_REGEX.toLowerCase());
+        final Matcher matcher = pairRegex.matcher(mRedirectUri.toLowerCase());
```
AdamBJohnsonx, line 485: Mm.  I'm kinda suspicious here, but if you want to do this you can make your pattern case insensitive with Pattern.compile(BROKER_REDIRECT_URI_REGEX, Pattern.CASE_INSENSITIVE) and don't do all the string conversion.  Can we store this compiled pattern somewhere?

shahzaibj:
@rpdome Actually are redirect uri truly case insensitive? Does the server really not care?

rpdome:
> @rpdome Actually are redirect uri truly case insensitive? Does the server really not care?

We were able to get it to work (token acquired) by making the redirect URL in the json file lower case. Given that the one registered in the portal is still upper case, I'd say it's case insensitive.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
rpdome: Let me try that option. Should work too.

AdamBJohnsonx: And we should surely add a test for this behavior. If this is going to be something like 'com.example.application', should we escape the . characters, or are we ok with com.exampletapplication matching?

shahzaibj:
> > @rpdome Actually are redirect uri truly case insensitive? Does the server really not care?
> 
> We were able to get it to work (token acquired) by making the redirect URL in the json file lower case. Given that the one registered in the portal is still upper case, I'd say it's case insensitive.

Looking around at some documentation, they may actually be case-sensitive: 
- https://docs.microsoft.com/en-us/azure/active-directory/develop/reply-url
- https://github.com/MicrosoftDocs/azure-docs/issues/40422

IMO, we should may be engage with the server team (and may be the portal team) to confirm this.

shahzaibj:
Also the behavior co

> Left a suggestion to confirm about case sensitivity with server teams before proceeding with this change. Feel free to add the rest of the team to the thread.

Also the behavior may be different in AAD vs B2C, so may be worth engaging with both of them.

rpdome:
@shahzaibj In that case, I'm more inclined to 'look for legacy redirect url' instead and let the server handle case insensitivity. What do you think?

shahzaibj:
@rpdome Why do we need to specifically look for it? We would fallback to legacy anyway right? Like we assume that the redirect uri is legacy and does not need broker support.

rpdome:
> @rpdome Why do we need to specifically look for it? We would fallback to legacy anyway right? Like we assume that the redirect uri is legacy and does not need broker support.

The goal here is to make sure that the legacy Redirect URI doesn't make its way to the broker.

rpdome:
Closing... I sent out an email for longer term proposal. Thanks for the review @shahzaibj  @AdamBJohnsonx

PR #1070, opened 2020-06-15
Add single account overload tests

shahzaibj:
Single Account PCA provides unique behaviour across each of its methods and overloads such as `signIn`, `signInAgain`, `acquireToken`, `signOut` etc. In this PR, I've added tests for those overloads.

Related PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/936


msal/src/test/java/com/microsoft/identity/client/e2e/tests/mocked/SingleAccountOverloadsMockedTest.java
```diff
@@ -0,0 +1,419 @@
...
+                Assert.assertTrue("Successfully signed out", true);
+            }
+
+            @Override
+            public void onError(@NonNull MsalException exception) {
+                fail(exception.getMessage());
+            }
+        });
+
+        RoboTestUtils.flushScheduler();
```
iambmelt, line 138: I see this at the conclusion of a few different tests; would this block make sense to put into an `@After` block?

(Also, I can't remember _why_ we need to flush the scheduler here; is this to reset thread state after test?)

shahzaibj: `flushScheduler` is being used in Robolectric tests to wait for the current runnable to finish executing and advance to the next runnable. It is similar to the `CountDownLatch` mechanism. 

This is not being used on per-test basis but rather on every request that is expected to go through the command dispatcher i.e any request that is going to be run on a different thread. So we use flushScheduler to advance in that case otherwise Robolectric tests will just be hung forever.

So essentially it can't go on the `@After` block as it would have to be done on the level of each token request or other API execution.

iambmelt: Thanks for clarifying, @shahzaibj -- resolving this

PR #1088, opened 2020-07-02
Update common sumbodule.

AdamBJohnsonx:
* Document the process.

### DIFF
--- a/docs/submodule-update.md
+++ b/docs/submodule-update.md
@@ -1,0 +1,11 @@
+
+In order to update the common submodule, we need to pull the latest version.
+
+```
+ cd common
+ git checkout dev
+ git pull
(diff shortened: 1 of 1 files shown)


docs/submodule-update.md
```diff
@@ -0,0 +1,11 @@
+
+In order to update the common submodule, we need to pull the latest version.
+
+```
+ cd common
+ git checkout dev
```
iambmelt, line 6: We might say "`dev` or whatever branch/commit is reqd" (since this value might not be `dev` in the case of a release or if there are experimental changes on a branch in common that you want to pick up in MSAL)

AdamBJohnsonx: Updated.

PR #1106, opened 2020-07-23
Post 1.5.5 Deployment Merge to Dev

iambmelt:
Completes the MSAL 1.5.5 deployment


msal/build.gradle
```diff
@@ -148,9 +148,9 @@ dependencies {
         transitive = false
     }
 
-    snapshotApi(group: 'com.microsoft.identity', name: 'common', version: '2.0.12-SNAPSHOT', changing: true)
+    snapshotApi(group: 'com.microsoft.identity', name: 'common', version: '2.0.12-hf1', changing: true)
```
shahzaibj, line 151: Did common already merge to dev? Can we change this to point to SNAPSHOT version?

iambmelt: Common doesn't need to merge to dev for this release, because it is already ahead of this release (we have already deployed 2.0.13, this release shipped 2.0.12-hf1 which doesn't contain all of those changes)

iambmelt: (This is to say, there shouldn't be any more 2.0.12-hf1 snapshots unless I am missing something?)

kreedula: Yup, there wont be any  2.0.12-hf1 snapshots since `common` dev is at `2.0.14` - so we can change to `2.0.14-SNAPSHOT`  for the nightly builds here

iambmelt: @kreedula In that case we can update both `distApi` and `snapshotApi`, right?

Such that:
```
snapshotApi(group: 'com.microsoft.identity', name: 'common', version: '2.0.14-SNAPSHOT', changing: true)
```
and
```
distApi("com.microsoft.identity:common:2.0.14") {
    transitive = false
}
```

Can you confirm my understanding?

kreedula: I am fine either ways, typically we update this when doing an MSAL release (i.e. we always have a newest common). This is weird case with hotfix. 

- We can leave it at `2.0.12-hf1` as it's the latest common MSAL is consuming in maven central or
- Given it's `dev` branch and we are updating it with latest common makes sense too.

iambmelt: @shahzaibj @kreedula Let's go ahead and rev this version to `2.0.14-SNAPSHOT` and `2.0.14` respectively. Even still, this state will be short-lived, as these will both be updated to `2.0.15` next week for the regularly scheduled release of `1.5.6`

iambmelt: Changes applied in 962c809d24e70f1e8e4885b7cdb70a5b228b54fa

PR #1111, opened 2020-07-26
Added showMsalRedirectUriInfo command and activity to MSAL

shoatman:
Hi Folks,

Given the very large number of customers confused about how to register their redirect URI and configure it within MSAL.  I added an activity to MSAL that can be launched using a static method of PublicClientApplication called showMsalRedirectUriInfo.  This method shows the following information.

@hamiltonha - Take a look if you want to change any strings.  Let's explore if we can replace current instructions with just invoking this method.

image

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -44,6 +44,7 @@ import com.microsoft.identity.client.exception.MsalArgumentException;
 import com.microsoft.identity.client.exception.MsalClientException;
 import com.microsoft.identity.client.exception.MsalDeclinedScopeException;
 import com.microsoft.identity.client.exception.MsalException;
+import com.microsoft.identity.client.helper.BrokerHelperActivity;
 import com.microsoft.identity.client.internal.AsyncResult;
 import com.microsoft.identity.client.internal.CommandParametersAdapter;
 import com.microsoft.identity.client.internal.controllers.MSALControllerFactory;
@@ -1213,6 +1214,10 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
         return BuildConfig.VERSION_NAME;
     }
 
+    public static void showMsalRedirectUriInfo(Activity activity){
+        activity.startActivity(BrokerHelperActivity.createStartIntent(activity.getApplicationContext()));
+    }
+
     @Override
     public PublicClientApplicationConfiguration getConfiguration() {
         return mPublicClientConfiguration;
@@ -1916,4 +1921,5 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
 
         return isAccountHomeTenant;
     }
+
 }
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -1214,7 +1214,11 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
         return BuildConfig.VERSION_NAME;
     }
 
-    public static void showMsalRedirectUriInfo(Activity activity){
+    /**
+     * Presents an activity that includes the package name, signature, redirect URI and manifest entry required for your application
+     * @param activity
+     */
+    public static void showExpectedMsalRedirectUriInfo(Activity activity){
         activity.startActivity(BrokerHelperActivity.createStartIntent(activity.getApplicationContext()));
     }
 
--- a/msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
+++ b/msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
@@ -1,0 +1,81 @@
+package com.microsoft.identity.client.helper;
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.Intent;
+import android.os.Bundle;
+import android.util.Log;
+import android.widget.TextView;
+
+import androidx.annotation.Nullable;
+
+import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
+import com.microsoft.identity.common.internal.broker.BrokerValidator;
+import com.microsoft.identity.common.internal.broker.PackageHelper;
+import com.microsoft.identity.msal.R;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+
+public class BrokerHelperActivity extends Activity {
+
+    TextView mPackageName;
+    TextView mSignature;
+    TextView mRedirect;
+    TextView mManifest;
+
+    String mManifestTemplate= "<activity android:name=\"com.microsoft.identity.client.BrowserTabActivity\">\n" +
+            "    <intent-filter>\n" +
+            "        <action android:name=\"android.intent.action.VIEW\" />\n" +
+            "        <category android:name=\"android.intent.category.DEFAULT\" />\n" +
+            "        <category android:name=\"android.intent.category.BROWSABLE\" />\n" +
+            "        <data\n" +
+            "            android:host=\"%s\"\n" +
+            "            android:path=\"%s\"\n" +
+            "            android:scheme=\"msauth\" />\n" +
+            "    </intent-filter>\n" +
+            "</activity>";
+
+    public static Intent createStartIntent(final Context context) {
+        final Intent intent = new Intent(context, BrokerHelperActivity.class);
+          return intent;
+    }
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.broker_helper);
+
+        mPackageName = findViewById(R.id.txtPackageName);
+        mSignature = findViewById(R.id.txtSignature);
+        mRedirect = findViewById(R.id.txtRedirect);
+        mManifest = findViewById(R.id.txtManifest);
+
+        mSignature.append(getSignature(false));
+        mRedirect.append(BrokerValidator.getBrokerRedirectUri(this.getApplicationContext(), this.getPackageName()));
+        mManifest.append(String.format(mManifestTemplate, this.getPackageName(), getSignature(false)));
+        mPackageName.append(this.getPackageName());
+
+
+    }
+
+    private String getSignature(Boolean urlEncode){
+
+        final PackageHelper info = new PackageHelper(this.getApplicationContext().getPackageManager());
+        final String signatureDigest = info.getCurrentSignatureForPackage(this.getPackageName());
+        String signature = "";
+
+        try {
+            if(urlEncode) {
+                signature = URLEncoder.encode(signatureDigest, AuthenticationConstants.ENCODING_UTF8);
+            }else{
+                signature = signatureDigest;
+            }
+        } catch (UnsupportedEncodingException e) {
+            Log.e("getSignature", "Error encoding signature digest", e);
+        }
+
+        return signature;
+
+    }
+}
--- a/msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
+++ b/msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
@@ -17,14 +17,12 @@ import com.microsoft.identity.msal.R;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 
+/**
+ * Activity to show the expected redirect URL (which includes support for Broker)
+ */
 public class BrokerHelperActivity extends Activity {
 
-    TextView mPackageName;
-    TextView mSignature;
-    TextView mRedirect;
-    TextView mManifest;
-
-    String mManifestTemplate= "<activity android:name=\"com.microsoft.identity.client.BrowserTabActivity\">\n" +
+    private final String MANIFEST_TEMPLATE = "<activity android:name=\"com.microsoft.identity.client.BrowserTabActivity\">\n" +
             "    <intent-filter>\n" +
             "        <action android:name=\"android.intent.action.VIEW\" />\n" +
             "        <category android:name=\"android.intent.category.DEFAULT\" />\n" +
@@ -35,10 +33,14 @@ public class BrokerHelperActivity extends Activity {
             "            android:scheme=\"msauth\" />\n" +
             "    </intent-filter>\n" +
             "</activity>";
+    TextView mPackageName;
+    TextView mSignature;
+    TextView mRedirect;
+    TextView mManifest;
 
     public static Intent createStartIntent(final Context context) {
         final Intent intent = new Intent(context, BrokerHelperActivity.class);
-          return intent;
+        return intent;
     }
 
     @Override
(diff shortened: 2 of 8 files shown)


msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
```diff
@@ -0,0 +1,81 @@
...
+                signature = URLEncoder.encode(signatureDigest, AuthenticationConstants.ENCODING_UTF8);
+            }else{
+                signature = signatureDigest;
+            }
+        } catch (UnsupportedEncodingException e) {
+            Log.e("getSignature", "Error encoding signature digest", e);
+        }
+
+        return signature;
+
```
AdamBJohnsonx, line 79: nit: extra line.

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
```diff
@@ -0,0 +1,81 @@
...
+        mSignature.append(getSignature(false));
+        mRedirect.append(BrokerValidator.getBrokerRedirectUri(this.getApplicationContext(), this.getPackageName()));
+        mManifest.append(String.format(mManifestTemplate, this.getPackageName(), getSignature(false)));
+        mPackageName.append(this.getPackageName());
+
+
+    }
+
+    private String getSignature(Boolean urlEncode){
```
AdamBJohnsonx, line 62: nits: final parameter,- can this be a primitive?

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
```diff
@@ -0,0 +1,81 @@
...
+        String signature = "";
+
+        try {
+            if(urlEncode) {
+                signature = URLEncoder.encode(signatureDigest, AuthenticationConstants.ENCODING_UTF8);
+            }else{
+                signature = signatureDigest;
+            }
+        } catch (UnsupportedEncodingException e) {
+            Log.e("getSignature", "Error encoding signature digest", e);
```
AdamBJohnsonx, line 75: I think we should be explicit here, since this is truly a "can't happen" case as UTF-8 has been folded into Java.  This isn't thrown for any random error encoding the string, it's thrown if the character set specified by name is not supported.  It's "Character encoding " + AuthenticationConstants.ENCODING_UTF8 + " is not supported."  You could even use Lombok to @SneakyThrows an UnsupportedEncodingException, supposing that the upper layer can responsibly handle a "WorldIsBrokenException".

msal/src/main/res/layout/broker_helper.xml
```diff
@@ -0,0 +1,111 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+
```
AdamBJohnsonx, line 8: nit: extra lines?

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
```diff
@@ -118,6 +122,8 @@ public View onCreateView(final LayoutInflater inflater, final ViewGroup containe
         mPopSection = view.findViewById(R.id.pop_section);
         mLoginHintSection = view.findViewById(R.id.login_hint_section);
 
+
+
```
AdamBJohnsonx, line 126: nit: extra lines

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
```diff
@@ -0,0 +1,81 @@
...
+        mSignature = findViewById(R.id.txtSignature);
+        mRedirect = findViewById(R.id.txtRedirect);
+        mManifest = findViewById(R.id.txtManifest);
+
+        mSignature.append(getSignature(false));
+        mRedirect.append(BrokerValidator.getBrokerRedirectUri(this.getApplicationContext(), this.getPackageName()));
+        mManifest.append(String.format(mManifestTemplate, this.getPackageName(), getSignature(false)));
+        mPackageName.append(this.getPackageName());
+
+
```
AdamBJohnsonx, line 59: nit: extra lines?

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
```diff
@@ -0,0 +1,81 @@
...
+import java.net.URLEncoder;
+
+public class BrokerHelperActivity extends Activity {
+
+    TextView mPackageName;
+    TextView mSignature;
+    TextView mRedirect;
+    TextView mManifest;
+
+    String mManifestTemplate= "<activity android:name=\"com.microsoft.identity.client.BrowserTabActivity\">\n" +
```
AdamBJohnsonx, line 27:
```suggestion
    String mManifestTemplate = "<activity android:name=\"com.microsoft.identity.client.BrowserTabActivity\">\n" +

```

Om83:
hi @shoatman while I understand the concern, I am wondering if the activity is the best way to resolve this. Few things to consider:
1- This would require localization support
2- This is also a little bit of chicken and egg problem. Are we asking users to register the app first or are we asking them to use this implementation in their app to get the right values and then go for registration. Which is different how our documentation exists in many places and how other platforms attempt it.

My thinking is that whole confusion arises because user has to enter the value in two different places (AndroidManifest and the config.json) and the URL encoding being used in redirect URI where as that is not the case when we are recording the signature hash. Further the tutorial suggests that the path contain a slash at the beginning.

I would suggest that we just capture packageName and signatureHash from the user compute the rest ourself creating a consistent behavior with the portal.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1213,6 +1214,10 @@ public static String getSdkVersion() {
         return BuildConfig.VERSION_NAME;
     }
 
+    public static void showMsalRedirectUriInfo(Activity activity){
```
shahzaibj, line 1217: Needs javadoc.....Also should we rename this to something like `showExpectedMsalRedirectUriInfo` (to indicate that the redirect uri displayed here isn't the one they have already registered (in case they have) but rather the one that we actually expect)?

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
```diff
@@ -0,0 +1,81 @@
...
+
+import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
+import com.microsoft.identity.common.internal.broker.BrokerValidator;
+import com.microsoft.identity.common.internal.broker.PackageHelper;
+import com.microsoft.identity.msal.R;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+
+public class BrokerHelperActivity extends Activity {
```
shahzaibj, line 20: Could add some javadoc here

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
shahzaibj, line 27: Could be `private`

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
```diff
@@ -0,0 +1,81 @@
...
+
+        try {
+            if(urlEncode) {
+                signature = URLEncoder.encode(signatureDigest, AuthenticationConstants.ENCODING_UTF8);
+            }else{
+                signature = signatureDigest;
+            }
+        } catch (UnsupportedEncodingException e) {
+            Log.e("getSignature", "Error encoding signature digest", e);
+        }
```
shahzaibj, line 76: Do we also want to display an error on the `BrokerHelperActivity` when this happens? Looks like an empty string will be shown

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
AdamBJohnsonx: ... and final.  and a constant?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: I like that suggestion.  Thanks

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
shoatman: ack

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
shoatman: ack

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
shoatman: ack

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
shoatman: I don't expect this to happen... ;-).  Ack on suggestion for logging message.

msal/src/main/res/layout/broker_helper.xml
shoatman: ack

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
shoatman: ack

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
shoatman: ack

shoatman: ack, ack

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
shoatman: Hadn't considered that since I don't expect this to happen... wondering in this case if I should convert this to a runtime exception.

shoatman:
> hi @shoatman while I understand the concern, I am wondering if the activity is the best way to resolve this. Few things to consider:
> 1- This would require localization support
> 2- This is also a little bit of chicken and egg problem. Are we asking users to register the app first or are we asking them to use this implementation in their app to get the right values and then go for registration. Which is different how our documentation exists in many places and how other platforms attempt it.
> 
> My thinking is that whole confusion arises because user has to enter the value in two different places (AndroidManifest and the config.json) and the URL encoding being used in redirect URI where as that is not the case when we are recording the signature hash. Further the tutorial suggests that the path contain a slash at the beginning.
> 
> I would suggest that we just capture packageName and signatureHash from the user compute the rest ourself creating a consistent behavior with the portal.

I don't think localization is required in the short term.  It certainly wouldn't hurt.
I don't think is a chicken and egg problem.  You have to have a signing key to register your app... so that  comes first... and many people are confused about which signing key is being used to sign their app.  With this approach... you cannot get it wrong.

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
AdamBJohnsonx: If you sneaky it out it'll land in the same place. ;-)

shoatman: I chose to convert checked exception to runtime in this case.  So I'll log and then throw a runtime exception.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1213,6 +1214,14 @@ public static String getSdkVersion() {
         return BuildConfig.VERSION_NAME;
     }
 
+    /**
+     * Presents an activity that includes the package name, signature, redirect URI and manifest entry required for your application
```
shahzaibj, line 1218: may be `shows` instead of `includes`?

msal/src/main/java/com/microsoft/identity/client/helper/BrokerHelperActivity.java
```diff
@@ -0,0 +1,81 @@
...
+
+        try {
+            if (urlEncode) {
+                signature = URLEncoder.encode(signatureDigest, AuthenticationConstants.ENCODING_UTF8);
+            } else {
+                signature = signatureDigest;
+            }
+        } catch (UnsupportedEncodingException e) {
+            Log.e("getSignature", "Character encoding " + AuthenticationConstants.ENCODING_UTF8 + " is not supported.", e);
+            throw new RuntimeException("Unexpected: Unable to get the signature for this application package.");
```
AdamBJohnsonx, line 76: should we not attach the causing exception?

shahzaibj:
> This PR seems to largely remove the need for the `package-inspector` utility; should we retire that? Or perhaps file a backlog item to remove/repurpose that utility?

@iambmelt Package Inspector is much more generic. I've been using it recently to find out exact package names for first part apps such as Outlook, Teams etc as I've added UI Tests for those. I would prefer to keep the package-inspector.

iambmelt:
> > This PR seems to largely remove the need for the `package-inspector` utility; should we retire that? Or perhaps file a backlog item to remove/repurpose that utility?
> 
> @iambmelt Package Inspector is much more generic. I've been using it recently to find out exact package names for first part apps such as Outlook, Teams etc as I've added UI Tests for those. I would prefer to keep the package-inspector.

OK! I'm glad you're getting some use out of it! In that case, let's keep it!

(And let me know if you have any feature requests! 😉)

rpdome:
My take on this is that we might not need the UI. 
In fact, if I were a developer, I would want a function that would just print something to logcat (or some text file) that I could just copy & paste the values (as opposed to read & type values manually - which could be error-prone)

shoatman:
> My take on this is that we might not need the UI.
> In fact, if I were a developer, I would want a function that would just print something to logcat (or some text file) that I could just copy & paste the values (as opposed to read & type values manually - which could be error-prone)

I think we can do both.  It is possible to cut and paste from the UI... i enabled it. I'm happy to also output the same information to logcat.

Om83:
We already have FAQ for redirect URI https://github.com/AzureAD/microsoft-authentication-library-for-android/wiki/MSAL-FAQ. We should update it to capture how to use the activity and where to use these values.

PR #1124, opened 2020-08-07
Run checkstyle, pmd, lint, javadocJar, and sourcesJar also for assembleDistRelease

iambmelt:
See title.

Relevant workitem:
- [PBI-1096031](https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20Android/Backlog%20items/?workitem=1096031) (corpnet reqd)

There will be another change to accompany this in the build pipeline. That change can't be PR'd, as it's a build definition change on the pipeline


msal/build.gradle
```diff
@@ -277,7 +277,7 @@ publishing {
 }
 
 tasks.whenTaskAdded { task ->
-    if (task.name == 'assembleDebug' || task.name == 'assembleRelease') {
+    if (task.name == 'assembleDebug' || task.name == 'assembleRelease' || task.name == 'assembleDistRelease') {
```
shahzaibj, line 280: What about the other tasks?

shahzaibj: There are more assemble tasks - let's run for all of them

shahzaibj, line 280:
```suggestion
    if (task.name.contains("assemble)) {
```

shahzaibj: We can do something like this I think

kreedula: Good point, that way we catch `lint` and other errors early

iambmelt: These are our current `assemble` variants -- I don't think we necessarily want to enable this for `assembleSnapshot` (as it will slow that build down, generate unused artifacts). We could do `task.name.contains('assemble') && !task.name.equals('assembleSnapshot')`.

Thoughts? @shahzaibj

```
brianmel@brianmel-devbox MINGW64 /c/dev/android-complete/msal (iambmelt/simple-build-prep)
$ ./gradlew tasks | egrep assemble
assemble - Assemble main outputs for all the variants.
assembleAndroidTest - Assembles all the Test applications.
assembleDebug - Assembles main outputs for all Debug variants.
assembleDist - Assembles main outputs for all Dist variants.
assembleLocal - Assembles main outputs for all Local variants.
assembleRelease - Assembles main outputs for all Release variants.
assembleSnapshot - Assembles main outputs for all Snapshot variants.

```

shahzaibj: I think there are even more tasks actually:

```
- assemble
- assembleLocal
- assembleLocalDebug
- assembleLocalRelease
- assembleSnapshot
- assembleSnapshotDebug
- assembleSnapshotRelease
- assembleDist
- assembleDistDebug
- assembleDistRelease 
```
There's actually even more tasks for testing (`assembleLocalReleaseUnitTest` and many more)....I didn't list them here.

If we want to avoid running for snapshot, then I think we want to avoid for each build type of snapshot (debug vs release vs task for assemble test).

May be we can do this:

`if (task.name.contains('assemble') && !task.name.contains('Snapshot'))`

but most likely we also want to avoid the test tasks so then, we can do like:

`if (task.name.contains('assemble') && !task.name.contains('Snapshot') && !task.name.contains('Test'))`

I'd leave up to you to decide how want to do it @iambmelt .

Also if @kreedula has any input.

msal/build.gradle
```diff
@@ -277,7 +277,7 @@ publishing {
 }
 
 tasks.whenTaskAdded { task ->
-    if (task.name == 'assembleDebug' || task.name == 'assembleRelease') {
+    if (task.name.contains('assemble') && !task.name.contains('Snapshot') && !task.name.contains('Test')) {
```
kreedula, line 280: I think it's better to have two separate if loops

`if (task.name.contains('assemble')) {
                task.dependsOn 'checkstyle', 'pmd', 'lint'
    }`

`if (task.name.contains('assemble') && !task.name.contains('Snapshot') && !task.name.contains('Test')) {
        task.dependsOn  'javadocJar', 'sourcesJar'
    }	    }`

iambmelt: 653bf3835853ae48449895940f772c68164a4856

msal/build.gradle
iambmelt: See 653bf3835853ae48449895940f772c68164a4856

msal/build.gradle
```diff
@@ -277,7 +277,14 @@ publishing {
...
-    if (task.name == 'assembleDebug' || task.name == 'assembleRelease') {
-        task.dependsOn 'checkstyle', 'pmd', 'lint', 'javadocJar', 'sourcesJar'
+    if (task.name.contains('assemble')) {
+        task.dependsOn 'checkstyle', 'pmd'
+    }
+
+    if (task.name.contains('assemble')
+        && !task.name.contains('Snapshot')
+        && !task.name.contains('Test')
+        && !task.name.contains('Local')) {
```
AdamBJohnsonx, line 287: Don't we want lint for local?

iambmelt: See commit message for da2fc37bfbca71501f556f8baaa723d1df72d0a9

Running `lint` was also running `dist` tasks for reasons I didn't fully troubleshoot

iambmelt: (Try it locally, you'll see a `dist` build kickoff and subsequently fail)

PR #1126, opened 2020-08-10
Expose RT to broker apps

rpdome:
Allowing Broker apps to extract BRT to perform WPJ.

related PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/996


shahzaibj:
@rpdome Can you add more details? Perhaps a link to an ADO work item? I'm not clear on what we are trying to accomplish and why? (Broker apps aren't using MSAL right? how is this helping them?)

rpdome:
> @rpdome Can you add more details? Perhaps a link to an ADO work item? I'm not clear on what we are trying to accomplish and why? (Broker apps aren't using MSAL right? how is this helping them?)

They will, for device code flow.
https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20Android/Backlog%20items/?workitem=1076089

changelog
```diff
@@ -2,6 +2,7 @@ MSAL Wiki : https://github.com/AzureAD/microsoft-authentication-library-for-andr
 
 Version v.Next
 ----------
+- Expose RT to Broker apps.
```
shoatman, line 5: You may want to say broker refresh token here explicitly.  As well as indicate why this support is needed.  Once it's shipped any broker app will be able to use this... in any scenario, but it would be good to know which scenario originally required it.

msal/src/main/java/com/microsoft/identity/client/IBrokerRtAccessor.java
```diff
@@ -0,0 +1,38 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+package com.microsoft.identity.client;
+
+import androidx.annotation.NonNull;
+
+import com.microsoft.identity.client.exception.MsalClientException;
+
+public interface IBrokerRtAccessor {
```
shoatman, line 29: I'd personally spell out refresh token here.  But this is an opinion, not required.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1121,6 +1138,22 @@ private void setupTelemetry(@NonNull final Context context,
                 .build();
     }
 
+    @Override
+    public String getBrokerRt(@NonNull final String accountObjectId) throws MsalClientException {
```
shoatman, line 1142: Why does this need to be part of the public API of MSAL?  Is there an objection to this being accessed via an internal or private method?

rpdome: CP and AuthApp has to call it. Currently they're using ADAL to obtain Broker AT/RT, then pass it into WPJ's Join() operation as parameters.

iambmelt: @rpdome @shoatman We could do like we did for TSL APIs and hide it away.... There's an `ITokenShare` API that `PublicClientApplication` needs to be cast to before those methods can be used. I think exposing this method here _and_ not allowing non-broker apps to call it is going to be incredible frustrating for those developers who have requested we expose RT in the past

rpdome: That's what I'm aiming for. IBrokerRefreshTokenAccessor = ITokenShare.

iambmelt: @rpdome Sounds just fine to me then 👍

shoatman: @iambmelt - One of the goals of our libraries is not to expose refresh tokens.  Anyone can access the refresh token if they really want to... but we don't want to make it easy.  Also, what will we do when we enable refresh token PoP?  Are we planning to allow them to export the private key as well?

shoatman: @rpdome - I think we should have a quick meeting to discuss this in more detail.

iambmelt: @rpdome @shoatman Agree, let's sync on these points

rpdome: Updated as synced.

msal/src/main/java/com/microsoft/identity/client/internal/BrokerClientIdRefreshTokenAccessor.java
```diff
@@ -0,0 +1,118 @@
...
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+package com.microsoft.identity.client.internal;
```
iambmelt, line 24: Do we want to put this in something like `com.microsoft.identity.client.internal.api` ?

Issue #1128, opened 2020-08-13
report legacy MSAL redirect uri more prominently and descriptively


clockworkant:
I spent quite a few hours getting frustrated as to why the broker (MS Authenticator) wasn't being called. Everything looked fine in the configuration and the manifest, however it turned out that the redirect uri was incorrect. I had to debug the library to discover the following information in a comment. It's very frustrating as i was seeing the mUseBroker value being toggled to false even when i'd explictly set it to try in the configure file

PublicClientApplicationConfiguration.java line 554
`
if (!isBrokerRedirectUri()) {
            // This means that the app is still using the legacy local-only MSAL Redirect uri (already removed from the new portal).
            // If this is the case, we can assume that the user doesn't need Broker support.
            Logger.info(TAG, "The app is still using legacy MSAL redirect uri. Switch to MSAL local auth.");
            mUseBroker = false;
            return;
        }
`
I want some logging in the logcat that says that the broker can't be used. Maybe output the expected value for the redirectUri as well so it can be used to update the configs. The current log message is far too vague!!

shoatman:
Adam made the message a warning here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1130

We also added a helper method for getting the expected values for the broker:

https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1111

Sorry for the frustrating experience!

clockworkant:
Thanks for the quick turn around on this issue. 🙏

Sorry to be a pain but is it possible to highlight in the warning that the broker is being turned off when in legacy mode. Currently that piece of information is somewhat hidden as a comment in the code above the warning message:
`// If this is the case, we can assume that the user doesn't need Broker support.`

It was a real shock to me that the method `checkIntentFilterAddedToAppManifestForBrokerFlow()` that contains this had the side effect of changing the value for `mUseBroker` to false. This is why I would like it highlighted in the warning message so future developers don't have to read the sourcecode to see this side effect.

PR #1130, opened 2020-08-14
Better messaging on broker redirect uri mismatch.

AdamBJohnsonx:


msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -480,10 +483,11 @@ private void nullConfigurationCheck(String configKey, String configValue) {
     }
 
     private boolean isBrokerRedirectUri() {
-        final String BROKER_REDIRECT_URI_REGEX = "msauth://" + mAppContext.getPackageName() + "/.*";
-        final Pattern pairRegex = Pattern.compile(BROKER_REDIRECT_URI_REGEX);
-        final Matcher matcher = pairRegex.matcher(mRedirectUri);
-        return matcher.matches();
+        final Matcher matcher = BROKER_REDIRECT_URI_REGEX.matcher(mRedirectUri);
+        if (matcher.matches() && ObjectUtils.equals(mAppContext.getPackageName(), matcher.group(1))) {
```
iambmelt, line 487: Optionally return the entire expression?

```java
return matcher.matches() && ObjectUtils.equals(mAppContext.getPackageName(), matcher.group(1));
```

AdamBJohnsonx: Yes, that's better.

AdamBJohnsonx:
This is in reference to https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/1128

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -551,10 +555,12 @@ public void checkIntentFilterAddedToAppManifestForBrokerFlow() throws MsalClient
             return;
         }
 
-        if (!isBrokerRedirectUri()) {
+        if (!isBrokerRedirectUri(mRedirectUri, mAppContext.getPackageName())) {
             // This means that the app is still using the legacy local-only MSAL Redirect uri (already removed from the new portal).
             // If this is the case, we can assume that the user doesn't need Broker support.
-            Logger.info(TAG, "The app is still using legacy MSAL redirect uri. Switch to MSAL local auth.");
+            Logger.info(TAG, "The app is still using legacy MSAL redirect uri. Switch to MSAL local auth."
```
shoatman, line 561: See that this was an info message previously.  We may want to warn here instead.

AdamBJohnsonx, line 561: Agreed.
```suggestion
            Logger.warn(TAG, "The app is still using legacy MSAL redirect uri. Switch to MSAL local auth."
```

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -78,6 +81,8 @@
 public class PublicClientApplicationConfiguration {
     private static final String TAG = PublicClientApplicationConfiguration.class.getSimpleName();
 
+    private static final Pattern BROKER_REDIRECT_URI_REGEX = Pattern.compile("msauth://([^/]*)/.*");
```
shahzaibj, line 84: Is this meant to be used outside of the context of an Android app? Because this regex seems to accept invalid android application package names

AdamBJohnsonx: Not really - it's private. But the idea is that this pulls out the authority segment of a URI with the correct scheme, and then we compare that string with equals on the package name. But there's only one usage of it...

shahzaibj: Okay, I think it is fine as this only being used in this class....but just wanted to point out that the group 1 which is supposed to match the package name, seems to accept anything but a slash, so I think this would accept redirect uri that are formed via package names that aren't actually valid Android application ids....but I think since this is all internal it is probably okay. However, it may still be good to make it more specific. 

FYI, Application id aka "Android app package name" naming conventions are listed here: https://developer.android.com/studio/build/application-id

AdamBJohnsonx: Yeah, I realize what it was doing... but my mind was in the regex prison, and we don't need to use a pattern at all.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -551,10 +555,12 @@ public void checkIntentFilterAddedToAppManifestForBrokerFlow() throws MsalClient
...
         }
 
-        if (!isBrokerRedirectUri()) {
+        if (!isBrokerRedirectUri(mRedirectUri, mAppContext.getPackageName())) {
             // This means that the app is still using the legacy local-only MSAL Redirect uri (already removed from the new portal).
             // If this is the case, we can assume that the user doesn't need Broker support.
-            Logger.info(TAG, "The app is still using legacy MSAL redirect uri. Switch to MSAL local auth.");
+            Logger.warn(TAG, "The app is still using legacy MSAL redirect uri. Switch to MSAL local auth."
+                + "  For brokered auth, the redirect URI is expected to conform to 'msauth://<authority>/.*' where the authority in "
+                + "that uri is the package name.");
```
shahzaibj, line 563:
```suggestion
            Logger.warn(TAG, "The app is still using legacy MSAL redirect uri. Switch to MSAL local auth."
                + "  For brokered auth, the redirect URI is expected to conform to 'msauth://<authority>/.*' where the authority in "
                + "that uri is the package name of the app. This package name is listed as 'applicationId' in the build.gradle file.");
```

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -479,11 +484,10 @@ private void nullConfigurationCheck(String configKey, String configValue) {
         }
     }
 
-    private boolean isBrokerRedirectUri() {
-        final String BROKER_REDIRECT_URI_REGEX = "msauth://" + mAppContext.getPackageName() + "/.*";
-        final Pattern pairRegex = Pattern.compile(BROKER_REDIRECT_URI_REGEX);
-        final Matcher matcher = pairRegex.matcher(mRedirectUri);
-        return matcher.matches();
+    @VisibleForTesting
+    public static boolean isBrokerRedirectUri(final @NonNull String redirectUri, final @Nullable String packageName) {
```
shahzaibj, line 488: Does the package name have to be nullable?

AdamBJohnsonx: There's no requirement for the restriction to NonNull.

shahzaibj: So what should be return from this method if package name is null? I don't see a null check here

AdamBJohnsonx: It doesn't matter, so I'm making it @NonNull

PR #1132, opened 2020-08-19
Consume Test Rule to write results to CSV + Enable Logging for Robolectric Tests

shahzaibj:
See this for complete context: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1025


msal/src/test/java/com/microsoft/identity/client/e2e/rules/RobolectricLoggingRule.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+import org.robolectric.shadows.ShadowLog;
+
+/**
+ * A JUnit rule to enable logging during a Robolectric Test.
+ */
+public class RobolectricLoggingRule implements TestRule {
```
rpdome, line 35: should we put this in common?

rpdome (approved):

msal/src/test/java/com/microsoft/identity/client/e2e/rules/RobolectricLoggingRule.java
shahzaibj: Ideally yes, but unfortunately we can't put Robolectric as a direct dependency in common, whenever we tried that in the past, we ran into issues so for that reason we have to put Robolectric utilities directly in MSAL, and Broker repos.

AdamBJohnsonx: Should we create a new package that any of them can depend on, then?

If we can't do:
<pre>
MSAL -> Common (with Utilities) -> Robolectric
         ^
         |
Broker --+
</pre>

Can we make it:

<pre>
MSAL/Broker -> Common
            |
            +-> RobolectricTestUtils (or whatever)
</pre>

PR #1134, opened 2020-08-20
Resurrect b2c ui tests

shahzaibj:
- Remove Load Lab User Rule and move the logic to `@Before` to allow creating more dynamic lab queries (as needed in the case of parameterized tests such as the B2C Test)
- Remove ignore annot from B2C Test


msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/AbstractMsalUiTest.java
```diff
@@ -85,35 +86,46 @@
...
     }
 
+    private void loadLabUser() {
+        if (getLabUserQuery() != null) {
+            mLoginHint = LabUserHelper.loadUserForTest(getLabUserQuery());
+        } else if (getTempUserType() != null) {
+            mLoginHint = LabUserHelper.loadTempUser(getTempUserType());
+            try {
+                // temp user takes some time to actually being created even though it may be
+                // returned by the LAB API. Adding a wait here before we proceed with the test.
```
iambmelt, line 119: 😞

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/AbstractMsalUiTest.java
```diff
@@ -85,35 +86,46 @@
...
+                Thread.sleep(TEMP_USER_WAIT_TIME);
```
AdamBJohnsonx, line 120: Should we bake this into loadTempUser?

shahzaibj: I'm inclined not to do that for now, and instead let the caller decide if they want to sleep or not. Because let's say if someone called it but wasn't going to use the temp account right away and instead used it some time later (depending on how they have their test), they may not want/need the sleep.

PR #1137, opened 2020-08-25
Add device code flow support to MSAL Test app

rpdome:
- Added a button for triggering DCF.
- Added a "Status" section which displays the latest toast message.

Screen Shot 2020-08-24 at 6 13 08 PM

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
@@ -91,7 +91,7 @@ public interface IPublicClientApplication {
      * @param scopes the desired access scopes
      * @param callback callback object used to communicate with the API throughout the protocol
      */
-    void acquireTokenWithDeviceCode(@Nullable String[] scopes, @NonNull final DeviceCodeFlowCallback callback);
+    void acquireTokenWithDeviceCode(@NonNull String[] scopes, @NonNull final DeviceCodeFlowCallback callback);
 
     /**
      * Returns the PublicClientConfiguration for this instance of PublicClientApplication.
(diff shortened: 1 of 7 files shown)


msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
```diff
@@ -91,7 +91,7 @@ void acquireToken(@NonNull final Activity activity,
      * @param scopes the desired access scopes
      * @param callback callback object used to communicate with the API throughout the protocol
      */
-    void acquireTokenWithDeviceCode(@Nullable String[] scopes, @NonNull final DeviceCodeFlowCallback callback);
+    void acquireTokenWithDeviceCode(@NonNull String[] scopes, @NonNull final DeviceCodeFlowCallback callback);
```
shahzaibj, line 94: 👍

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalWrapper.java
```diff
@@ -193,6 +194,33 @@ public void acquireTokenSilentWithResource(@NonNull RequestOptions requestOption
 
     abstract void acquireTokenSilentAsyncInternal(@NonNull final AcquireTokenSilentParameters parameters);
 
+    public void acquireTokenSilentWithDeviceCodeFlow(@NonNull RequestOptions requestOptions,
```
shahzaibj, line 197: nit: rename to `acquireTokenWithDeviceCodeFlow`

shahzaibj:
So status is NOT displaying a toast, and rather displaying in the TextView, right?

Also what does the Copy button copy? The user code or the uri or both?

rpdome:
> So status is NOT displaying a toast, and rather displaying in the TextView, right?
> 
> Also what does the Copy button copy? The user code or the uri or both?

I'm keeping the toast. The TextView will hold the messages that is displayed by the toast. 
The copy button will copy the whole message. (So now you can copy both code and URI in this scenario, or error code/message when an error is returned).

PR #1138, opened 2020-08-25
Update method calls to account for ExecutionException added to ResultFuture

iambmelt:
Extension of:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1012

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
@@ -50,7 +50,9 @@ import com.microsoft.identity.common.internal.migration.TokenMigrationCallback;
 import com.microsoft.identity.common.internal.result.ResultFuture;
 
 import java.util.List;
+import java.util.concurrent.ExecutionException;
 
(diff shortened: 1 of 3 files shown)


msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
```diff
@@ -160,12 +162,21 @@ public void onError(MsalException exception) {
             }
         }, PublicApiId.MULTIPLE_ACCOUNT_PCA_GET_ACCOUNTS);
 
-        final AsyncResult<List<IAccount>> result = future.get();
+        try {
+            final AsyncResult<List<IAccount>> result = future.get();
```
AdamBJohnsonx, line 166: This is not a new problem, just something to think about.  This is an unbounded-in-time get on a future - it could block forever if you let it, though if the scope is simple enough you can understand it all it might be OK.  In the past, in similar situations, I've allowed clients to specify a timeout in their configuration (supplying a default that was well-known and safe), and then used a custom future implementation that blocked for only that amount of time (specified at construction), yielding a TimeoutException when appropriate.  We might want to consider that kind of approach.

PR #1141, opened 2020-08-26
Define async result success as not getting an exception

shahzaibj:
Address https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1122071

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/AsyncResult.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/AsyncResult.java
@@ -29,26 +29,21 @@ import com.microsoft.identity.client.exception.MsalException;
 public class AsyncResult<T> {
     private T mResult;
     private MsalException mMsalException;
-    private boolean mSuccess = false;
 
-    public AsyncResult(@Nullable T result, @Nullable MsalException exception){
+    public AsyncResult(@Nullable T result, @Nullable MsalException exception) {
         mResult = result;
         mMsalException = exception;
-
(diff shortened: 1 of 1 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/AsyncResult.java
```diff
@@ -29,26 +29,21 @@
 public class AsyncResult<T> {
     private T mResult;
     private MsalException mMsalException;
-    private boolean mSuccess = false;
 
-    public AsyncResult(@Nullable T result, @Nullable MsalException exception){
+    public AsyncResult(@Nullable T result, @Nullable MsalException exception) {
         mResult = result;
         mMsalException = exception;
```
iambmelt, line 35: Optionally, we could consider throwing a RuntimeException if both params are `null` (as this would be a bug and should never happen)

shahzaibj: Isn't this what we discussed yesterday and we concluded that we would allow both to be `null` as is the case for the `getAccount(identifier)` API? (it returns `null` without exception if an account for the identifier doesn't exist).

iambmelt: 🤦‍♂️ 

Duh!

iambmelt: Long day 😆

Issue #1122, opened 2020-08-06
"prompt" parameter not set correctly during authentication


koesclem:
The MSAL for Android allows to provide a `Prompt` parameter during interactive token requests by using `SingleAccountPublicClientApplication.signIn` or by providing `AcquireTokenParameters`. Possible `Prompt` parameters are: `[SELECT_ACCOUNT, LOGIN, CONSENT, WHEN_REQUIRED]`. As per documentation of the MSAL for Android's code those should map to their OIDC specification counterparts: `[select_account, login, consent, none]`.

Of those mentioned prompt parameters, all seem to work as expected, except for the `WHEN_REQUIRED` `Prompt`. The javadoc states the following:

```
/**
 * acquireToken will not send the prompt parameter to the authorize endpoint.  The user may be prompted to login or to consent as required by the request.
 */
```

After a `Prompt` `WHEN_REQUIRED` is supplied it is later overridden by the MSAL when creating the `InteractiveTokenCommandParameters` in the `CommandParametersAdapter.createInteractiveTokenCommandParameters` method using the following code segment:

```
private static OpenIdConnectPromptParameter getPromptParameter(@NonNull final AcquireTokenParameters parameters) {
    if (parameters.getPrompt() == null || parameters.getPrompt() == Prompt.WHEN_REQUIRED) {
        return OpenIdConnectPromptParameter.SELECT_ACCOUNT;
    } else {
        return parameters.getPrompt().toOpenIdConnectPromptParameter();
    }
}
```

It seems there is no option to provide the OIDC specification's `prompt=none` parameter and therefore there is no way of performing an authentication without the need for the user to input their credentials during login, even though there might be a valid session provided by the browser. You can also observe the set `prompt` parameter by inspecting the resulting authentication url opened in the Browser/CustomTab. 

In fact, even when fixing the above code segment the `Prompt.toOpenIdConnectPromptParameter()` method never results in an `OpenIdConnectPromptParameter.NONE` parameter since it seems to not support a `Prompt.WHEN_REQUIRED`.

MSAL Version: 1.6.0

kreedula:
Thank you for reporting this.  I have added a backlog item to fix or infact see if we can add support for `prompt = none`

tanmaymanolkar1:
[...]

PR #1144, opened 2020-08-28
Adding the functionality of Prompt Parameter UNSET

tanmaymanolkar1:
Fix for the github issue #1122

Added support for OpenIdConnectPromptParameter.NONE and reworked the logic to handle Prompt.WHEN_REQUIRED


shahzaibj:
@tanmaymanolkar1 So the customer is happy with this fix/naming convention? The customer seemed to be looking specifically for something called as `NONE` that does the same thing as `UNSENT`

tanmaymanolkar1:
@shahzaibj the customer just needs a way to hit the NONE parameter, which is being provided. The fixing of the naming is done so that the mapping internally is consistent and less obscure.

PR #1152, opened 2020-09-14
Add metadata to TSL result payload

iambmelt:
Extends work in:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1033

Exposes RT/Account metadata such that it can be consumed by TSL.
Adds additional metadata to _external_ TSL-API.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -48,11 +48,11 @@ import com.microsoft.identity.client.exception.MsalServiceException;
 import com.microsoft.identity.client.helper.BrokerHelperActivity;
 import com.microsoft.identity.client.internal.AsyncResult;
 import com.microsoft.identity.client.internal.CommandParametersAdapter;
-import com.microsoft.identity.common.internal.controllers.LocalMSALController;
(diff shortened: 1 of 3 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1132,12 +1133,13 @@ private void setupTelemetry(@NonNull final Context context,
     }
 
     @Override
-    public String getOrgIdFamilyRefreshToken(@NonNull final String identifier) throws MsalClientException {
+    public TokenShareResult getOrgIdFamilyRefreshTokenWithMetadata(@NonNull final String identifier) throws MsalClientException {
```
shoatman, line 1136: Is this public API change?   I'd talk to Teams to see if they are will to take this as a change without incrementing the major version number... unless we can bundle this with the prompt and other breaking changes.

shoatman (changes requested):
Not sure if we need changes here or if we just need to put this in a separate feature branch with other breaking changes.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shoatman: Resolved offline

shoatman (approved):
lgtm

PR #1156, opened 2020-09-18
Enable verbose logging for Robolectric tests

shahzaibj:
See title

### DIFF
--- a/msal/src/test/java/com/microsoft/identity/client/e2e/tests/PublicClientApplicationAbstractTest.java
+++ b/msal/src/test/java/com/microsoft/identity/client/e2e/tests/PublicClientApplicationAbstractTest.java
@@ -28,6 +28,7 @@ import android.content.Context;
 import androidx.test.core.app.ApplicationProvider;
 
 import com.microsoft.identity.client.IPublicClientApplication;
+import com.microsoft.identity.client.Logger;
 import com.microsoft.identity.client.PublicClientApplication;
 import com.microsoft.identity.client.exception.MsalException;
 import com.microsoft.identity.common.internal.controllers.CommandDispatcherHelper;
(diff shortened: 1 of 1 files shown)


msal/src/test/java/com/microsoft/identity/client/e2e/tests/PublicClientApplicationAbstractTest.java
```diff
@@ -53,6 +54,9 @@ public void setup() {
         mContext = ApplicationProvider.getApplicationContext();
         mActivity = TestUtils.getMockActivity(mContext);
         setupPCA();
+        Logger.getInstance().setEnableLogcatLog(true);
+        Logger.getInstance().setEnablePII(true);
+        Logger.getInstance().setLogLevel(Logger.LogLevel.VERBOSE);
```
AdamBJohnsonx, line 59: No problem here, but should we have a data-driven configuration for this logger?

AdamBJohnsonx (approved):

msal/src/test/java/com/microsoft/identity/client/e2e/tests/PublicClientApplicationAbstractTest.java
shahzaibj: Not sure what you mean, can you elaborate a bit more?

AdamBJohnsonx: Often, logging parameters are set up via a file or other resource external to the application, not strictly in code.  I see that this is wrapping android's logging, which I guess we're trying to shield from too much messaging.  I was expecting to see a change to logging config.  It seems like we might want one?  Normally, in these configurations, you can turn on or off logging or control log level by class or package (in this case it would be tag).

shahzaibj: We could also do it through the MSAL config file. See this: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-configuration#logging

Although that does similar things to what can be achieved via code. We do not have class level or package level config at this time.

rpdome (approved):

PR #1158, opened 2020-09-22
For msal automation use common pulled with MSAL

shahzaibj:
Exclude common from testutilities. So that when we are testing dist of MSAL, then we can use the common pulled via MSAL dist.

Ideally ultimately we need to refactor testutilities to avoid it taking common as a dependency.


AdamBJohnsonx:
As a point of curiosity, why avoid taking common as a dependency?  Is there too much business logic in common, and should we create a separate utilities package so that truly common things can just be shared?

shahzaibj:
> As a point of curiosity, why avoid taking common as a dependency? Is there too much business logic in common, and should we create a separate utilities package so that truly common things can just be shared?

I just want to split `testutils` into two separate modules, once that depends on common and the other that doesn't depend on common. `testutils` was initially created for the specific purpose that we needed something that could consume common, primarily because we needed such behavior for Robolectric tests. Now that it is being used more broadly in UI Automation in native msal, broker as well as in OneAuth.....I think there is always a chance to run into dependency conflicts with different versions of common being pulled. That's why I want to avoid it....also because the dependency of common isn't really needed for UI Automation (at least thus far), so we can have two completely separate testutilities libraries, one for Robolectric tests and one for UI Automation. We're kind of already in that direction with the inception of the `uiautomationutilities` but it internally depends on `testutils` for some things...so we just need to get rid of that I think.

AdamBJohnsonx:
> > As a point of curiosity, why avoid taking common as a dependency? Is there too much business logic in common, and should we create a separate utilities package so that truly common things can just be shared?
> 
> I just want to split `testutils` into two separate modules, once that depends on common and the other that doesn't depend on common. `testutils` was initially created for the specific purpose that we needed something that could consume common, primarily because we needed such behavior for Robolectric tests. Now that it is being used more broadly in UI Automation in native msal, broker as well as in OneAuth.....I think there is always a chance to run into dependency conflicts with different versions of common being pulled. That's why I want to avoid it....also because the dependency of common isn't really needed for UI Automation (at least thus far), so we can have two completely separate testutilities libraries, one for Robolectric tests and one for UI Automation. We're kind of already in that direction with the inception of the `uiautomationutilities` but it internally depends on `testutils` for some things...so we just need to get rid of that I think.

OK, cool. I'm just trying to suss out if there's more work to do.

AdamBJohnsonx (approved):
I saw this before and have no problems with it.

PR #1161, opened 2020-09-22
- fixes #1157 a bug where proguard would import other packages

### LINKED ISSUE #1157: Current proguard rule is keeping way too much library code
**Describe the bug**
This library Proguard file has -**keep class com.microsoft.\*\* \{ \*; \}** which keeps all the files inside that package which is very inefficient.

**Smartphone (please complete the following information):**
 - Device: Not Device specific but proguard rule specific
 - Android Version: NA
 - Browser NA
 - MSAL Version 2.0.0

**To Reproduce**
-  Add the 1. implementation 'com.microsoft.graph:microsoft-graph:2.2.+' and 2. implementation 'com.microsoft.identity.client:msal:1.5.+' 
- Turn on the R8/ProGuard in gradle file, compile the code with the libraries dependencies but without actually using any code from the library.
- Generate the APK(for the variant where proguard is enabled)
- Notice the size(generated APK should be around 3.3 MB)
- Now remove the implementation 'com.microsoft.identity.client:msal:1.5.+' dependency from the gradle file(now new APK size will be 1.6 MB round **51% size** reduction)

**Expected behavior**
When we are not using any code from these libraries then added size due to libs should be close to 0

**Actual Behavior**
It adds around 1.7 MB of size as it's keeping all its file and code under a way too general keep the rule

**Additional context**
While in Android, this library is supposed to be used with [MS-Graph-Java lib](https://github.com/microsoftgraph/msgraph-sdk-java) which clearly states that their library should be used with proper proguard configuration otherwise resulting APK will be bloated with unnecessary code([Link to ReadMe](https://github.com/microsoftgraph/msgraph-sdk-java#13-enable-proguard-android)) while this library doesn't support proper proguard rules and due to wrong(or way too general proguard rule also keeping the files from the Graph library) and hence increasing the size

See the corresponding issue in MS-Graph-Java lib: https://github.com/microsoftgraph/msgraph-sdk-java/issues/462

baywet:
fixes #1157 
Related https://github.com/microsoftgraph/msgraph-sdk-java/issues/462


msal/consumer-rules.pro
```diff
@@ -17,7 +17,7 @@
 #}
 
 ##---------------Begin: proguard configuration for MSAL  --------
--keep class com.microsoft.** { *; }
+-keep class com.microsoft.identity.client.** { *; }
```
rpdome, line 20: I think this should be com.microsoft.identity.**
Is it good enough for your app?

rpdome, line 20: I think we could keep these instead.
 com.microsoft.identity.**
 com.microsoft.device.**
 
would that be good enough for your app?

baywet:
```suggestion
-keep class com.microsoft.identity.** { *; }
```
I have added the device rule in another PR for the common lib
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1047

baywet: would this be enough or do you want the rule to be duplicated here?

AdamBJohnsonx, line 20:
```suggestion
-keep class com.microsoft.identity.** { *; }
```

baywet: applied (other reviewer commented the same)

AdamBJohnsonx (approved):

AdamBJohnsonx:
We'll need to update the changelog for this.

msal/consumer-rules.pro
```diff
@@ -17,7 +17,7 @@
 #}
 
 ##---------------Begin: proguard configuration for MSAL  --------
--keep class com.microsoft.** { *; }
+-keep class com.microsoft.identity.** { *; }
```
baywet, line 20:
```suggestion
-keep class com.microsoft.identity.** { *; }
-keep com.microsoft.device.display.** { *; }
```

baywet:
@AdamBJohnsonx Thanks for the additional comments. I just made the last changes. It should be good to merge and release now.

AdamBJohnsonx (approved):

iambmelt (approved):
LGTM

AdamBJohnsonx:
Since the MSAL PR validation cannot run from a fork, I've merged this into my branch: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1203

baywet:
thanks, I'll close this one and the other one in common once the one you've created gets merged.

PR #1167, opened 2020-09-28
Upgrade msal build for new version of gradle.

AdamBJohnsonx:


msal/build.gradle
```diff
@@ -1,7 +1,10 @@
+plugins {
+    id "org.sonarqube" version "2.6.2"
```
rpdome, line 2: is this being used?

config/checkstyle/checkstyle.xml
```diff
@@ -11,15 +11,15 @@ android checkstyle
...
   <module name="SuppressionFilter">
     <property name="file" value="${checkstyleConfigDir}/suppressions.xml"/>
   </module>
   <property name="severity" value="warning"/>
   <module name="TreeWalker">
-    <module name="FileContentsHolder"/>
     <module name="SuppressWarningsHolder" />
     <property name="tabWidth" value="4"/>
+    <module name="SuppressionCommentFilter" />
+<!--
```
rpdome, line 22: May I ask why are we commenting these out?

rpdome: nvm.. just noticed this is a draft pr.

AdamBJohnsonx: Because they don't pass right now.  My goal is to have a checkstyle that passes and fails the build if it fails.  Then work on removing commented rules.

PR #1169, opened 2020-09-30
Expose expiresIn in MSAL Device Code flow callback

rpdome:
MSTeams would like to display the 'time until expires' on their device.

Related PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1064

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
@@ -31,6 +31,7 @@ import androidx.annotation.WorkerThread;
 import com.microsoft.identity.client.exception.MsalException;
 import com.microsoft.identity.common.internal.controllers.TaskCompletedCallbackWithError;
 
+import java.util.Date;
 import java.util.List;
 
 public interface IPublicClientApplication {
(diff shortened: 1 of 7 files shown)


msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
```diff
@@ -190,8 +190,9 @@ void acquireToken(@NonNull final Activity activity,
          * @param vUri verification uri
          * @param userCode user code
          * @param message instruction message
+         * @param expiresIn the number of seconds before the DCF session expires.
```
AdamBJohnsonx, line 193: This is a really unfortunate value.  I would prefer to see the UTC time that the DCF session expires at, because that signal, while subject to clock skew, would at least not degrade over the time that the system waits before getting around to displaying it.

But if this could be a Duration (number of seconds) or an Instant (precise time point) at least we would help people writing calls to this method sort out their parameters.

rpdome: That's a good idea. Let me do that.

(Note that I did call out that they should be using this for displaying purpose only - not to use it to decide when they should retry - we're already returning a specific exception and they should use that as a signal instead . In that sense - a slight delay should be acceptable. I'll also add a comment to stress on that too.

msal/src/test/java/com/microsoft/identity/client/e2e/shadows/ShadowDeviceCodeFlowCommandSuccessful.java
```diff
@@ -50,11 +51,15 @@
 
     @Implementation
     public AcquireTokenResult execute() {
+        final Date expiryDate = new Date();
+        expiryDate.setTime(expiryDate.getTime() + 900000);
```
shahzaibj, line 55: What are we doing here? and why?

rpdome: 900 seconds is the default timeout.

shahzaibj (approved):

iambmelt (approved):

PR #1174, opened 2020-10-04
Have an opinion about using the broker.  Default to true.

AdamBJohnsonx:


msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -286,7 +286,7 @@ public Boolean getMultipleCloudsSupported() {
      * @return The boolean indicator of whether or not to use the broker.
      */
     public Boolean getUseBroker() {
-        return mUseBroker;
+        return mUseBroker == null ? true : false;
```
shahzaibj, line 289:
```suggestion
        return (mUseBroker == null) || mUseBroker;
```

shahzaibj: Suggested a change above, looks like we are ignoring the supplied value if it is not null

AdamBJohnsonx: You're right.  I wrote this while not paying attention.

shahzaibj (approved):

AdamBJohnsonx:
This is likely a breaking change.  We should have taken this at 2.0.

PR #1177, opened 2020-10-06
Add MSAL Automation Flavors

shahzaibj:
In this PR, I've added broker as a flavor to the MSAL Automation App. This allows us to choose a build variant against which to run automation. So we can run against the following:

- local debug MSAL against Broker Host
- local debug MSAL against debug build of Authenticator app
- dist release MSAL against release build of Authenticator app
- dist release MSAL against release build of Company Portal app

Also did some refactoring as needed. This PR looks big but it is actually small as a lot of similar changes are made in multiple files. Those changes are like:

- Don't override `getBroker` method in every test
- Just use the broker from the automation flavor

Also consumed the changes in this Common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1070


rpdome:
How do you decide which test cases to add "@SupportedBrokers"?

shahzaibj:
> How do you decide which test cases to add "@SupportedBrokers"?

@rpdome If the test case is requiring enrollment, then I am adding that annotation and restricting supported broker to CP.

Also if the test case is requiring Shared Device Registration, then also I am adding that annotation and restricting broker to BrokerHost and Authenticator.

That's pretty much it.....and if the annotation is not provided on the test case, then we assume that all brokers are supported.

rpdome:
> > How do you decide which test cases to add "@SupportedBrokers"?
> 
> @rpdome If the test case is requiring enrollment, then I am adding that annotation and restricting supported broker to CP.
> 
> Also if the test case is requiring Shared Device Registration, then also I am adding that annotation and restricting broker to BrokerHost and Authenticator.
> 
> That's pretty much it.....and if the annotation is not provided on the test case, then we assume that all brokers are supported.

just saw that in the other pr. thanks!

msalautomationapp/build.gradle
```diff
@@ -50,26 +50,60 @@ android {
...
+            dimension "broker"
+        }
+
+        BrokerHost {
+            dimension "broker"
+        }
+    }
+
+    variantFilter { variant ->
+        def flavorName = variant.flavors*.name
```
rpdome, line 90: How does this work? Does it matter what flavor/build the automation test app is built with?

Should we use the signature hash of the APKs instead to see if the combination is valid?

shahzaibj: All we are doing here is excluding specific build variants from gradle because such variants are just not supported under any circumstance. For instance, we don't support using Company Portal with a `localDebug` build of MSAL because MSAL will just not talk to it.

shahzaibj: So we aren't checking here if a specific flavor of MSAL can work with a specific flavor of broker. All we are doing here is just excluding some variants that we don't support in the MSAL Automation App (at least as of today).

rpdome: what I meant is that I can technically upload MSAL/Broker with any build/flavor and pick the 'right' build/flavor for the msalautomationapp, right?

shahzaibj: Broker yes.....MSAL is selected by the build variant (local vs dist).

So yes technically you can select Broker flavor as BrokerHost and instead supply Authenticator app when running the automation. In that case the tests will fail because the tests will be looking for a BrokerHost apk and if they don't find it then they will fail.

AdamBJohnsonx (commented):
Should we have constants or enums for our magic strings?

shahzaibj:
> Should we have constants or enums for our magic strings?

Are you referring to the names of the brokers? Ideally yes enums (or at least constants)....I just don't know if I can also access those in the `build.gradle` when declaring flavors....or if I can do the way around.....get all flavors as enums/constants and access them in JAVA code (without having to re-declare them)

AdamBJohnsonx:
> > Should we have constants or enums for our magic strings?
> 
> Are you referring to the names of the brokers? Ideally yes enums (or at least constants)....I just don't know if I can also access those in the `build.gradle` when declaring flavors....or if I can do the other way around.....get all flavors as enums/constants and access them in JAVA code (without having to re-declare them)

Basically, can we put this in a constants class and then share them between the build and the source.

shahzaibj:
> > > Should we have constants or enums for our magic strings?
> > 
> > 
> > Are you referring to the names of the brokers? Ideally yes enums (or at least constants)....I just don't know if I can also access those in the `build.gradle` when declaring flavors....or if I can do the other way around.....get all flavors as enums/constants and access them in JAVA code (without having to re-declare them)
> 
> Basically, can we put this in a constants class and then share them between the build and the source.

@AdamBJohnsonx Addressed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1177/commits/4e5a5dd0781f77f40256cdbfae2ca9b2ee7898f4

AdamBJohnsonx (approved):

PR #1178, opened 2020-10-06
Update submodule to latest

iambmelt:
~Updates submodule to latest -- should fail, due to unit test in `PublicClientApplicationTest#testSecretKeysAreSet`~

Above blocker reverted by:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1071


tanmaymanolkar1 (approved):

AdamBJohnsonx:
In reality, it's a test using .equals on an array.

PR #1182, opened 2020-10-09
Add changelog enforcement.

AdamBJohnsonx:


.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,24 @@
...
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # Make sure we have a changelog entry
+  changelog:
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      - uses: actions/checkout@v2
+      - uses: dangoslen/changelog-enforcer@v1.4.0
```
rpdome, line 20: Have we got (do we need) an approval for this?

rpdome, line 20: Have we got (do we need) an approval for these two?

rpdome:
Thanks for doing this. once this is checked in, would you mind adding the same thing to common, Broker, ADAL as well?

.github/workflows/pull-request.yml
rpdome: The [OSS guide](https://docs.opensource.microsoft.com/content/using/index.html) seems to be focusing on open source components that are part of the product only. I'd suggest reaching out to them to see if there's anything we have to do in order to use it as part of our build pipeline.

AdamBJohnsonx:
> Thanks for doing this. once this is checked in, would you mind adding the same thing to Broker (and ADAL?) as well?

I did that.  Ideas to reach out to on OSS, though?

.github/workflows/pull-request.yml
AdamBJohnsonx: There doesn't seem to be any guidance right now.  Eventually they'll end up needing to have a path for these, but right now there's nothing.  I'm thinking we use it now, and clone it if we need to.

iambmelt (approved):

PR #1183, opened 2020-10-09
add Coverage Report tasks for Msal

ayushmeghwani:
This Pull Request is related to ITEM - https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1152996

This PR adds the tasks which generates tasks for all buildFlavour combination to generate Coverage Report.

To generate the Coverage report:

Run the task
Then run -> ./gradlew localDebugMsalBothTestCoverageReport (where local(or snapshot or dist) is Build, debug(or release) is Flavour, and Both(or Android or unit) is Test Type)
Reports will be available in msal/msal/build/reports/jacoco


msal/build.gradle
```diff
@@ -124,6 +133,54 @@ task sourcesJar(type: Jar) {
...
+    // When no product flavors defined, use empty string
+    if (!productFlavors) productFlavors.add('')
+    productFlavors.each { productFlavorName ->
+        buildTypes.each { buildTypeName ->
+            def sourceName
+            if (!productFlavorName) {
+                sourceName = "${buildTypeName}"
+            } else {
+                sourceName = "${productFlavorName}${buildTypeName.capitalize()}"
+            }
```
shahzaibj, line 155: see my comments about these in the common PR. same comments apply here

ayushmeghwani: added buildVariants to loop over, have a look.

msal/build.gradle
```diff
@@ -5,6 +5,16 @@ apply plugin: 'maven'
 apply plugin: 'maven-publish'
 apply from: 'versioning/version_tasks.gradle'
 group = 'com.microsoft.identity.client'
+apply plugin: 'jacoco'
+
+jacoco {
+    reportsDir = file("$buildDir/reports/unitTests")
+}
+
+//To include Robolectric tests in the Jacoco report, flag -> "includeNolocationClasses" is set to true
```
shahzaibj, line 14: Can you provide more details as to why this is needed?

msal/build.gradle
```diff
@@ -124,6 +134,38 @@ task sourcesJar(type: Jar) {
     destinationDir = reporting.file("$project.buildDir/outputs/jar/")
 }
 
+task createMsalCoverageReportTask() {
+    group = "Msal Reporting"
+    description = "Generate Coverage Report tasks for every build Flavour"
```
shahzaibj, line 139:
```suggestion
    description = "Generate Coverage Report tasks for every build variant"
```

msal/build.gradle
```diff
@@ -124,6 +134,38 @@ task sourcesJar(type: Jar) {
...
+    android.libraryVariants.all { variant ->
+        def buildVariant = variant.name
+        def testTaskName = "test${buildVariant.capitalize()}UnitTest"
+
+        // Tasks for generating Coverage reports for every buildVariant
```
shahzaibj, line 144:
```suggestion
        // Tasks for generating Coverage reports for every build variant
```

msal/build.gradle
```diff
@@ -124,6 +134,38 @@ task sourcesJar(type: Jar) {
...
+        task "${buildVariant}MsalCoverageReport" (type:JacocoReport, dependsOn: ["$testTaskName", "create${buildVariant.capitalize()}CoverageReport"]) {
```
shahzaibj, line 145: So tell me if my understanding is right - what we're doing here is as follows:

- For each build variant generate a new coverage report task
- This coverage report task will run unit test for the build variant specified and perform code coverage. It will also call the pre-generated jacoco task for capturing code coverage for instrumented tests. (The pre-generated task does both 
 i.e. run instrumented tests and capture coverage)

You are only using the pre-generated code coverage task to avoid having to re-do the work for doing coverage for instrumented tests?

msal/build.gradle
```diff
@@ -124,6 +134,38 @@ task sourcesJar(type: Jar) {
     destinationDir = reporting.file("$project.buildDir/outputs/jar/")
 }
 
+task createMsalCoverageReportTask() {
+    group = "Msal Reporting"
```
shahzaibj, line 138: Let's change this to `Code Coverage`

shahzaibj: Same applies to other repos as well

msal/build.gradle
```diff
@@ -124,6 +134,38 @@ task sourcesJar(type: Jar) {
...
+            description = "Generate Msal Coverage Reports(UnitTest & AndroidTests) on the ${buildVariant.capitalize()}"
+
+            def coverageSourceDirs = "${projectDir}/src/main/java"
+            def fileFilter = ['**/R.class', '**/R$*.class', '**/BuildConfig.*', '**/Manifest*.*', '**/*Test.*']
+            def javaClasses = fileTree(
+                    dir: "${buildDir}/intermediates/javac/${buildVariant}/classes",
+                    excludes: fileFilter
+            )
+            sourceDirectories.from = files([coverageSourceDirs])
+            additionalSourceDirs.from = files(coverageSourceDirs)
```
shahzaibj, line 156: What are additionalSourceDirs?

msal/build.gradle
```diff
@@ -124,6 +134,38 @@ task sourcesJar(type: Jar) {
...
+    android.libraryVariants.all { variant ->
+        def buildVariant = variant.name
+        def testTaskName = "test${buildVariant.capitalize()}UnitTest"
+
+        // Tasks for generating Coverage reports for every buildVariant
+        task "${buildVariant}MsalCoverageReport" (type:JacocoReport, dependsOn: ["$testTaskName", "create${buildVariant.capitalize()}CoverageReport"]) {
+            group = "Msal Reporting"
+            description = "Generate Msal Coverage Reports(UnitTest & AndroidTests) on the ${buildVariant.capitalize()}"
+
+            def coverageSourceDirs = "${projectDir}/src/main/java"
```
shahzaibj, line 149: How is the source code in common dependency being considered? It did not seem to be considered (at least when running from Android-Complete)

msal/build.gradle
shahzaibj: I think the better thing to do (as discussed the other day) is to create a separate task for coverage for unit tests. Then you can have a third task then can just combine the tasks for coverage report for unit tests + instrumented tests. This would allow us to be more flexible.

msal/build.gradle
```diff
@@ -124,6 +134,38 @@ task sourcesJar(type: Jar) {
     destinationDir = reporting.file("$project.buildDir/outputs/jar/")
 }
 
+task createMsalCoverageReportTask() {
```
shahzaibj, line 137: This probably doesn't need to be a task of its own. I don't think we ever need to call this directly, right? 

It can just be a method that's called on the build variants, and just go directly in the `android` section of the file -- IIRC we are already looping over build variants somewhere in this file so you can just follow that pattern.

shahzaibj (changes requested):
Left some questions and suggestion - please also consider this feedback for all the other repos as well

msal/build.gradle
ayushmeghwani: So whenever this flag is not set to true or enabled. there is no coverage which happens for Roboelectric tests and as msal and other repos contains these Roboelectric tests. so we have to enable this class. And in earlier versions of jacoco, this flag was already enabled in jacoco plugin itself but in latest versions it is not there so we have to do it manually.

msal/build.gradle
ayushmeghwani: Yes you are right, So Code coverage task is dependent on 2 tasks:

- the first one runs all the unit tests and and jacoco instruments all the src files and with the help of these two, jacoco creates final coverage report

- And second task is one which does both the task for instrument tests, so there is no need of any addition work on that task.


And as per our discussion that day, as Azure dev ops does not support instrument tests, so in all for every build there will be three tasks:

1. Task which would generate report for both unit tests and instrumented tests
2. Task which would generate report only for unit tests
3. Task which would generate report only for instrumented tests

msal/build.gradle
ayushmeghwani: No that is not the case, the actual path of coverageSourceDirs is "android-complete\msal\msal/src/main/java" , where path of projectDir is "android-complete\msal\msal". So jacoco use all the src classes and instrument it to generate coverage report. so that is the purpose of this coverageSourceDirs .

msal/build.gradle
ayushmeghwani: According to documentation,
sourceDirectories - Source sets that coverage should be reported for.
additionalSourceDirs - Additional source dirs for the classes coverage data is being reported for,
So after investigation, sourceDirectories is only sufficient to cover all the classess for the coverage report, additionalSourceDirs is redundant here

msal/build.gradle
ayushmeghwani: changed.

msal/build.gradle
ayushmeghwani: changed.

msal/build.gradle
ayushmeghwani: changed.

msal/build.gradle
ayushmeghwani: generation of all the new tasks are now implemented in already loop-over sub section of android block of build.gradle file

msal/build.gradle
```diff
@@ -79,6 +89,59 @@ android {
...
+            )
+            sourceDirectories.from = files([coverageSourceDirs])
+            classDirectories.from = files([javaClasses])
+            executionData.from = fileTree(dir: "$buildDir", includes: [
+                    "jacoco/${testTaskName}.exec"
+            ])
+            reports {
+                xml.enabled = true
+                html.enabled = true
+            }
```
shahzaibj, line 115: Do we need to do all this? Can't this task just depend on code coverage task for unit and android test? (I think all this task needs to do is to just run those other two tasks)

msal/build.gradle
```diff
@@ -79,6 +89,59 @@ android {
...
+        }
+
+        // Tasks for generating Coverage reports(AndroidTest) for every build variant
+        task "${buildVariant}MsalAndroidTestCoverageReport" (type:JacocoReport, dependsOn: ["create${buildVariant.capitalize()}CoverageReport"]) {
+            group = "Code Coverage"
+            description = "Generate Msal Coverage Report(AndroidTest) on the ${buildVariant.capitalize()}"
+        }
```
shahzaibj, line 144: Don't we have tasks already (auto generated) for Code Coverage for instrumented tests?

msalautomationapp/build.gradle
```diff
@@ -157,3 +160,17 @@ dependencies {
...
+    description = "Generate AndroidTests Coverage Report tasks for every build Flavour"
+    android.applicationVariants.all { variant ->
+        def buildVariant = variant.name
+
+        // Tasks for generating Coverage reports for every buildVariant
+        task "${buildVariant}MsalAutomationCoverageReport" (type:JacocoReport, dependsOn: ["create${buildVariant.capitalize()}CoverageReport"]) {
+            group = "MsalAutomation Reporting"
+            description = "Generate MsalAutomation AndroidTests Coverage Reports on the ${buildVariant.capitalize()}"
+        }
+    }
```
shahzaibj, line 175: Same feedback as MSAL, this doesn't need to be a task - it can just be looping over variants in the `android` section of `build.gradle`

shahzaibj: In fact, since this is Instrumented Test, I think these should be auto generated?

msal/build.gradle
ayushmeghwani: For unit tests, we have to do all of this. The "testLocalDebugUnitTest" task runs all the test, that's it but to generate coverage report we are doing all of this. including source files and raw output of task "testLocalDebugUnitTest" is stored in "testLocalDebugUnitTest.exec" file that's why we use it.

But for Android tests, there is no need to do anything. just run the auto generated task "createLocalDebugCoverageReport" and it will generate the coverage report, that's why parent task depends on this task so that both reports could be generated with execution of one task only.

msal/build.gradle
ayushmeghwani: common repos are included in coverage report

msal/build.gradle
ayushmeghwani: added one generic method to reduce repetition of code.

msal/build.gradle
ayushmeghwani: Those tasks does not cover all the files for coverage, and misses on showing coverage in file. so added new method and tasks

msalautomationapp/build.gradle
ayushmeghwani: added new changes which create coverages tasks as per the requirement

msal/build.gradle
```diff
@@ -79,6 +85,11 @@ android {
             def fileName = "${archivesBaseName}-${version}.aar"
             outputFileName = fileName
         }
+
+        def buildVariant = variant.name
+        createCoverageTask(buildVariant, "UnitTest")
+        createCoverageTask(buildVariant, "AndroidTest")
+        createCoverageTask(buildVariant, "BothTest")
```
shahzaibj, line 92: The Strings `UnitTest`, `AndroidTest` and `BothTest` can be stored in some constants as they are being reused.

msal/build.gradle
```diff
@@ -96,6 +107,52 @@ android {
     }
 }
 
+ext.createCoverageTask = { buildVariant, testType ->
+    def unitTest = "test${buildVariant.capitalize()}UnitTest"
```
shahzaibj, line 111:
```suggestion
    def unitTestTask = "test${buildVariant.capitalize()}UnitTest"
```

msal/build.gradle
```diff
@@ -96,6 +107,52 @@ android {
...
+    def androidTest = "connected${buildVariant.capitalize()}AndroidTest"
```
shahzaibj, line 112:
```suggestion
    def androidTestTask = "connected${buildVariant.capitalize()}AndroidTest"
```

msal/build.gradle
```diff
@@ -96,6 +107,52 @@ android {
...
+                excludes: fileFilter
+        )
+
+        sourceDirectories.from = files([coverageSourceDirs])
+        additionalSourceDirs.from = files(additionalCoverageSourceDirs)
+        additionalClassDirs.from = files([additionalJavaClasses])
+        classDirectories.from = files([javaClasses])
+        executionData.from = fileTree(dir: "$buildDir", includes: executionFiles)
+        reports {
+            xml.enabled = true
```
shahzaibj, line 150: can we also get html?

msal/build.gradle
```diff
@@ -96,6 +107,52 @@ android {
...
+        executionFiles = [unitTestFile, androidTestFile]
+    } else if (testType.matches("UnitTest")){
+        taskList = [unitTest]
+        executionFiles = [unitTestFile]
+    } else {
+        taskList = [androidTest]
+        executionFiles = [androidTestFile]
+    }
+
+    task "${buildVariant}Msal${testType}CoverageReport" (type:JacocoReport, dependsOn: taskList) {
```
shahzaibj, line 127: Let's add a little documentation for this task

msal/build.gradle
ayushmeghwani: html reports are already being generated, by default. So, I didn't add any flag for it

msal/build.gradle
ayushmeghwani: changed.

msal/build.gradle
ayushmeghwani: changed

msal/build.gradle
ayushmeghwani: added

msal/build.gradle
ayushmeghwani: added

shahzaibj (approved):
LGTM!

Thanks @ayushmeghwani !

PR #1193, opened 2020-10-14
Automating test case 850457

kkrishnareddy:
Automating test case 850457


msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/TestCase850457.java
```diff
@@ -0,0 +1,155 @@
+package com.microsoft.identity.client.msal.automationapp.testpass.broker;
```
shahzaibj, line 1: add license

kkrishnareddy: added license

PR #1214, opened 2020-11-09
Fix asm conflict

shahzaibj:
use asm pulled with Robolectric

This has just started to surface with latest Android Studio....no idea why we didn't encounter this before


AdamBJohnsonx (approved):
Should we specify the version of asm?

shahzaibj:
> Should we specify the version of asm?

We could.....but probably that will just be hard coding whatever Robolectric is using...because iirc robolectric doesn't work well unless using the version it is requesting....so I think I'll just leave as is for now.

rpdome (approved):

PR #1218, opened 2020-11-11
Temporarily ignore all tests in migration tenants

shahzaibj:
The lab team is not actively maintaining these accounts and hence these are failing from time to time....will add back once/if lab team confirms that it is okay to do so.


tanmaymanolkar1 (approved):

tanmaymanolkar1:
would these tests be not considered in the test pass till this change is undone?

shahzaibj:
> would these tests be not considered in the test pass till this change is undone?

They were never part of the test pass.....these were added during BF/FF migration testing...and I just left this there and considered these as bonus validation....but now they are getting unreliable and hence removing...but it has no impact on our release test suites.

PR #1231, opened 2020-11-13
UI automation changes for ci

shahzaibj:
Changes per changes made in the following:

- AzureAD/microsoft-authentication-library-common-for-android#1094
- AzureAD/microsoft-authentication-library-common-for-android#1127


msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/AbstractMsalBrokerTest.java
```diff
@@ -44,20 +42,14 @@
...
-    @Rule(order = 7)
-    public final TestRule powerLiftIncidentRule = new PowerLiftIncidentRule(mBroker);
-
-    @Rule(order = 8)
-    public final TestRule deviceEnrollmentFailureRecoveryRule = new DeviceEnrollmentFailureRecoveryRule();
-
+    @NonNull
     @Override
     public ITestBroker getBroker() {
+        // only initialize once....so calling getBroker from anywhere returns the same instance
```
iambmelt, line 48: Are these parallelized? (Or could they be in the future?) Does this need to be a `synchronized` call?

iambmelt (approved):
Minor question otherwise LGTM thanks @shahzaibj !

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/AbstractMsalBrokerTest.java
shahzaibj: I can make it synchronized....but this is currently only called once per test case (each test case would create new instance of this class)

shahzaibj: sorry that wasn't entirely correct

shahzaibj: It is not parallelized right now......I can make it synchronized....but each test case would create new instance of this class...so it shouldn't really matter

rpdome (approved):

PR #1235, opened 2020-11-20
Just remove the test annotation from ignored tests....don't want these to even be reported on Pipeline for now

shahzaibj:
These are probably going to be deleted anyway....(if the Lab team can't fix outstanding issues)

### DIFF
--- a/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/b2c/B2CIdLabSisoPolicyTest.java
+++ b/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/b2c/B2CIdLabSisoPolicyTest.java
@@ -46,7 +46,6 @@ import org.junit.runners.Parameterized;
 
 import java.util.Arrays;
 
-@RunWith(Parameterized.class)
 public class B2CIdLabSisoPolicyTest extends AbstractB2CTest {
 
     final static B2CProvider[] b2CProviders = new B2CProvider[]{
@@ -72,7 +71,6 @@ public class B2CIdLabSisoPolicyTest extends AbstractB2CTest {
         return mB2cProvider;
     }
 
-    @Test
     public void testCanLoginWithLocalAndSocialAccounts() {
         final TokenRequestLatch latch = new TokenRequestLatch(1);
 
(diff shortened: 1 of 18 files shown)


msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/b2c/B2CIdLabSisoPolicyTest.java
```diff
@@ -46,7 +46,6 @@
 
 import java.util.Arrays;
 
-@RunWith(Parameterized.class)
```
ayushmeghwani, line 49: When There is no test now, using this class we just can keep this annotation as it is.
Though I understand if no test is using it, what's the point of having this annotation and also if this annotation requires any processing where some cost is spent in terms of resources/time. Just a thought

ayushmeghwani (changes requested):

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/b2c/B2CIdLabSisoPolicyTest.java
```diff
@@ -46,7 +46,6 @@
...
 public class B2CIdLabSisoPolicyTest extends AbstractB2CTest {
```
ayushmeghwani, line 49: When There is no test now, using this class we just can keep this annotation as it is.
Though I understand if no test is using it, what's the point of having this annotation and also if this annotation requires any processing where some cost is spent in terms of resources/time. Just a thought

kkrishnareddy (approved):

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/b2c/B2CIdLabSisoPolicyTest.java
shahzaibj: If the annotation is kept then Android Studio think there is a test in this class that is meant to be run.....so to avoid Android Studio being confused, I just removed the annotation.

PR #1237, opened 2020-11-24
Add a new checkstyle file that works with the new version

AdamBJohnsonx:


config/checkstyle/checkstyle.xml
```diff
@@ -116,7 +114,6 @@ android checkstyle
     </module>
   </module>
   <module name="NewlineAtEndOfFile">
-    <property name="severity" value="ignore"/>
```
AdamBJohnsonx, line 119: I want this check.

PR #1263, opened 2020-12-24
add MSAL Sdk Wrapper for Automation Testcases

ayushmeghwani:
MSAL SDK Wrapper for Automation TestCases - which will remove the dependency on latches.


shahzaibj:
My understanding is that this PR will no longer be needed as we've already approved this one: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1264 (which is basically a super set of this). @ayushmeghwani Please confirm.

ayushmeghwani:
> My understanding is that this PR will no longer be needed as we've already approved this one: #1264 (which is basically a super set of this). @ayushmeghwani Please confirm.

Yes this PR is not needed now as all of the changes here are already in the PR #1264 , so once #1264 is merged in I will close this PR. So, yes you are right.

PR #1268, opened 2021-01-06
adding product flavor to support all testcases in msalautomationapp

kkrishnareddy:
Adds a product flavor to automation app that doesn't specify a specific broker to use but rather picks the appropriate broker based on whether we are testing local vs dist as well as based on which brokers are actually supported for the test being executed.


shahzaibj (approved):
LGTM! Thanks @kkrishnareddy!

PR #1270, opened 2021-01-07
Warn if SAPCA used with B2C. Throw Exception if Multiple Policies are Configured

iambmelt:
Impetus:
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1256989

Description:
- SAPCA (SingleAccountPublicClientApplication) should not be used with B2C, unless a single policy exists (and only 1 ever will exist)
- Logs a warning if SAPCA is used with B2C
- Throws an Exception if SAPCA is used with B2C _and_ multiple policies (expressed as `Authorities`) are defined

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -114,6 +114,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
 import static com.microsoft.identity.client.PublicClientApplicationConfigurationFactory.initializeConfiguration;
+import static com.microsoft.identity.client.exception.MsalClientException.SAPCA_USE_WITH_MULTI_POLICY_B2C;
 import static com.microsoft.identity.client.exception.MsalClientException.UNKNOWN_ERROR;
 import static com.microsoft.identity.client.internal.CommandParametersAdapter.createGenerateShrCommandParameters;
 import static com.microsoft.identity.client.internal.MsalUtils.throwOnMainThread;
@@ -939,7 +940,7 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
                                     );
 
                                     if (config.getAuthorities().size() > 1) {
-                                        throw new MsalClientException("SingleAccountPublicClientApplication cannot be used with multiple B2C policies.");
+                                        throw new MsalClientException(SAPCA_USE_WITH_MULTI_POLICY_B2C);
                                     }
                                 }
 
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -903,6 +903,13 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
             config.setRedirectUri(redirectUri);
         }
 
+        try {
+            validateAccountModeConfiguration(config);
+        } catch (final MsalClientException e) {
+            listener.onError(e);
+            return;
+        }
+
         final CommandParameters params = CommandParametersAdapter.createCommandParameters(config, config.getOAuth2TokenCache());
 
         final BaseController controller;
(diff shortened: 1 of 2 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -930,6 +931,19 @@ public void onTaskCompleted(Boolean isSharedDevice) {
...
+                                            TAG,
+                                            "Warning! B2C applications should use MultipleAccountPublicClientApplication. "
+                                                    + "Use of SingleAccount mode with multiple IEF policies is unsupported."
+                                    );
+
+                                    if (config.getAuthorities().size() > 1) {
+                                        throw new MsalClientException(SAPCA_USE_WITH_MULTI_POLICY_B2C);
+                                    }
+                                }
+
```
shahzaibj, line 946: Could potentially do all this check earlier (prior to doing the GetDeviceModeCommand) and fail early if we are destined to fail

shahzaibj (approved):
One minor comment, else LGTM

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
iambmelt: Thanks @shahzaibj -- applied this feedback in c3e5a4bd8b4bd37e0c4a10bd03c8c747021fffdd

iambmelt:
> One minor comment, else LGTM

@shahzaibj Any idea why the `SingleAccountOverloadsMockedTest` would fail during validation of this PR?

shahzaibj:
> > One minor comment, else LGTM
> 
> @shahzaibj Any idea why the `SingleAccountOverloadsMockedTest` would fail during validation of this PR?

@iambmelt Looking into it - seems like some interesting stuff is happening and we end up making a network request when you shouldn't have. This is the error output:

java.lang.AssertionError: AADSTS90002: Tenant 'test.onmicrosoft.com' not found. This may happen if there are no active subscriptions for the tenant. Check to make sure you have the correct tenant ID. Check with your subscription administrator.
Trace ID: 673c2368-0200-4350-bb45-141ca8150f00
Correlation ID: 28236ade-cd12-4d3b-90d1-04529acd95fd
Timestamp: 2021-01-08 00:13:47Z

shahzaibj:
@iambmelt My guess is that the submodule update is causing the problem here due to changes introduced here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/commit/cf22fa1f54bdf49d6572336b7656621bcc51268a

so we probably end up not shadowing the new methods...still looking to confirm this

shahzaibj:
Yeah we're going to need to implement a Robolectric shadow for [HttpClient.java](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/be024d94e1c3e71186b437bd40387fff7361a682/common/src/main/java/com/microsoft/identity/common/internal/net/HttpClient.java#L39)- we don't currently have one so any thing that is not meant to hit network will end up hitting the network because the old shadow is irrelevant now because that one was created for the [HttpRequest.java](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/be024d94e1c3e71186b437bd40387fff7361a682/common/src/main/java/com/microsoft/identity/common/internal/net/HttpRequest.java) and that is not being used anymore.

shahzaibj:
@iambmelt See comments above - if you want to get this PR merged in quickly then you can perhaps avoid the submodule update for now and someone can pick up the item to add robolectric shadow for the new HttpClient and then once that is done then update the submodule here (perhaps a good one for @gitaumoses4 to pick up). We can discuss in scrum tomorrow...or if you want to just do it right now @iambmelt then go for it!

iambmelt:
> @iambmelt See comments above - if you want to get this PR merged in quickly then you can perhaps avoid the submodule update for now and someone can pick up the item to add robolectric shadow for the new HttpClient and then once that is done then update the submodule here (perhaps a good one for @gitaumoses4 to pick up). We can discuss in scrum tomorrow...or if you want to just do it right now @iambmelt then go for it!

Thanks @shahzaibj for the detailed notes. I think for sake of expediency, I'll forego the submodule update as part of this PR and we can coordinate with @gitaumoses4 to see that the requisite shadow is written in order for us to successfully ingest his work

PR #1286, opened 2021-01-22
Automation: Initialize activity before loading lab user

shahzaibj:
Temp user creation takes time and Firebase is not happy blocking activity launch while it waits for that and throws the below error. This PR fixes that issue.

Error:

java.lang.RuntimeException: Could not launch intent Intent { act=android.intent.action.MAIN flg=0x10000000 cmp=com.msft.identity.client.sample.local/com.microsoft.identity.client.msal.automationapp.MainActivity } within 45 seconds. Perhaps the main thread has not gone idle within a reasonable amount of time? There could be an animation or something constantly repainting the screen. Or the activity is doing network calls on creation? See the threaddump logs. For your reference the last time the event queue was idle before your activity launch request was 1611302798256 and now the last time the queue went idle was: 1611302843262. If these numbers are the same your activity might be hogging the event queue.


gitaumoses4 (approved):
Looks good.

PR #1292, opened 2021-01-27
Don't throw run time exception when resetting logger, allow resetting logger

shahzaibj:


msal/src/main/java/com/microsoft/identity/client/Logger.java
```diff
@@ -138,10 +138,6 @@ public synchronized void setExternalLogger(final ILoggerCallback externalLogger)
             return;
         }
 
-        if (null != mExternalLogger) {
-            throw new IllegalStateException("External logger is already set, cannot be set again.");
-        }
-
```
AdamBJohnsonx, line 144: Why do we have a reference to this mExternalLogger if it isn't used anywhere but inside this function?

msal/src/main/java/com/microsoft/identity/client/Logger.java
```diff
@@ -138,10 +138,6 @@ public synchronized void setExternalLogger(final ILoggerCallback externalLogger)
...
         // If mExternalLogger is not set. Then implement the ILoggerCallback interface in common-core.
```
AdamBJohnsonx, line 141: Probably this comment is now incorrect.

msal/src/main/java/com/microsoft/identity/client/Logger.java
shahzaibj: yeah no idea....I can remove it

msal/src/main/java/com/microsoft/identity/client/Logger.java
```diff
@@ -138,10 +138,6 @@ public synchronized void setExternalLogger(final ILoggerCallback externalLogger)
...
         final com.microsoft.identity.common.internal.logging.Logger logger =
```
AdamBJohnsonx, line 142: It looks like we have a synchronized keyword on this method... I'm not clear why.  The synchronization needs to happen inside *this* class, in common, around the reference to mExternalLogger.  And synchronization needs to happen both on read and on write.

msal/src/main/java/com/microsoft/identity/client/Logger.java
shahzaibj: Addressed

msal/src/main/java/com/microsoft/identity/client/Logger.java
shahzaibj: Addressed

msal/src/main/java/com/microsoft/identity/client/Logger.java
AdamBJohnsonx: The synchronized keyword on this method is not actually protecting anything.  You should kill it.  The anonymous construction should be thread-safe (every construction happens-before its assignment), and the references inside of that construction are final.  This synchronization says nothing about one invocation of this method relative to another, so it's not actually accomplishing anything.

iamgusain:
Curious, When and why do we need to reset the logger?

msal/src/main/java/com/microsoft/identity/client/Logger.java
```diff
@@ -63,8 +63,6 @@
 public final class Logger {
```
iamgusain, line 63: >Logger [](start = 19, length = 6)

should this class need to have a private constructor to prevent anybody creating an instance of Logger? 

private Logger() {};

msal/src/main/java/com/microsoft/identity/client/Logger.java
```diff
@@ -127,22 +125,16 @@ public void setLogLevel(final LogLevel logLevel) {
...
     /**
      * Set the custom logger. Configures external logging to configure a callback that
-     * the sdk will use to pass each log message. Overriding the logger callback is not allowed.
+     * the sdk will use to pass each log message.
      *
      * @param externalLogger The reference to the {@link ILoggerCallback} that can
      *                       output the logs to the designated places.
-     * @throws IllegalStateException if external logger is already set, and the caller is trying to set it again.
      */
     public synchronized void setExternalLogger(final ILoggerCallback externalLogger) {
```
iamgusain, line 133: >setExternalLogger [](start = 29, length = 17)

Just curious: is this a way to push logs to additional targets than logcat?

AdamBJohnsonx (approved):
My minor comment about getting rid of the synchronized keyword can be addressed at your leisure.

iambmelt:
Can we put a description in this PR to clarify why this change is needed?

PR #1305, opened 2021-02-16
Update test configuration to support java 8 bytescode in tests

AdamBJohnsonx:


rpdome (approved):

msal/build.gradle
```diff
@@ -23,6 +23,22 @@ def bothTestType = "BothTest"
...
+            // Sets Java compatibility to Java 8
+            sourceCompatibility JavaVersion.VERSION_1_8
+            targetCompatibility JavaVersion.VERSION_1_8
+        }
+        dependencies {
+            coreLibraryDesugaring "com.android.tools:desugar_jdk_libs:$rootProject.ext.coreLibraryDesugaringVersion"
+        }
+        defaultConfig {
+            multiDexEnabled true
+        }
```
shahzaibj, line 39: I see this got moved....I believe this is also only needed for test?

msal/build.gradle
```diff
@@ -37,7 +53,6 @@ android {
     compileSdkVersion rootProject.ext.compileSdkVersion
 
     defaultConfig {
```
AdamBJohnsonx, line 55:
```suggestion
    defaultConfig {
        multiDexEnabled true
```

msal/build.gradle
AdamBJohnsonx, line 39:
```suggestion
```

AdamBJohnsonx: I put it back.  It really doesn't matter for our libraries, since they don't produce APKs, but we might have a transitive dependency somewhere.

shahzaibj (approved):

PR #1309, opened 2021-02-22
Consume custom test runner created in uiautomationutilities

shahzaibj:
Related common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1242


AdamBJohnsonx (approved):
I think this is just "Consume" instead of "Consumer".

shahzaibj:
> I think this is just "Consume" instead of "Consumer".

yeah just a typo :D

PR #1319, opened 2021-03-01
Disable java 1.8

rpdome:
For some reason, this is also affect the common library itself - the calling app is still asking for JAVA 1.8.

For now, I'll disable this to unblock the release. Once we have capacity, we could revisit this to have this re-enabled for test projects.

related PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1252

### DIFF
--- a/msal/build.gradle
+++ b/msal/build.gradle
@@ -23,18 +23,18 @@ def bothTestType = "BothTest"
 
 android {
 
-    testOptions {
-        compileOptions {
-            // Flag to enable support for the new language APIs
-            coreLibraryDesugaringEnabled true
-            // Sets Java compatibility to Java 8
-            sourceCompatibility JavaVersion.VERSION_1_8
-            targetCompatibility JavaVersion.VERSION_1_8
-        }
-        dependencies {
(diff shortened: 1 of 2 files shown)


msal/build.gradle
```diff
@@ -23,18 +23,18 @@ def bothTestType = "BothTest"
 
 android {
 
-    testOptions {
-        compileOptions {
-            // Flag to enable support for the new language APIs
-            coreLibraryDesugaringEnabled true
-            // Sets Java compatibility to Java 8
-            sourceCompatibility JavaVersion.VERSION_1_8
-            targetCompatibility JavaVersion.VERSION_1_8
```
AdamBJohnsonx, line 32: Can we explicitly set these to 1_7 and false for the compile options?

AdamBJohnsonx (approved):

msal/build.gradle
rpdome: Same answer as the other one.

I'm inclined not to test this in the last minute (MSAL release is already overdue). Commenting out is safer.

I created a [work item](https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20Android/Backlog%20items/?workitem=1313685) and we can spend time to test this out thoroughly in dev.

PR #1324, opened 2021-03-06
Do not allow "" to pass PublicClientApplicationConfiguration validation

iambmelt:
It is possible for `""` to be passed in as part of `PublicClientApplicationConfiguration` and it will not be caught during validation, since only `null` is checked


changelog
```diff
@@ -1,7 +1,8 @@
 MSAL Wiki : https://github.com/AzureAD/microsoft-authentication-library-for-android/wiki
 
 vNext
-- [MINOR] Changes to Broker Validation to allow setting whether to trust debug brokers (prod brokers are always trusted).
+- [MINOR] Changes to Broker Validation to allow setting whether to trust debug brokers (prod brokers are always trusted) (#1274)
+- [PATCH] Better PublicClientApplicationConfiguration validation to fail when empty string is passed
```
iambmelt, line 5:
```suggestion
- [PATCH] Better PublicClientApplicationConfiguration validation to fail when empty string is passed (#1324)
```

AdamBJohnsonx:
Could you add a test for this?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

tanmaymanolkar1 (approved):

msal/src/androidTest/java/com/microsoft/identity/client/PublicClientApplicationTest.java
```diff
@@ -106,6 +106,29 @@ public void tearDown() {
...
+    public void testConfigValidationFailsOnEmptyClientId() throws MsalException, InterruptedException {
+        final Context context = new PublicClientApplicationTest.MockContext(mAppContext);
+        mockPackageManagerWithDefaultFlag(context);
+        mockHasCustomTabRedirect(context);
+
+        final IMultipleAccountPublicClientApplication app = PublicClientApplication.createMultipleAccountPublicClientApplication(
+                context,
+                R.raw.test_pcaconfig_empty_clientid
+        );
+    }
```
shahzaibj, line 131: Since we are writing new tests, can we just use Robolectric and avoid mocks?

iambmelt: I'm really not looking to take on a bunch of test refactor work to get these changes in, if this is like 5 minutes of work OK - what needs to happen to convert these tests?

shahzaibj: They would need to be written in a new class and add the annotation for using [RobolectricTestRunner](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/AcquireTokenNetworkTest.java#L54-L55). After that you just call the API that you want to test - no need to mock package manager or anything. The tricky part would be to pass resource id string, when I first wrote these tests I was having trouble with passing resource id so I did [this](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/test/java/com/microsoft/identity/client/e2e/tests/PublicClientApplicationAbstractTest.java#L64-L71). I think generally writing Robolectric tests is faster than writing other tests, but it is totally up to you if you want to do this refactor or not.

iambmelt: Thanks @shahzaibj -- I took a quick pass at this, looks like a nontrivial amount of work to refactor these tests and rip out the mocking. Given this test is primarily about loading resource id based configs, I think I'm going to decline this for now and build on the existing tests we have for validating the configuration. I'm happy to take a backlog improvement item for these tests, this is just bit more than I want to bite off for this otherwise small PR

shahzaibj: Sounds good, @iambmelt 👍

shahzaibj (approved):

Issue #1314, opened 2021-02-24
Fatal Exception: java.lang.NullPointerException


pedrofarnetaniabi:
**Describe the bug**
We're having this exception reported by crashlytics since MSAL version 2.0.4 (currently we are using MSAL version 2.0.5 that also has this problem).

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy A30s (and many others including other brands)
 - Android Version: 10
 - MSAL Version: 2.0.4 and 2.0.5

**Stacktrace**
> 
       Fatal Exception: java.lang.NullPointerException: uriString
       at android.net.Uri$StringUri.<init>(Uri.java:496)
       at android.net.Uri$StringUri.<init>(Uri.java:486)
       at android.net.Uri.parse(Uri.java:458)
       at com.microsoft.identity.client.internal.MsalUtils.hasCustomTabRedirectActivity(MsalUtils.java:249)
       at com.microsoft.identity.client.PublicClientApplicationConfiguration.checkIntentFilterAddedToAppManifestForBrokerFlow(PublicClientApplicationConfiguration.java:522)
       at com.microsoft.identity.client.PublicClientApplication.initializeApplication(PublicClientApplication.java:1046)
       at com.microsoft.identity.client.PublicClientApplication.<init>(PublicClientApplication.java:1029)
       at com.microsoft.identity.client.MultipleAccountPublicClientApplication.<init>(MultipleAccountPublicClientApplication.java:64)
       at com.microsoft.identity.client.PublicClientApplication$9.onTaskCompleted(PublicClientApplication.java:928)
       at com.microsoft.identity.client.PublicClientApplication$9.onTaskCompleted(PublicClientApplication.java:914)
       at com.microsoft.identity.common.internal.controllers.CommandDispatcher.commandCallbackOnTaskCompleted(CommandDispatcher.java:346)
       at com.microsoft.identity.common.internal.controllers.CommandDispatcher.access$1200(CommandDispatcher.java:74)
       at com.microsoft.identity.common.internal.controllers.CommandDispatcher$3.run(CommandDispatcher.java:325)
       at android.os.Handler.handleCallback(Handler.java:883)
       at android.os.Handler.dispatchMessage(Handler.java:100)
       at android.os.Looper.loop(Looper.java:214)
       at android.app.ActivityThread.main(ActivityThread.java:7397)
       at java.lang.reflect.Method.invoke(Method.java)
[...]

PR #1327, opened 2021-03-09
Add additional validation to redirect uri

iambmelt:
Apropos of #1314 -- this exception _should_ have been caught when [performing validation here](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfigurationFactory.java#L91), however from communicating with the customer, they are passing in configuration that has been transformed/serialized from a remotely-loaded config document controlling various aspects of their app init. As a quirk of this implementation, when passing `null` through their JSON parser the value was converted to a String `"null"` and subsequently bypassed this validation, before throwing a `NullPointerException` on _certain_ devices when invoking `Uri.parse`.

Summary of changes:
- Add check for `"null"` when validating manifest inputs
- Adds license info to other files
- ~Adds tests for this code path via Robolectric~
    * Removed, ended up making a smaller change than originally suggested


msal/src/test/java/android/text/TextUtils.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Workaround for:
+// java.lang.RuntimeException: Method isEmpty in android.text.TextUtils not mocked.
+// See http://g.co/androidstudio/not-mocked for details.
+package android.text;
+
+public class TextUtils {
```
rpdome, line 22: Isn't this available in android?

rpdome (approved):

msal/src/test/java/com/microsoft/identity/client/PublicClientApplicationConfigurationTest.java
```diff
@@ -1,19 +1,78 @@
...
+    @Test
+    @Ignore // Ignore test due to mocking gaps http://g.co/androidstudio/not-mocked
+    public void testValidRedirect() {
+        final PublicClientApplicationConfiguration config = new PublicClientApplicationConfiguration();
+        config.setRedirectUri("msauth://authority");
+        config.validateConfiguration();
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    @Ignore // Ignore test due to mocking gaps http://g.co/androidstudio/not-mocked
```
AdamBJohnsonx, line 72: This seems... unfortunate.  There should probably be a backlog item for a new approach referenced here.

msal/src/test/java/android/text/TextUtils.java
rpdome: ah... Alternatively, we could use our own StringUtil.

AdamBJohnsonx: Which is a little different.  our isEmtpy is a non-performant version of isBlank(), since we trim whitespace.  We should change it to use a static pattern.

msal/src/test/java/com/microsoft/identity/client/PublicClientApplicationConfigurationTest.java
iambmelt: Created:
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1323372

PR #1334, opened 2021-03-11
Changes to MSAL/MSAL test app to support passing extra parameters

AdamBJohnsonx:
I was looking for a way to trigger the PRTv3 flow inside the broker, and I thought it might be easiest just to add a field exposing new options.  Additionally, I did not know of a way to add authorize query strings in the test app, so I added those as well.

I considered using that field for what I was doing, but thought that the confusion that would result from pushing things in that field and then removing them before passing them to the authorize endpoint would be bad.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
+++ b/msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
@@ -42,7 +42,6 @@ import java.util.UUID;
  */
 public abstract class TokenParameters {
 
-    private List<Pair<String, String>> mExtraQueryStringParameters;
     private List<String> mScopes;
     private IAccount mAccount;
     private String mAuthority;
@@ -59,7 +58,6 @@ public abstract class TokenParameters {
         mScopes = builder.mScopes;
         mAuthenticationScheme = builder.mAuthenticationScheme;
         mCorrelationId = builder.mCorrelationId;
-        mExtraQueryStringParameters = builder.mExtraQueryStringParameters;
         mExtraOptions = builder.mExtraOptions;
     }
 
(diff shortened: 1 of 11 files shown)


testapps/testapp/build.gradle
```diff
@@ -94,6 +94,8 @@ android {
 }
 
 dependencies {
+    //implementation 'org.projectlombok:lombok'
+
```
AdamBJohnsonx, line 98:
```suggestion
```

msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
```diff
@@ -185,6 +211,18 @@ public B withScopes(List<String> scopes) {
             return self();
         }
 
+        public B withExtraOptions(List<Pair<String, String>> options) {
+            mExtraOptions = options == null ? options : new ArrayList<Pair<String, String>>(options);
+            return self();
+        }
+
+        public B withAuthorizationQueryStringParameters(
```
shahzaibj, line 219: Looks like this is now base class `TokenParameters` which means it will also be on the `SilentTokenParameters` - in that case this name that contains `authorization` in it would seem strange to be a parameter for a silent token request.

msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
```diff
@@ -154,13 +159,34 @@ public String getCorrelationId() {
...
+     * Otherwise... would recommend not touching.
+     *
+     * @return
+     */
+    public List<Pair<String, String>> getExtraQueryStringParameters() {
+        return mExtraQueryStringParameters;
+    }
+
+    /**
+     * If you have been instructed that the client requires extra parameters, supply them here.
```
shahzaibj, line 173: Can we use the same word (parameters vs options)? Also can you add some examples of what might these extra parameters be?

msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
AdamBJohnsonx: I can shift it back up, honestly.  It just gets dropped off of requests if it isn't there - it was a legacy of me thinking I might use it for triggering, and then I thought better of it and wasn't sure whether we should move it back.

msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
AdamBJohnsonx: Yes.  Probably options.

In this case, the only magic parameter that we will be dealing with would be triggers for special or potentially experimental client side behavior.

PR #1337, opened 2021-03-17
Migrate classic pipeline to YAML

p3dr0rv:
- Moved all azure DevOps classic pipelines in IDDP to YAMLs
- Add steps templates to common for: 
  - CredScan & Policheck
  - Install Automation Cert
  - gcloud
  -  Upload to nexus sonatype
  - GPG Sigining
  - Generate release artifacts


azure-pipelines/experiment/msal-code-coverage-experiment-clone.yml
```diff
@@ -0,0 +1,59 @@
...
+# Variable 'toEmailAccount' was defined in the Variables tab
+name: $(date:yyyyMMdd)$(rev:.r)
+resources:
+  repositories:
+  - repository: self
+    type: git
+    ref: aymeghwa-CodeCoverage
+  - repository: common
+    type: github
+    name: AzureAD/microsoft-authentication-library-common-for-android
```
shahzaibj, line 20: weird....is this a pipeline for the common repo? Seems like it is, then why is it being checked into MSAL repo?

shahzaibj (changes requested):
May be I'm missing something here....but the repository looks wrong for all of these....should be MSAL but the yml files seem to indicate common

azure-pipelines/experiment/msal-code-coverage-experiment-clone.yml
shahzaibj: Actually seems it is meant to be for MSAL....but the repository block seems wrong

p3dr0rv: The pipeline is for the msal repo, but it uses templates that exists in the common repo

shoatman: MSAL has a submodule for common; however I think in this case you want the latest files in common is that correct?

shoatman (changes requested):
Pedro - Are these all the pipelines in one PR?  It might be simpler to do a few at a time.  Also i think it may be eaiser to schedule a meeting to review these together as way to expedite the review process.  @shahzaibj - Thoughts?

PR #1339, opened 2021-03-18
Add correlation ID to MSAL test app

rpdome:
Cut a test build for eSTS team to test something out. Thought it would be useful for us in general also.


shahzaibj (approved):
LGTM!

PR #1346, opened 2021-03-24
Revert not passing environment in removeAccount call for backward compat

iamgusain:
We need to revert the change to pass environment to null, to remove account api for backward compat.
As older version of broker will reject request if environment is set to null in remove account bundler, causing the remove account scenario to break.

Instead we will locally override the environment value to null in local Msal Controller (for msal only scenario) and in broker (for broker scenario) [PR 1287](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1287)


rpdome (approved):

shahzaibj (approved):

shahzaibj:
let's change to RC2 as well

iamgusain:
> let's change to RC2 as well

I don't see a RC2 branch created yet.

shahzaibj:
> > let's change to RC2 as well
> 
> I don't see a RC2 branch created yet.

Correct. I usually prefer having one release branches instead of having separate branches for each RC. You can work with @paulkagiri to update the version number appropriately.

PR #1355, opened 2021-04-05
Nicer names for cacherecord

AdamBJohnsonx:
This started out smaller, put updating the common submodule caused a truly strange error.  In the course of
debugging, I've made the failure messages coming from some of our tests much more verbose, so that I can find
the errors without needing to use a debugger.

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -39,7 +39,7 @@ ext {
     powerMockVersion = "2.0.9"
     runnerVersion = "1.2.0"
     rulesVersion = "1.2.0"
-    robolectricVersion = "4.5.1"
+    robolectricVersion = "4.3.1"
     uiAutomatorVersion = "2.2.0"
 
     // TODO: adal automation test app.
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -36,10 +36,10 @@ ext {
     mockitoAndroidVersion = "3.6.28"
     multidexVersion = "2.0.1"
     nimbusVersion = "8.2"
-    powerMockVersion = "1.6.6"
+    powerMockVersion = "2.0.9"
     runnerVersion = "1.2.0"
     rulesVersion = "1.2.0"
-    robolectricVersion = "4.3.1"
+    robolectricVersion = "4.5.1"
     uiAutomatorVersion = "2.2.0"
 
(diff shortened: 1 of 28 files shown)


iambmelt (approved):
LGTM

msal/build.gradle
```diff
@@ -143,7 +142,9 @@ android {
                 // example: ./gradlew testLocalDebugUnitTest -Plabtest
                 exclude 'com/microsoft/identity/client/e2e/tests/network'
             }
-
+            testLogging {
+                exceptionFormat = "full"
```
iambmelt, line 146: 👍

gradle/versions.gradle
```diff
@@ -36,7 +36,7 @@ ext {
     mockitoAndroidVersion = "3.6.28"
     multidexVersion = "2.0.1"
     nimbusVersion = "8.2"
-    powerMockVersion = "1.6.6"
+    powerMockVersion = "2.0.9"
```
shahzaibj, line 39: Can we update in other places such as android-complete as well?

iambmelt (approved):

rpdome (approved):

msal/src/test/java/com/microsoft/identity/client/e2e/tests/mocked/SingleAccountAcquireTokenMockTest.java
```diff
@@ -66,7 +66,8 @@ public void onAccountChanged(@Nullable IAccount priorAccount, @Nullable IAccount
 
             @Override
             public void onError(@NonNull MsalException exception) {
-                fail("No current account found.");
+                fail("Interpreting an exception as no accounts found.  Unexpected exception: " + exception.getMessage() +
```
shahzaibj, line 69: could also do `throw new AssertionError(exception)` and then we won't have to log the exception message and stack trace ourselves.

gradle/versions.gradle
AdamBJohnsonx: I think android-complete is updated.

AdamBJohnsonx: nope/

shahzaibj (approved):
Just some feedback around changing `fail` and assert statements to `throw new AssertionError(exception)`

PR #1357, opened 2021-04-13
Log and throw exception if there are other apps listening for the url…

shahzaibj:
… scheme

Why making these changes?  -  Read email thread titled "MSAL authentication issue"

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
@@ -552,7 +552,8 @@ public class PublicClientApplicationConfiguration {
             );
 
             if (!hasCustomTabRedirectActivity) {
-                final Uri redirectUri = Uri.parse(mRedirectUri);
+                final Uri redirectUri = Uri.parse(mRedirectUri); 
+                
                 throw new MsalClientException(
                         APP_MANIFEST_VALIDATION_ERROR,
                         "Intent filter for: " +
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
@@ -543,32 +543,34 @@ public class PublicClientApplicationConfiguration {
 
     @SuppressWarnings("PMD")
     public void checkIntentFilterAddedToAppManifestForBrokerFlow() throws MsalClientException {
-        final boolean hasCustomTabRedirectActivity = MsalUtils.hasCustomTabRedirectActivity(
-                mAppContext,
-                mRedirectUri
-        );
-
         if ((getAuthorizationAgent() == AuthorizationAgent.DEFAULT
-                || getAuthorizationAgent() == AuthorizationAgent.BROWSER)
-                && !hasCustomTabRedirectActivity) {
-            final Uri redirectUri = Uri.parse(mRedirectUri);
-            throw new MsalClientException(
-                    APP_MANIFEST_VALIDATION_ERROR,
-                    "Intent filter for: " +
-                            BrowserTabActivity.class.getSimpleName() +
-                            " is missing. " +
-                            " Please make sure you have the following activity in your AndroidManifest.xml \n\n" +
-                            "<activity android:name=\"com.microsoft.identity.client.BrowserTabActivity\">" + "\n" +
-                            "\t" + "<intent-filter>" + "\n" +
-                            "\t\t" + "<action android:name=\"android.intent.action.VIEW\" />" + "\n" +
-                            "\t\t" + "<category android:name=\"android.intent.category.DEFAULT\" />" + "\n" +
-                            "\t\t" + "<category android:name=\"android.intent.category.BROWSABLE\" />" + "\n" +
-                            "\t\t" + "<data" + "\n" +
-                            "\t\t\t" + "android:host=\"" + redirectUri.getHost() + "\"" + "\n" +
-                            "\t\t\t" + "android:path=\"" + redirectUri.getPath() + "\"" + "\n" +
-                            "\t\t\t" + "android:scheme=\"" + redirectUri.getScheme() + "\" />" + "\n" +
-                            "\t" + "</intent-filter>" + "\n" +
-                            "</activity>" + "\n");
+                || getAuthorizationAgent() == AuthorizationAgent.BROWSER)) {
+
+            final boolean hasCustomTabRedirectActivity = MsalUtils.validateCustomTabRedirectActivity(
+                    mAppContext,
+                    mRedirectUri
+            );
+
+            if (!hasCustomTabRedirectActivity) {
+                final Uri redirectUri = Uri.parse(mRedirectUri);
+                throw new MsalClientException(
+                        APP_MANIFEST_VALIDATION_ERROR,
+                        "Intent filter for: " +
+                                BrowserTabActivity.class.getSimpleName() +
+                                " is missing. " +
+                                " Please make sure you have the following activity in your AndroidManifest.xml \n\n" +
+                                "<activity android:name=\"com.microsoft.identity.client.BrowserTabActivity\">" + "\n" +
+                                "\t" + "<intent-filter>" + "\n" +
+                                "\t\t" + "<action android:name=\"android.intent.action.VIEW\" />" + "\n" +
+                                "\t\t" + "<category android:name=\"android.intent.category.DEFAULT\" />" + "\n" +
+                                "\t\t" + "<category android:name=\"android.intent.category.BROWSABLE\" />" + "\n" +
+                                "\t\t" + "<data" + "\n" +
+                                "\t\t\t" + "android:host=\"" + redirectUri.getHost() + "\"" + "\n" +
+                                "\t\t\t" + "android:path=\"" + redirectUri.getPath() + "\"" + "\n" +
+                                "\t\t\t" + "android:scheme=\"" + redirectUri.getScheme() + "\" />" + "\n" +
+                                "\t" + "</intent-filter>" + "\n" +
+                                "</activity>" + "\n");
+            }
         }
 
         if (!mUseBroker) {
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
@@ -398,7 +398,7 @@ public class PublicClientApplicationConfiguration {
     }
 
     private void checkManifestPermissions() {
-        if (this.handleNullTaskAffinity) {
+        if (this.handleNullTaskAffinity != null && this.handleNullTaskAffinity) {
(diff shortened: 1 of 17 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
```diff
@@ -259,12 +280,19 @@ public static boolean hasCustomTabRedirectActivity(@NonNull final Context contex
...
+            if (activityInfo.name.equals(BrowserTabActivity.class.getName()) &&
+                    activityInfo.packageName.equals(context.getPackageName())) {
                 hasActivity = true;
             } else {
                 // another application is listening for this url scheme, don't open
                 // Custom Tab for security reasons
-                return false;
+                Logger.warn(TAG, "Another application is listening for the URL scheme. " +
+                        "This app's package name is " + activityInfo.packageName);
+                throw new MsalClientException(
```
iamgusain, line 291: We don't seem to do anything with this exception, so I don't see much value in throwing it.

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
```diff
@@ -259,12 +280,19 @@ public static boolean hasCustomTabRedirectActivity(@NonNull final Context contex
...
 
-            if (activityInfo.name.equals(BrowserTabActivity.class.getName())) {
+            if (activityInfo.name.equals(BrowserTabActivity.class.getName()) &&
+                    activityInfo.packageName.equals(context.getPackageName())) {
                 hasActivity = true;
             } else {
                 // another application is listening for this url scheme, don't open
                 // Custom Tab for security reasons
-                return false;
+                Logger.warn(TAG, "Another application is listening for the URL scheme. " +
```
iamgusain, line 289: nit: Another application {appName} is listening for the url Scheme

iamgusain, line 289: add methodName to the TAG ?

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
shahzaibj: That's right...we are throwing it...all the way to the caller

shahzaibj: The caller will receive it in `onError` callback

iamgusain: This suggests otherwise... line 243
try {
            return validateCustomTabRedirectActivity(context, url);
        } catch (final MsalClientException e) {
            return false;
        }

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
```diff
@@ -234,8 +237,26 @@ public static int getExpiryOrDefault(String expiresIn) {
      * @param url
      * @return
      */
+    @Deprecated
     public static boolean hasCustomTabRedirectActivity(@NonNull final Context context,
                                                        @NonNull final String url) {
+        try {
+            return validateCustomTabRedirectActivity(context, url);
+        } catch (final MsalClientException e) {
+            return false;
```
tanmaymanolkar1, line 246: I see that, based on the boolean you return the exception, a bit interesting to not pass on the exact exception?

tanmaymanolkar1 (approved):
LGTM, some comments already made here, that should cover up.

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
shahzaibj: This is the API I deprecated...I don't want to modify the behavior of the existing API as it is Public API...or we could choose to do it...I think this wasn't meant to be called by customers so not sure why it is Public...but now that it was public..I chose not to change the behavior

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
shahzaibj: That is the API I deprecated (`hasCustomTabRedirectActivity` vs `validateCustomTabRedirectActivity`) ...I don't want to modify the behavior of the existing API as it is Public API...or we could choose to do it...I think this wasn't meant to be called by customers so not sure why it is Public...but now that it was public..I chose not to change the behavior

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
```diff
@@ -234,8 +237,26 @@ public static int getExpiryOrDefault(String expiresIn) {
...
+
+    /**
+     * validateCustomTabRedirectActivity - Ensures that the developer has properly configured their
+     * AndroidManifest to expose the BrowserTabActivity.
+     *
+     * @param context
+     * @param url
+     * @return
+     */
+    public static boolean validateCustomTabRedirectActivity(@NonNull final Context context,
```
AdamBJohnsonx, line 258: Does this one need to be public?

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
```diff
@@ -234,8 +237,26 @@ public static int getExpiryOrDefault(String expiresIn) {
...
+            return false;
+        }
+    }
+
+    /**
+     * validateCustomTabRedirectActivity - Ensures that the developer has properly configured their
+     * AndroidManifest to expose the BrowserTabActivity.
+     *
+     * @param context
+     * @param url
```
AdamBJohnsonx, line 255: nit: javadoc, or if it isn't meant to be public, no javadoc.

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
AdamBJohnsonx: what's the scheme?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -543,32 +543,34 @@ private void verifyRedirectUriWithAppSignature() throws MsalClientException {
...
+                || getAuthorizationAgent() == AuthorizationAgent.BROWSER)) {
+
+            final boolean hasCustomTabRedirectActivity = MsalUtils.validateCustomTabRedirectActivity(
+                    mAppContext,
+                    mRedirectUri
+            );
+
+            if (!hasCustomTabRedirectActivity) {
+                final Uri redirectUri = Uri.parse(mRedirectUri);
+                throw new MsalClientException(
```
shahzaibj, line 556:
```suggestion
                final Uri redirectUri = Uri.parse(mRedirectUri); 
                
                throw new MsalClientException(
```

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
shahzaibj: I can move it to the class and make it private but then we need to figure out something for this other method that I deprecated: https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/72ed05116f398f0d763544410d88d419b0040aa3/msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java#L241

Can choose to just delete it and increment a major version. Thoughts?

AdamBJohnsonx: The deprecated one is fine as is.  We need a task to remove it in the next MV (in the fall).

shahzaibj: So basically I modified the deprecated API to call my new method (to avoid code duplication). So if I move my new method to some other class and make it private then I won't be able to access it here inside the deprecated method. One thing I can do is just duplicate the code in both the classes for now until the next MV.

AdamBJohnsonx: Leaving the code duplicated seems like the best path, otherwise we create another public API.

shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1357/commits/91fe570fc177b9cc298d2d85147cebff4f98344d

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
shahzaibj: Addressed

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
shahzaibj: I changed it back to the original code in the deprecated method and opted to have code duplication instead so we can easily remove it in the future. Addressed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1357/commits/91fe570fc177b9cc298d2d85147cebff4f98344d

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
shahzaibj: I changed it back to the original code in the deprecated method and opted to have code duplication instead so we can easily remove it in the future. Addressed here: 91fe570

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -393,11 +397,11 @@ public Authority getDefaultAuthority() {
         }
     }
 
-    private void checkManifestPermissions(){
-        if(this.handleNullTaskAffinity){
+    private void checkManifestPermissions() {
+        if (this.handleNullTaskAffinity) {
```
AdamBJohnsonx, line 401: handleNullTaskAffinity is a Boolean - probably would like to either convert it to a boolean or handle the potential for a null

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1357/commits/14d1c7be6f2e756427b9d41569586a107aa97993

shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1357/commits/8235fb5ce92a7a74f904bd1d33cc3009167c98ff

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1357/commits/36700831510bbbe5339b7b97602a4d43d1320c83

AdamBJohnsonx (approved):

PR #1364, opened 2021-04-15
Release/2.1.0

paulkagiri:
Version 2.1.0
----------
- [PATCH] Add extra '/' in the example intent filter (#1323)
- [PATCH] Better PublicClientApplicationConfiguration validation to fail when empty string is passed (#1324)
- [PATCH] Improve redirect_uri validation of PCA manifest (#1327)


AdamBJohnsonx (approved):

gitaumoses4 (approved):
LGTM

changelog
```diff
@@ -1,8 +1,15 @@
 MSAL Wiki : https://github.com/AzureAD/microsoft-authentication-library-for-android/wiki
+Version 2.1.0
+----------
+- [PATCH] Add extra '/' in the example intent filter (#1323)
+- [PATCH] Better PublicClientApplicationConfiguration validation to fail when empty string is passed (#1324)
+- [PATCH] Improve redirect_uri validation of PCA manifest (#1327)
```
shahzaibj, line 6: Since all the changes are PATCH, we should NOT increment the MINOR version and should only increment the patch version i.e we should be at `2.0.10` instead of `2.1.0`

shahzaibj (changes requested):
Left some feedback. Based on the changes it seems that the correct version needs to be `2.0.10`

changelog
paulkagiri: You're right @shahzaibj, let me make the correct release version

paulkagiri: I have made https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1365 and will close this PR

PR #1368, opened 2021-04-19
Shoatman/fix instrumented tests

shoatman:
- Fixed bugs in mergeConfiguration
- Removed empty ignored instrumented tests
- Returned 2 ignored tests to the living
- Adding support for desugaring
- Updated instrumented test script to use desugaring
- Update Configuration test code to ensure that default config is merged when the method includes a call to validate

NOTE: I'm thinking that we should record in configuration whether the defaults have been merged in.  So that we can fail validation immediately it has not.

### DIFF
--- a/msal/src/androidTest/java/com/microsoft/identity/client/PublicClientConfigurationTest.java
+++ b/msal/src/androidTest/java/com/microsoft/identity/client/PublicClientConfigurationTest.java
@@ -294,7 +294,7 @@ public class PublicClientConfigurationTest {
      */
     @Test(expected = IllegalArgumentException.class)
     public void testUnknownAudienceException() {
-        final PublicClientApplicationConfiguration configWithInvalidAudience = loadConfig(R.raw.test_pcaconfig_unknown_audience);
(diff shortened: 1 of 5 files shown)


iambmelt (approved):

AdamBJohnsonx (approved):

msal/src/androidTest/java/com/microsoft/identity/client/PublicClientConfigurationTest.java
```diff
@@ -294,7 +293,7 @@ public void testUnknownAuthorityException() {
      */
     @Test(expected = IllegalArgumentException.class)
     public void testUnknownAudienceException() {
-        final PublicClientApplicationConfiguration configWithInvalidAudience = loadConfig(R.raw.test_pcaconfig_unknown_audience);
+        final PublicClientApplicationConfiguration configWithInvalidAudience = loadConfigAndMerge(R.raw.test_pcaconfig_unknown_audience);
```
rpdome, line 296: we might want to just replace loadConfig with loadConfigAndMerge in every function.

rpdome (approved):

PR #1371, opened 2021-04-26
Check nullability of handleNullTaskAffinity

rpdome:
Tests are failing because some unit tests are not doing the config merge, and this field is not set in those config files.

I could've added the field in those config files, but figured it might make more sense doing a null check here.

also added -psugar to the build script.


shoatman (approved):

p3dr0rv (approved):

rpdome:
Shoudn't be needed once https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1368 goes in

PR #1380, opened 2021-05-01
Fix msal local variant build after consuming common4j

shahzaibj:
We need to consume common as transitive true now because it needs to resolve classes from common4j on the classpath

If we don't do this then MSAL Test App won't build because classes from common4j are not on the Runtime classpath and we get the following error:

error: cannot access ITelemetryDefaultObserver
    class TelemetryDefaultObserver implements ITelemetryDefaultObserver {
    ^
  class file for com.microsoft.identity.common.java.telemetry.observers.ITelemetryDefaultObserver not found

Related common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1338


msal/build.gradle
```diff
@@ -288,7 +288,8 @@ dependencies {
     androidTestImplementation "org.mockito:mockito-android:$rootProject.ext.mockitoAndroidVersion"
     // 'local' flavor dependencies
     localApi(project(":common")) {
-        transitive = false
+        transitive = true
```
iambmelt, line 291: As I understand it, we have 2 options here right?
- Consume the dependency transitively or....
- Add the dependency directly

Is there any particular reason to favor a transitive consumption model?

shahzaibj: Yes, we have both options. I think for now since this is just source either should be fine but I think once this gets published my idea behind transitive was that msal shouldn't need to think about which version of `common4j` to consume. MSAL just consumes a specific version of `common` and then that specific version of common will know which version of `common4j` to use.

Do you see any risks with that?

shahzaibj: Do you know what is considered as best practice in regards to this? Seems in our current MSAL we exclude transitives when consuming common, however, when we consume MSAL in MSAL Test App then we don't take a direct dependency on common but rather let transitive do its thing.

rpdome: I would prefer this to be hidden from consumers.

shahzaibj: @rpdome So to translate that - you are in the `transitive` boat as well?

iambmelt (approved):
Approving; we can reassess the consumption model for this as necessary later

msal/build.gradle
```diff
@@ -288,11 +288,13 @@ dependencies {
...
+        exclude module: "jcip-annotations"
```
rpdome, line 292: This is being used by broker, but not MSAL. Why are we excluding it here?

shahzaibj: Removed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1380/commits/b58463de8199c9b744566d015d20583e1515badf

iambmelt:
oh, I already signed off here :D

rpdome (approved):

PR #1389, opened 2021-05-06
Remove usage of internal-ish classes in test cases

rpdome:
- HttpResponse in android-common is a wrapper of java-common's. It can't be casted directly, and Common's TestUtil is already adopting the latter (which Is why it broke here).

- HttpClient is an interface and was already moved to java-common

Also update common


shahzaibj (approved):
If tests pass then let's ship it

iambmelt (approved):
Same feedback as @shahzaibj - if it builds clean :shipit:

rpdome:
Needed to downgrade power mock.

AdamBJohnsonx:
From what to what?


From: Dome Pongmongkol ***@***.***>
Date: Thursday, May 6, 2021 at 12:50 PM
To: AzureAD/microsoft-authentication-library-for-android ***@***.***>
Cc: Adam Johnson ***@***.***>, Review requested ***@***.***>
Subject: Re: [AzureAD/microsoft-authentication-library-for-android] Remove usage of internal-ish classes in test cases (#1389)

Needed to downgrade power mock.

PR #1394, opened 2021-05-12
Update test cases after migrating ests telemetry to common4j

rpdome:
SchemaConstants is internal, and is now moved to common4j


shahzaibj (approved):

iambmelt (approved):
FYI OneAuth is also impacted by some of this. We will need to think about getting a build ready so OneAuth can update its common version once this is all stabilized

PR #1395, opened 2021-05-13
Bugfix for PackageInspector

iambmelt:
Fixes a bug with querying package info when apps have an account type in AccountManager


package-inspector/src/main/java/com/microsoft/inspector/MainActivity.java
```diff
@@ -121,8 +122,14 @@ private void filterAndDisplayPackages(@Nullable final String filterText) {
             @Override
             public void onItemClick(AdapterView<?> adapterView, View view, int position, long l) {
                 try {
+                    String pkgName = packageNames.get(position);
+
+                    if (pkgName.endsWith(ACCOUNT_AFFINITY_SUFFIX)) {
```
shahzaibj, line 127: Why does it have that suffix? seems sort of a strange suffix

iambmelt: Any app that has an account type in AccountManager has one put on there so you can differentiate in the list... I'm totally open to changing it to something else

image

iambmelt: image

shahzaibj: I see. I think this is fine

shahzaibj (approved):
LGTM. Thanks @iambmelt

rpdome (approved):

PR #1397, opened 2021-05-14
Update Test App - To create multiple concurrent tasks

shoatman:
- Added new StartActivity and made this the first activity launched by the test app
- StartActivity in turn has a button to launch a new instance of MainActivity
- MainActivity has android:documentLaunchMode="always" which results in a new task every time the button is pressed
- From the recents view once launch task is clicked twice.  You will see 2 separate tasks each containing an instance of MainActivity
- This is necessary to to manually test the multiple concurrent sign in support that Samsung has requested.

Document Launch Mode:
https://developer.android.com/guide/topics/manifest/activity-element#dlmode
https://www.youtube.com/watch?v=4Y3JMvbcxQE

### DIFF
--- a/testapps/testapp/src/main/AndroidManifest.xml
+++ b/testapps/testapp/src/main/AndroidManifest.xml
@@ -41,12 +41,13 @@
         android:allowBackup="false"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
-        android:theme="@style/AppTheme">
+        android:theme="@style/AppTheme"
+        android:resizeableActivity="true">
         <!-- taskAffinity is set equal to empty string.  This translated to "null" taskAffinity at runtime.
             null task affinity may be used when creating multi-window apps-->
         <!-- https://developer.android.com/guide/topics/manifest/activity-element#aff -->
         <activity
-            android:name="com.microsoft.identity.client.testapp.MainActivity"
+            android:name="com.microsoft.identity.client.testapp.StartActivity"
             android:theme="@style/AppTheme.NoActionBar"
             android:windowSoftInputMode="stateHidden"
             android:taskAffinity="">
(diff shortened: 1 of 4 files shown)


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/StartActivity.java
```diff
@@ -0,0 +1,73 @@
...
+import android.util.AttributeSet;
+import android.view.View;
+import android.widget.Button;
+
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AppCompatActivity;
+
+public class StartActivity extends AppCompatActivity {
+
+    //private static final String TAG = MainActivity.class.getSimpleName();
```
AdamBJohnsonx, line 37:
```suggestion
    //private static final String TAG = StartActivity.class.getSimpleName();
```

testapps/testapp/src/main/AndroidManifest.xml
```diff
@@ -56,9 +57,17 @@
             </intent-filter>
         </activity>
 
+        <activity
+            android:name="com.microsoft.identity.client.testapp.MainActivity"
+            android:theme="@style/AppTheme.NoActionBar"
+            android:windowSoftInputMode="stateHidden"
+            android:taskAffinity=""
+            android:documentLaunchMode="always">
+        </activity>
```
AdamBJohnsonx, line 66: No intent filter here?

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/StartActivity.java
```diff
@@ -0,0 +1,73 @@
...
+
+            }
+        });
+
+    }
+
+    @Override
+    public View onCreateView(View parent, String name, Context context, AttributeSet attrs) {
+
+
```
AdamBJohnsonx, line 70:
```suggestion
```

AdamBJohnsonx (approved):
This is just adding test app features to launch new activities.  Cool.

iamgusain (approved):
:shipit:

testapps/testapp/src/main/AndroidManifest.xml
shoatman: No.  It's just need for the activity that you want to be used by the launcher.

PR #1398, opened 2021-05-16
Add additional metadata about the fetched token.

AdamBJohnsonx:
Actually dump some of the token data when we make a request, to aid in testing.

### DIFF
--- a/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
+++ b/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
@@ -63,6 +63,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.spec.InvalidKeySpecException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 
 import javax.crypto.SecretKey;
 import javax.crypto.SecretKeyFactory;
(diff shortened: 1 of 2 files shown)


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/ResultFragment.java
```diff
@@ -43,9 +43,16 @@
 public class ResultFragment extends Fragment {
 
     public static final String CORRELATION_ID = "correlation_id";
+    public static final String ACCOUNT_TENANT_ID = "account_tenant_id" ;
     static final String ACCESS_TOKEN = "access_token";
     static final String DISPLAYABLE = "displayable";
     static final String STRING_DATA_TO_DISPLAY = "string_data_to_display";
+    public static final String TENANT = "tenant";
```
shoatman, line 50: Is this home tenant vs account tenant which is from a specific authority which may or may not be the same as the home tenant?

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/ResultFragment.java
```diff
@@ -43,9 +43,16 @@
 public class ResultFragment extends Fragment {
 
     public static final String CORRELATION_ID = "correlation_id";
+    public static final String ACCOUNT_TENANT_ID = "account_tenant_id" ;
```
shoatman, line 46: We probably should distinguish between home tenant and guest tenants and their respective identifiers.  I think the account object has these different IDs are part of tenant profiles.

shoatman (changes requested):
Great to have more information in the result output.  I think we should make clear the difference between home tenant vs other tenants.  Thanks

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/ResultFragment.java
AdamBJohnsonx: To the best of my knowledge here, this is the tenant that issued the token, which may not be the same as the one specified in the authority parameter of the request.  Examining this value could show you the tenant mismatch that was seen in our recent problems.

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
```diff
@@ -193,7 +194,25 @@ public boolean onNavigationItemSelected(final MenuItem item) {
...
+                final Map<String, ?> claims1 = mAuthResult.getAccount().getClaims();
+                if (claims1 != null) {
+                    for (Map.Entry<String, ?> e : claims1.entrySet()) {
+                        claims.append(e.getKey() + " -> " + String.valueOf(e.getValue()));
+                    }
+                } else {
+                    claims.append("NONE!");
+                }
+                bundle.putString(ResultFragment.CLAIMS, claims1.toString());
+                bundle.putString(ResultFragment.ACCOUNT_TENANT, mAuthResult.getAccount().getTenantId());
```
rpdome, line 210: nit: ACCOUNT_TENANT_ID?

rpdome: alternatively, we could just print the whole mAuthResult.

PR #1399, opened 2021-05-16
Added 2 new settings to PCA Configuration

shoatman:
- concurrent_authorization_requests (boolean, default: false) - controls whether concurrent authorization requests are allowed
- authorization_in_current_task  (boolean, default:false)-  controls whether authorization occurs in the same task as the activity that started interactive request or occurs in a new task
- concurrent_authorization_requests setting is used in initializeConfigurationInternal to configure the command dispatcher to allow a new max threads in the interactive thread pool

### DIFF
--- a/msal/build.gradle
+++ b/msal/build.gradle
@@ -291,12 +291,14 @@ dependencies {
         transitive = true
     }
 
-    snapshotApi(group: 'com.microsoft.identity', name: 'common', version: '3.4.3', changing: true)
+    //snapshotApi(group: 'com.microsoft.identity', name: 'common', version: '3.4.3', changing: true)
 
     //TODO: we will have to change transitive to true once common4j is published
+    /*
     distApi("com.microsoft.identity:common:3.4.3") {
         transitive = false
     }
+    */
 
 }
 
--- a/msal/build.gradle
+++ b/msal/build.gradle
@@ -291,14 +291,13 @@ dependencies {
         transitive = true
     }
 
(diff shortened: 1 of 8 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfigurationFactory.java
```diff
@@ -50,6 +52,7 @@
 
 public class PublicClientApplicationConfigurationFactory {
     private static final String TAG = PublicClientApplicationConfigurationFactory.class.getSimpleName();
+    private static final int DEFAULT_MAX_THREAD_POOL_INTERACTIVE_CONCURRENT = 5;
```
gitaumoses4, line 55: Just curious is there any specific reason for choosing 5 here?

shoatman: Nope.  Open to suggestion, but didn't seem reasonable for it to be unlimited.  We used 5 for the silent pool in command dispatcher, so i made it match.

gitaumoses4 (approved):
Looks good. No more comments.

msal/build.gradle
```diff
@@ -298,6 +298,7 @@ dependencies {
         transitive = false
     }
 
+
 }
```
iamgusain, line 302: nit: remove

iamgusain (approved):
:shipit:

PR #1404, opened 2021-05-20
Fix msal dist dep conflict

shahzaibj:
@p3dr0rv noticed an issue with dependency resolution while building msal dist variant. dist was pulling in common 3.4.3 while testutils was trying to pull local module and that was causing the build to fail.


p3dr0rv (approved):
Thanks Shahzaib
I see MSAL Test App failing,  do we need to enable desugaring?

shahzaibj:
> Thanks Shahzaib
> I see MSAL Test App failing, do we need to enable desugaring?

Weird. We shouldn't have to...I'll take a look.

PR #1412, opened 2021-06-01
Fix test cases due to AuthRequests migration

rpdome:


msal/src/test/java/com/microsoft/identity/client/RoboTestCacheHelper.java
```diff
@@ -26,7 +26,7 @@
 import com.microsoft.identity.common.internal.authorities.AccountsInOneOrganization;
 import com.microsoft.identity.common.internal.authorities.Authority;
 import com.microsoft.identity.common.internal.cache.ICacheRecord;
-import com.microsoft.identity.common.internal.providers.microsoft.microsoftsts.MicrosoftStsAuthorizationRequest;
+import com.microsoft.identity.common.java.providers.microsoft.microsoftsts.MicrosoftStsAuthorizationRequest;
```
AdamBJohnsonx, line 29: I'm wondering if we need the 'com.microsoft.identity.common.java' package at all, and can't just use something like  com.microsoft.identity.common.providers here.

rpdome: I think we (as a team) agreed that we want the package name to be different between android-common and java-common. 

Changing the former means changing the whole library, so I went with the 2nd option.

rpdome: Will change them in a separate PR in one go.

AdamBJohnsonx (approved):

Issue #1416, opened 2021-06-03
When using an app URI with the token request the app URI is lowercased before being sent to server


shoatman:
**Describe the bug**
AAD returns an access token with an audience value based on the app URI provided as part of the resource identifier with the token request.  Although this feels like a server bug we are going to fix in the client to expedite the fix.

**Smartphone (please complete the following information):**
- All

**Stacktrace**
No failure in the code.  Resulting access token has the incorrect aud claim and as a result of the resource server receiving the access token rejects it.  Subsequent requests to MSAL will return the invalid token.

**To Reproduce**

1. Make a token request for a resource using a valid mixed case app id URI
2. Token return successfully
3. Submit token to resource server
4. Token rejected because audience is lower cased rather than mixed case as expected.

PR #1420, opened 2021-06-07
Fix mixed casing issue for acquire token with resource

shoatman:
Fixes the following issue: https://github.com/AzureAD/microsoft-authentication-library-for-android/issues/1416

### DIFF
--- a/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalWrapper.java
+++ b/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalWrapper.java
@@ -96,7 +96,7 @@ abstract class MsalWrapper {
 
         final AcquireTokenParameters.Builder builder = getAcquireTokenParametersBuilder(activity, requestOptions, callback);
         builder.withAuthorizationQueryStringParameters(null);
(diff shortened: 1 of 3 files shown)


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalWrapper.java
```diff
@@ -96,7 +96,7 @@ public void acquireTokenWithResource(@NonNull final Activity activity,
 
         final AcquireTokenParameters.Builder builder = getAcquireTokenParametersBuilder(activity, requestOptions, callback);
         builder.withAuthorizationQueryStringParameters(null);
-        builder.withResource(requestOptions.getScopes().toLowerCase().trim());
+        builder.withResource(requestOptions.getScopes().trim());
```
AdamBJohnsonx, line 99: Can we add a unit tests here that makes sure that the resource that we asked for here is present in the request that was generated?

shoatman: I haven't even run the existing unit tests yet, which this may have broken.  Nor have i testing with Broker... so lots of testing still to perform.

shahzaibj (approved):

PR #1423, opened 2021-06-09
Kill HTTP layer adapters. make changes to accommodate common

rpdome:
See: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1396

### DIFF
--- a/msal/src/androidTest/java/com/microsoft/identity/client/AndroidTestMockUtil.java
+++ b/msal/src/androidTest/java/com/microsoft/identity/client/AndroidTestMockUtil.java
@@ -31,6 +31,8 @@ import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.net.SocketTimeoutException;
 
+import javax.net.ssl.HttpsURLConnection;
+
 /**
  * Util class for mocking.
  */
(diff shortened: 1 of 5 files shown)


iambmelt:
@rpdome Can you include a changelog update in this PR calling out that this is breaking?

rpdome:
> @rpdome Can you include a changelog update in this PR calling out that this is breaking?
> 
> Edit - this PR doesn't appear to be breaking at this level - we should still have a changelog update calling out this change :)

In a sense, this PR is an equivalent of "update common". The actual work isn't in this layer.

I could definitely add it to change log though (didn't add it _now_ because I thought it would be very 'noisy', since there will be more PR on the way that looks like this).

rpdome:
NVM, just saw your edit. I suspect that it might have something to do with shadows. Let me take a look.

iambmelt (approved):
I don't see any issue with these changes, though it appears there may a failure or cancellation of some test: perhaps this is expected
image

rpdome:
> I don't see any issue with these changes, though it appears there may a failure or cancellation of some test: perhaps this is expected
> image

Issue with robolectric's flushScheduler (not sure why it just showed up).
Switching to the LEGACY mode fixed it. There's a PBI for fixing this in the PAUSED mode https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/1377749

iambmelt (approved):

shahzaibj (approved):

AdamBJohnsonx (approved):

rpdome:
rebased dev.

PR #1426, opened 2021-06-09
Fix lint task problem, update version, update dist common dependency

shoatman:
- Removed lint as a task dependency for all things for now.  Lint requires you to specify the variant to apply it to.  Currently it was running for all variants which leads to compilation failures for different variants.
- Updated the version of library
- Updated the version of the dependency
https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20Android/Backlog%20items

### DIFF
--- a/msal/build.gradle
+++ b/msal/build.gradle
@@ -276,7 +276,7 @@ dependencies {
 
     snapshotApi(group: 'com.microsoft.identity', name: 'common', version: '3.4.3', changing: true)
 
-    distApi("com.microsoft.identity:common:3.4.3") {
+    distApi("com.microsoft.identity:common:3.4.4-RC1") {
         transitive = false
     }
 }
@@ -418,6 +418,8 @@ publishing {
 }
 
 tasks.whenTaskAdded { task ->
+
+
     if (task.name.contains('assemble')) {
         task.dependsOn 'pmd'
     }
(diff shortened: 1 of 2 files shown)


msal/build.gradle
```diff
@@ -418,6 +418,8 @@ publishing {
 }
 
 tasks.whenTaskAdded { task ->
+
+
```
shahzaibj, line 422: probably one empty line is good enough

msal/build.gradle
```diff
@@ -426,6 +428,10 @@ tasks.whenTaskAdded { task ->
         && !task.name.contains('Snapshot')
         && !task.name.contains('Test')
         && !task.name.contains('Local')) {
-        task.dependsOn 'lint', 'javadocJar', 'sourcesJar'
+        task.dependsOn 'javadocJar', 'sourcesJar'
     }
+
+
+
+
```
shahzaibj, line 436: probably one empty line is good enough

msal/build.gradle
```diff
@@ -426,6 +428,10 @@ tasks.whenTaskAdded { task ->
         && !task.name.contains('Snapshot')
         && !task.name.contains('Test')
         && !task.name.contains('Local')) {
-        task.dependsOn 'lint', 'javadocJar', 'sourcesJar'
```
AdamBJohnsonx, line 429: Was there a related pipeline update that we should reference?

AdamBJohnsonx (approved):
Just a question

shahzaibj (approved):
LGTM

msal/build.gradle
shoatman: I filed a workitem to fix this in the gradle.  Link above.

PR #1427, opened 2021-06-09
update ref in Maven release

p3dr0rv:
in the msal maven release yaml the reference was pointing to a incorrect branch , we change it to point dev

### DIFF
--- a/azure-pipelines/maven-release/msal-maven-release.yml
+++ b/azure-pipelines/maven-release/msal-maven-release.yml
@@ -18,7 +18,7 @@ resources:
   - repository: common
     type: github
     name: AzureAD/microsoft-authentication-library-common-for-android
-    ref: pedroro/release-pipelines
+    ref: dev
     endpoint: ANDROID_GITHUB
 
 jobs:


shoatman (approved):

azure-pipelines/maven-release/msal-maven-release.yml
```diff
@@ -18,7 +18,7 @@ resources:
   - repository: common
     type: github
     name: AzureAD/microsoft-authentication-library-common-for-android
-    ref: pedroro/release-pipelines
+    ref: dev
```
shahzaibj, line 21: At some point we should investigate if we can move away from this model and just reference the templates from the submodule

shahzaibj (approved):

azure-pipelines/maven-release/msal-maven-release.yml
p3dr0rv: Yes, that is what I would like to achieve at some point, but I have not had the time to investigate and do it this way

PR #1462, opened 2021-07-14
Initial commit of migration of shared preferences to common4j.

AdamBJohnsonx:
Adapt to https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1465


msal/build.gradle
```diff
@@ -259,6 +258,7 @@ task sourcesJar(type: Jar) {
 }
 
 dependencies {
+    testImplementation project(path: ':common4j')
```
rpdome, line 261: hmm. we shouldn't need this :/

rpdome (approved):

msal/build.gradle
shahzaibj: yeah seems strange. We might be trying to call a common4j API directly in a test in msal and then we need this. Ideally we shouldn't do that and only call the Android common from msal

shahzaibj (approved):

PR #1469, opened 2021-07-27
Update common, adopt build plugin

AdamBJohnsonx:
Rather than mess with trying to make the tests work, I've just brought over the build plugin.


scripts/run-instrumented-tests.sh
```diff
@@ -9,5 +9,5 @@ gradle -version
 echo =============================================
 echo Running instrumented tests
 echo =============================================
-gradle msal:connectedLocalDebugAndroidTest -i
+gradle msal:connectedLocalDebugAndroidTest Psugar=true -i
```
AdamBJohnsonx, line 12:
```suggestion
gradle msal:connectedLocalDebugAndroidTest -Psugar=true -i
```

PR #1495, opened 2021-08-20
Refresh on

aherciya:
Msal changes for refreshOn.


msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/local/TestCase1477854.java
```diff
@@ -0,0 +1,132 @@
...
+                        .handlePrompt(username, password);
+            }
+        }, TokenRequestTimeout.MEDIUM);
+
+        authResult.assertSuccess();
+
+        //fast forward 1:35 min - means refresh expired
+        TestContext.getTestContext().getTestDevice().getSettings().forwardDeviceTime(5700);
+
+        //acquire token silently - should refresh token, and get new AT
```
shoatman, line 88: It should return the current token... and then refresh.  I think we need to test that the current token is returned and that the token in the cache is refreshed.

shoatman (changes requested):
Just have a question about the test.  Otherwise looks right to me.

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/local/TestCase1477854.java
aherciya: Yes, I need to update this test case.. I think we need 2 silent calls essentially.

shoatman (approved):
lgtm

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/local/TestCase1477854.java
```diff
@@ -0,0 +1,169 @@
...
+        // f) Second Silent request returns access token B
+        authResult3.assertSuccess();
+
+        // g) Verify that 2nd silent request... after refresh in expired... now returns a new AT by verifying Silent 1 response not equal to silent 2 response.
+        Assert.assertFalse(firstSilentResult.isAccessTokenEqual(secondSilentResult.getAccessToken()));
+
+    }
+
+
+    @Override
```
iamgusain, line 144: should these both be null? getLabUserQuery, getTempUserType
I think we throw an exception in case both of these are null.

iamgusain (approved):
I left a comment in the test, otherwise LGTM

PR #1507, opened 2021-09-01
Token Cache Documentation

shoatman:
I"ll need to rebased this on dev, but you can read through it anyway.


docs/tokencache.md
```diff
@@ -2,16 +2,350 @@
...
+  - ID Token from other tenants that the account is a member of
+- Credentials
+  - Id Token
+  - Access Tokens
+  - Refresh Tokens
+  - etc....
+
+### Relationships
+
+1 Account has 0 or more credentials
```
iambmelt, line 36:
```suggestion
1 Account has 1 or more credentials (minimum, the `AccountRecord` + `RefreshTokenRecord`)
```

For the purposes of how calls like `getAccounts()` works, [the refresh token must be present](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java#L1193-L1198) for us to consider the account "available" -- this is to ensure we don't wind up in a situation where `getAccounts()` returns an `AccountRecord` for which no RT exists (meaning we couldn't do any silent auth)

docs/tokencache.md
```diff
@@ -2,16 +2,350 @@
...
+
+This section describes the base class for the higher level of the TokenCache.  The level responsible for translated an oAuth response into universal cache schema artifacts.
+
+There are different implementations of the cache based on the library using it and the scenarios it needs to support.
+
+> NOTE: I'm going to ignore the ADALOAuth2TokenCache in this document. It deals with interacting with the legacy ADAL token cache.
+
+### Abstract base class
+
+The abstract base class for all caches is the OAuth2TokenCache.  The physical schema of the cache in terms of account and credential records is decoupled from the protocol used to acquire the account and credential; however the cache interface knows about oAuth2 token responses and how to convert those to accounts and crentials and persist them to storage.
```
iambmelt, line 49: Accurate -- _ideally_ the `OAuth2TokenCache` should delegate to the strategy (or an object provided by the strategy) to produce the account/credential types... this wasn't implemented this way, but with the benefit of hindsight, this would be a future "nice-to-have"

docs/tokencache.md
```diff
@@ -2,16 +2,350 @@
...
+     * @throws ClientException If tokens cannot be successfully saved.
+     */
+    public abstract ICacheRecord save(final T oAuth2Strategy,
+                                      final U request,
+                                      final V response) throws ClientException;
+```
+
+### MSALOAuth2TokenCache
+
+The simplest version of the cache is the MSALOAuth2TokenCache.  There is one instance of this cache per PublicClientApplication created using MSAL within a client android application.  The important thing to note is that there is 1 cache per public client application identified by it's client id.
```
iambmelt, line 69: This is all true and correct -- this is how the cache is supposed to be used. (Fullstop!)

I will add (and maybe this doesn't belong in your doc, but is something to be aware of) there is one little wrinkle to this and it has to do with a quirk/hack to support TFW/TFL/TSL sharing a binary... there is 1 piece of hackery happening [here](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/dev/common4j/src/main/com/microsoft/identity/common/java/cache/MsalOAuth2TokenCache.java#L722-L750) where we attempt to look up an FRT even if it doesn't match our client id... It is assumed that only 1Ps will ever exercise this code path since TSL is 1P only and TFL/TFW are also 1P

iambmelt (approved):
Docs LGTM - please feel free to reach out if there are any questions you have as you work with these layers!

docs/tokencache.md
```diff
@@ -2,16 +2,350 @@
...
+     * @param t      AccountCredentialBase
+     * @param <T>    Generic type
+     * @return AccountCredentialBase
+     */
+    <T extends AccountCredentialBase> T fromCacheValue(final String string, Class<? extends AccountCredentialBase> t); // TODO consider throwing an Exception if parsing fails
+```
+
+## Storage
+
+Accounts and credentials are serialized and stored using the lowest common denominator storage method available on Andorid... SharedPreferences.
```
AdamBJohnsonx, line 162: I mean, we *could* just use files, and then wrap them using BDB or whatever k-v implementation we would like.  There may be good reasons to do this...

shoatman: Not commenting on how to make this better... i think the answer is likely move to SQL Lite rather than using shared preferences... ultimately shared preference is a file... wrapped.

AdamBJohnsonx (approved):

docs/tokencache.md
```diff
@@ -2,16 +2,350 @@
...
+                                      final V response) throws ClientException;
+```
+
+### MSALOAuth2TokenCache
+
+The simplest version of the cache is the MSALOAuth2TokenCache.  There is one instance of this cache per PublicClientApplication created using MSAL within a client android application.  The important thing to note is that there is 1 cache per public client application identified by it's client id.
+
+### Family of Client Ids (Special Case)
+
+Applications published by Microsoft may be authorized to share refresh tokens with one another.  All applications that are authorized to share tokens share a single MSALOAuth2TokenCache instance.  They effectively share a client id, which in this case is the family id (generally 1 is used to represent the family)
```
somalaya: How are we currently verifying if an App B is authorized to use the FRT populated by App A? I could find the ESTS call to fetch the token with FRT and verify that it is authorized to use it.  Do we have to make this network call every time for a new app?

shoatman: AndroidBrokerPlatformComponents.java in the broker repo has an isAuthorizedToShareTokens method that you may want to review.

docs/tokencache.md
```diff
@@ -2,16 +2,350 @@
...
 ## Introduction
 
-> PENDING
+The token cache implementation for MSAL, ADAL and Android broker is found in the Android Common Library.  
 
-## Class Diagram
+## References 
 
-> PENDING
+- Universal Cache Schema
```
shahzaibj: Needs to be a hyperlink I believe?

docs/tokencache.md
```diff
@@ -2,16 +2,350 @@
...
+## Universal cache and protocol response handling
 
-> PENDING
+When looking at the token cache there are really 2 levels to it.  The lower level is based on the universal cache schema referenced above.  The higher level deals with the result of a particular protocol request and mapping that to the universal schema.  
+
+So in effect we get an oAuth 2 Token Response, which in the case of MSAL includes an ID Token and an Access Token and we translate:
+
+ID Token & Client_Info -> Account & Tenant Profile
+Access Token -> Credential
+Refresh Token -> Credential
```
shahzaibj: These need to be changed to bulleted list (or list of some sort)...otherwise, it renders as follows:

> ID Token & Client_Info -> Account & Tenant Profile Access Token -> Credential Refresh Token -> Credential

docs/tokencache.md
```diff
@@ -2,16 +2,350 @@
...
-> PENDING
+- Universal Cache Schema
 
-## Interaction Diagram
+## Universal cache and protocol response handling
 
-> PENDING
+When looking at the token cache there are really 2 levels to it.  The lower level is based on the universal cache schema referenced above.  The higher level deals with the result of a particular protocol request and mapping that to the universal schema.  
+
+So in effect we get an oAuth 2 Token Response, which in the case of MSAL includes an ID Token and an Access Token and we translate:
```
shahzaibj, line 17:
```suggestion
So in effect we get an oAuth2 Token Response, which in the case of MSAL includes an ID Token and an Access Token and we translate:
```

docs/tokencache.md
```diff
@@ -2,16 +2,350 @@
...
+  - ID Token from other tenants that the account is a member of
+- Credentials
+  - Id Token
+  - Access Tokens
+  - Refresh Tokens
+  - etc....
+
+### Relationships
+
+1 Account has 1 or more credentials (minimum, the `AccountRecord` + `RefreshTokenRecord`)
```
shahzaibj, line 38: I don't really get what this means...particularly the part in parenthesis; can you elaborate?

shoatman: I'll do so in the doc, but in unified schema AccountRecords are considered credentials in addition to RefreshTokenRecords and AccessTokenRecords.... the reason that an account record is considered a credential is that it contains an id_token.

docs/tokencache.md
shoatman: Ack.  I'll update.

docs/tokencache.md
shoatman: Ack.  I'll add.

docs/tokencache.md
shahzaibj: Actually may be rather `OAuth 2.0`

This is just a nit...just trying myself to figure out the best way to write oauth2 :)

docs/tokencache.md
```diff
@@ -2,16 +2,350 @@
...
 
-## Interaction Diagram
+## Universal cache and protocol response handling
 
-> PENDING
+When looking at the token cache there are really 2 levels to it.  The lower level is based on the universal cache schema referenced above.  The higher level deals with the result of a particular protocol request and mapping that to the universal schema.  
+
+So in effect we get an oAuth 2 Token Response, which in the case of MSAL includes an ID Token and an Access Token and we translate:
+
+ID Token & Client_Info -> Account & Tenant Profile
```
rpdome: Do we also store ID token as a credential?

docs/tokencache.md
```diff
@@ -2,16 +2,350 @@
...
+                                      final U request,
+                                      final V response) throws ClientException;
+```
+
+### MSALOAuth2TokenCache
+
+The simplest version of the cache is the MSALOAuth2TokenCache.  There is one instance of this cache per PublicClientApplication created using MSAL within a client android application.  The important thing to note is that there is 1 cache per public client application identified by it's client id.
+
+### Family of Client Ids (Special Case)
+
```
rpdome, line 100: They effectively share a client id <<< Each app has their own client id, right? So I think this part could be confusing/misleading.

docs/tokencache.md
```diff
@@ -2,16 +2,350 @@
 
 ## Introduction
 
-> PENDING
+The token cache implementation for MSAL, ADAL and Android broker is found in the Android Common Library.
```
rpdome, line 5: If you have time, I would suggest adding a diagram to show the relationship between classes in this doc (and how the flow works). I think that'll greatly help newcomers.

docs/tokencache.md
```diff
@@ -2,16 +2,350 @@
...
+
+### AccessTokenRecord
+
+Inherits from Credential and contains additional fields specific to access tokens.
+
+### RefreshTokenRecord
+
+Inherits from Credential and contains additioonal fields specific to refresh tokens.
+
+### PrimaryRefreshTokenRecord
```
shahzaibj, line 129: This is interesting. I didn't actually knew we had this. Do you know what's the purpose of this class? Does having this type of record mean that even PRT is saved in SharedPreferences? From my knowledge, PRT was only saved in Account Manager.

Furthermore, how is this record different from [PRTv2](https://github.com/AzureAD/ad-accounts-for-android/blob/dev/broker4j/src/main/com/microsoft/identity/broker4j/broker/prtv2/PrtV2.java)? I think it would be good to document those differences here (or somewhere).

docs/tokencache.md
shoatman: Yes.  I'll update the docs to indicate that.

docs/tokencache.md
shoatman: I linked to the spec as well. ;-)

docs/tokencache.md
shoatman: I agree.  I updated this.

docs/tokencache.md
shoatman: This is being used by OneAuth for a 1P application that authorized to use the PRT protocol.  When we look at the refactoring for the latest PRT protocol.... this is technical issue that we can look into.  @mohitc1 - FYI

docs/tokencache.md
shoatman: Done.

PR #1517, opened 2021-09-17
Allow authenticator to pass older format for redirect uri

iamgusain:
### What is this change
Allowing Authenticator app to pass older format ("urn:ietf:........:oob") of redirect uri, and exempt it from redirect uri verification.

### Why we need this change
As part of Authenticator moving to msal, they are switching to Broker redirect uri format (msauth://<authority>).
On the broker side we have [logic to override the authenticator redirect uri to older format](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/a89e36d3ad4b7c5a9fe2d6bd3a37a418aa509c1a/common/src/main/java/com/microsoft/identity/common/internal/broker/PackageHelper.java#L144). 
This causes the [Broker side validation of the redirect uri](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/a89e36d3ad4b7c5a9fe2d6bd3a37a418aa509c1a/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerValidator.java#L305) when the active broker is going to be Company Portal (with older versions)
In order to solve this Authenticator is going to continue to pass old redirect uri for older version of Company portal.
However MSAL doesn't allow older format for redirect uri., 

###  How
We are adding an exemption inthe redirect uri verification logic. To check if the package name and signature matchess with Authenticator, pass the validation if the redirect uri passed is older format.

### Testing
Going to share a build with Authenticator team to validate the changes.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
@@ -648,7 +648,7 @@ public class PublicClientApplicationConfiguration {
                 }
             }
         } catch (final PackageManager.NameNotFoundException | NoSuchAlgorithmException e) {
-            Logger.error(TAG, "Unexpected error in getting package info/signature for Autheticator", e);
+            Logger.error(TAG, "Unexpected error in getting package info/signature for Authenticator", e);
         }
 
         return false;
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
@@ -489,7 +489,7 @@ public class PublicClientApplicationConfiguration {
         boolean isInvalid = false;
         if (mAppContext.getPackageName().equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME))
         {
-            isInvalid = verifyAuthenticatorRedirectUri();
+            isInvalid = !isValidAuthenticatorRedirectUri();
         } else {
             isInvalid = TextUtils.isEmpty(redirectUri) || !hasSchemeAndAuthority(redirectUri);
         }
@@ -604,7 +604,7 @@ public class PublicClientApplicationConfiguration {
 
         if (mAppContext.getPackageName().equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME))
         {
-            if (verifyAuthenticatorRedirectUri())
+            if (isValidAuthenticatorRedirectUri())
             {
                 return;
             }
@@ -623,7 +623,7 @@ public class PublicClientApplicationConfiguration {
         verifyRedirectUriWithAppSignature();
     }
 
-    private boolean verifyAuthenticatorRedirectUri() {
+    private boolean isValidAuthenticatorRedirectUri() {
         try {
             final PackageInfo info = mAppContext.getPackageManager().getPackageInfo(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME, PackageManager.GET_SIGNATURES);
             if (info != null && info.signatures != null && info.signatures.length > 0) {
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
@@ -536,15 +536,6 @@ public class PublicClientApplicationConfiguration {
                 final MessageDigest messageDigest = MessageDigest.getInstance("SHA");
                 messageDigest.update(signature.toByteArray());
                 final String signatureHash = Base64.encodeToString(messageDigest.digest(), Base64.NO_WRAP);
-
-                if (packageName.equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)
-                    && (signatureHash.equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_RELEASE_SIGNATURE)
-                        || signatureHash.equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_DEBUG_SIGNATURE))
-                    && mRedirectUri.equalsIgnoreCase(AuthenticationConstants.Broker.BROKER_REDIRECT_URI))
-                {
-                    return;
-                }
-
                 final Uri.Builder builder = new Uri.Builder();
                 final Uri uri = builder.scheme("msauth")
                         .authority(packageName)
@@ -604,6 +595,27 @@ public class PublicClientApplicationConfiguration {
 
         if (!mUseBroker) {
             return;
+        }
+
+        if (mAppContext.getPackageName().equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME))
+        {
+            try {
+                final PackageInfo info = mAppContext.getPackageManager().getPackageInfo(mAppContext.getPackageName(), PackageManager.GET_SIGNATURES);
+                if (info != null && info.signatures != null && info.signatures.length > 0) {
+                    Signature signature = info.signatures[0];
+                    MessageDigest md = MessageDigest.getInstance("SHA");
+                    md.update(signature.toByteArray());
+                    final String signatureHash = Base64.encodeToString(md.digest(), Base64.NO_WRAP);
+                    if (signatureHash.equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_RELEASE_SIGNATURE)
+                            || signatureHash.equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_DEBUG_SIGNATURE)
+                            && mRedirectUri.equalsIgnoreCase(AuthenticationConstants.Broker.BROKER_REDIRECT_URI)) {
+                        return;
+                    }
+                }
+            }
+            catch (PackageManager.NameNotFoundException | NoSuchAlgorithmException e) {
+                Logger.error(TAG, "Unexpected error in getting package info/signature for Autheticator", e);
+            }
         }
 
         if (!isBrokerRedirectUri(mRedirectUri, mAppContext.getPackageName())) {
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
@@ -39,6 +39,7 @@ import com.microsoft.identity.client.configuration.HttpConfiguration;
 import com.microsoft.identity.client.configuration.LoggerConfiguration;
 import com.microsoft.identity.client.exception.MsalClientException;
 import com.microsoft.identity.client.internal.MsalUtils;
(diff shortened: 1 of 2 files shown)


tanmaymanolkar1 (approved):
LGTM!

tanmaymanolkar1 (changes requested):
public static boolean isBrokerRedirectUri(final @NonNull String redirectUri, final @NonNull String packageName) : we would have to make some changes here as well in the same file to let the Auth app use broker if the send the oob.ietf... redirect URI

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -597,6 +602,14 @@ public void checkIntentFilterAddedToAppManifestForBrokerFlow() throws MsalClient
             return;
         }
 
+        if (mAppContext.getPackageName().equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME))
```
rpdome, line 605: Should this go into isBrokerRedirectUri()?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -597,6 +602,14 @@ public void checkIntentFilterAddedToAppManifestForBrokerFlow() throws MsalClient
             return;
```
rpdome, line 602: I'm wondering if the change from this line until the end of the function belongs here.

It might make more sense for this to go into validateConfiguration(), or, specfically, validateRedirectUrl().

With that, we wouldn't need to put this special check in multiple places.

rpdome (commented):
left some suggestions.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -485,8 +486,13 @@ void validateConfiguration() {
...
     private void validateRedirectUri(@NonNull final String redirectUri) {
-        final boolean isInvalid = TextUtils.isEmpty(redirectUri) || !hasSchemeAndAuthority(redirectUri);
-
+        boolean isInvalid = false;
+        if (mAppContext.getPackageName().equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME))
+        {
+            isInvalid = !isValidAuthenticatorRedirectUri();
+        } else {
+            isInvalid = TextUtils.isEmpty(redirectUri) || !hasSchemeAndAuthority(redirectUri);
+        }
```
shahzaibj, line 495: This formatting of if blocks doesn't match our repo code style

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -597,6 +602,14 @@ public void checkIntentFilterAddedToAppManifestForBrokerFlow() throws MsalClient
             return;
         }
 
+        if (mAppContext.getPackageName().equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME))
+        {
+            if (isValidAuthenticatorRedirectUri())
+            {
+                return;
+            }
+        }
```
shahzaibj, line 611: code formatting doesn't match repo style

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -610,4 +623,32 @@ public void checkIntentFilterAddedToAppManifestForBrokerFlow() throws MsalClient
         verifyRedirectUriWithAppSignature();
     }
 
+    private boolean isValidAuthenticatorRedirectUri() {
```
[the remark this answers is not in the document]
shahzaibj: +1

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -485,8 +486,13 @@ void validateConfiguration() {
     }
 
     private void validateRedirectUri(@NonNull final String redirectUri) {
-        final boolean isInvalid = TextUtils.isEmpty(redirectUri) || !hasSchemeAndAuthority(redirectUri);
-
+        boolean isInvalid = false;
+        if (mAppContext.getPackageName().equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME))
+        {
```
AdamBJohnsonx, line 491:
```suggestion
        if (mAppContext.getPackageName().equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)) {
```

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -597,6 +602,14 @@ public void checkIntentFilterAddedToAppManifestForBrokerFlow() throws MsalClient
             return;
         }
 
+        if (mAppContext.getPackageName().equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME))
+        {
```
AdamBJohnsonx, line 606:
```suggestion
        if (mAppContext.getPackageName().equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)) {
```

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -597,6 +602,14 @@ public void checkIntentFilterAddedToAppManifestForBrokerFlow() throws MsalClient
...
+            if (isValidAuthenticatorRedirectUri())
+            {
```
AdamBJohnsonx, line 608:
```suggestion
            if (isValidAuthenticatorRedirectUri()) {
```

shahzaibj (commented):
changelog?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
AdamBJohnsonx: Do we want to try and gate this on the active broker being CP?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -485,8 +486,13 @@ void validateConfiguration() {
     }
 
     private void validateRedirectUri(@NonNull final String redirectUri) {
-        final boolean isInvalid = TextUtils.isEmpty(redirectUri) || !hasSchemeAndAuthority(redirectUri);
-
+        boolean isInvalid = false;
+        if (mAppContext.getPackageName().equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME))
```
AdamBJohnsonx, line 490: We have a. null-safe comparison, we should use it.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -610,4 +623,32 @@ public void checkIntentFilterAddedToAppManifestForBrokerFlow() throws MsalClient
         verifyRedirectUriWithAppSignature();
     }
 
+    private boolean isValidAuthenticatorRedirectUri() {
+        try {
+            final PackageInfo info = mAppContext.getPackageManager().getPackageInfo(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME, PackageManager.GET_SIGNATURES);
+            if (info != null && info.signatures != null && info.signatures.length > 0) {
+                Signature signature = info.signatures[0];
```
AdamBJohnsonx, line 630: final where possible

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
iamgusain: I think the purpose of isBrokerRedirectUri() is to check the schema and then set/unset if the Broker should be used.
As this is a specific case I think we should keep it separate.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
iamgusain: can you point me to it

AdamBJohnsonx: StringUtil.equalsIgnoreCase or ObjectUtils.equals, I believe.

iamgusain: I reversed the comparison to be safer

AdamBJohnsonx (approved):

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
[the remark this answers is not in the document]
rpdome: I think what I was trying to say was.... 

1. Apart from the oob check, the rest of the logic is identical.
2. if we add an early-return check like i did, we could just remove that line.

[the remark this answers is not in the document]
rpdome: Early return also makes it easier to see the deltas, IMO.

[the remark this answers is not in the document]
iamgusain: We still need to verify the package name and signature before returning early, right
So it is not that early :).. 
Let me know if you think the suggested change is must?

rpdome:
as per convo with @titorodi, please also add msauth://Microsoft.AAD.BrokerPlugin to the allow list.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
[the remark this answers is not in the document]
rpdome: We don't. just move line 643 /w Package name check all the way up.

if (package name = authApp AND sig = the old broker one){
 return true
}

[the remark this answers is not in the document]
rpdome: (In general, i'm a fan of less code, if possible - it's easier to fix/read for other engineers who are not involved

In this case, i think it's a small change, so we should do it.)

tanmaymanolkar1:
@titorodi confirmed using urn:ietf:wg:oauth:2.0:oob for direct WPJ token acquisition works.
TODO is allowing the prod local and official redirect URIs on the server side for Authenticator

tanmaymanolkar1 (approved):

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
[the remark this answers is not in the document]
iamgusain: That will force us to also update the order from line 610-620. Otherwise the whole logic will fail.
I can make the change but I don't think it's necessary as we are going to remove this code anyway, 
So I am trying to avoid modifying anything that is existing and just adding additional checks for Authenticator.

rpdome:
> @titorodi confirmed using urn:ietf:wg:oauth:2.0:oob for direct WPJ token acquisition works.
> TODO is allowing the prod local and official redirect URIs on the server side for Authenticator

I assume that this means they're okay with urn:ietf:wg:oauth:2.0:oob?

rpdome (approved):

PR #1522, opened 2021-09-24
Tamanolk/additional logging get accounts

tanmaymanolkar1:


msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
```diff
@@ -260,11 +270,14 @@ public void onTaskCompleted(final List<ICacheRecord> result) {
...
                                             if (accountMatcher.matches(trimmedIdentifier, account)) {
+                                                Logger.info(TAG + methodName,
+                                                        "Account found, Successful callback for identifier:" + identifier);
                                                 callback.onTaskCompleted(account);
                                                 return;
                                             }
                                         }
-
+                                        Logger.info(TAG + methodName,
+                                                "No accounts found callback for identifier:" + identifier);
```
AdamBJohnsonx, line 280: Here, can we actually dump the list of accounts that we didn't find anything for?

tanmaymanolkar1: added

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -2058,6 +2078,22 @@ boolean matches(@NonNull final String localAccountId,
...
+                                    TAG + methodName,
+                                    "For localAccountId identifier: " + localAccountId +
+                                            " and profileEntry account id :" + profileEntry.getValue().getId() +
+                                            ", match : " + localAccountId.contains(profileEntry.getValue().getId())
+                            );
+                        }
+
                         if (!TextUtils.isEmpty(profileEntry.getValue().getId()) &&
                                 localAccountId.contains(profileEntry.getValue().getId())) {
                             return true;
```
AdamBJohnsonx, line 2099: Down there on line 2105, I'd like us to log the fact that we are returning false with no match.

AdamBJohnsonx (approved):

PR #1530, opened 2021-09-30
Build.Gradle fixes for msal release branch

iamgusain:
### In this change

1.  Fixing below error which occur when running javadoc task as part of publish

```
Could not determine the dependencies of task ':msal:javadoc'.
> Resolving dependency configuration 'implementation' is not allowed as it is defined as 'canBeResolved=false'.
Instead, a resolvable ('canBeResolved=true') dependency configuration that extends 'implementation' should be resolved.

```

2. Setting the transitive property on android-common dep to true, so that it brings common4j along with it, and we dont have to explicitly add common4j dependency.

### DIFF
--- a/msal/build.gradle
+++ b/msal/build.gradle
@@ -233,7 +233,7 @@ android {
 task javadoc(type: Javadoc) {
     failOnError false
     source = android.sourceSets.main.java.srcDirs
-    classpath += configurations.implementation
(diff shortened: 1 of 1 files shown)


AdamBJohnsonx (approved):
Close and closer to gradle 7 compatibility.

shahzaibj (approved):

PR #1554, opened 2021-11-17
Code coverage fixes

paulkagiri:
This PR aims to address 3 main issues:
- To fix the exception being raised when one tries to run instrumented tests with code coverage turned on.
    ```
    java.lang.NoClassDefFoundError: Failed resolution of: Lorg/jacoco/agent/rt/internal_3570298/Offline;
    at com.microsoft.identity.client.Logger.$jacocoInit(Unknown Source:13)
    at com.microsoft.identity.client.Logger.<clinit>(Unknown Source:0)
    ...
    ```

- Adding the ability to turn code coverage on or off easily without affecting the rest of the pipelines - https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1554/files#diff-e68f47086ad13a5c9cc3889ecde9cac93108998f41c2db78824c2738316f2eafR23

- Moving the code coverage pipeline from Github Actions to Azure Pipelines - this is because there was the need to run e2e tests which require the test automation credentials only available to Azure Pipelines


AdamBJohnsonx (approved):

shahzaibj:
What are all these jacoco agent properties files?

shahzaibj:
It is not clear to me how this PR fixes the issue...you seem to just turn coverage back on...how does that fix the issue?

AdamBJohnsonx:
> It is not clear to me how this PR fixes the issue...you seem to just turn coverage back on...how does that fix the issue?

IIRC, having the output set to none causes Jacobo not to try to load that class that isn't in the jar... but my memory could be playing tricks on me here.

paulkagiri:
> What are all these jacoco agent properties files?

These were to allow Jacoco to be properly initialized but we no longer need them as a code coverage enable/disable flag has been added - https://github.com/AzureAD/android-complete/pull/120

paulkagiri:
> 

It was fixing the issue by adding the files Jacoco needed to get properly initialized - see comment above

paulkagiri:
> > It is not clear to me how this PR fixes the issue...you seem to just turn coverage back on...how does that fix the issue?
> 
> IIRC, having the output set to none causes Jacobo not to try to load that class that isn't in the jar... but my memory could be playing tricks on me here.

See comment above - we no longer need the files

.github/workflows/codecov.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         run: ./gradlew clean msal:assembleLocal
 
       - name: Run Coverage
-        run: ./gradlew msal:localDebugMsalUnitTestCoverageReport -ProbolectricSdkVersion=$ROBOELECTRIC_SDK_VERSION
+        run: ./gradlew msal:localDebugMsalUnitTestCoverageReport -ProbolectricSdkVersion=$ROBOELECTRIC_SDK_VERSION -PcodeCoverageEnabled=true
```
shahzaibj, line 29: I think we need `-Plabtest` here. We have lots of E2E tests that use the Lab Api. Since these are E2E, they are going to provide lots of coverage. We need to know about these coverage details in the coverage report.

shahzaibj (changes requested):
Looks like coverage pipeline is not running e2e lab api tests.

.github/workflows/codecov.yml
paulkagiri: > 

The main issue that I encountered is setting up the lab api automation for installing certificates and fetching credentials since this is being ran from a Github action. I would appreciate if you could float me any ideas you have around achieving this

paulkagiri:
> Looks like coverage pipeline is not running e2e lab api tests.

https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1554#discussion_r758464826

.github/workflows/codecov.yml
shahzaibj: Codecov does support Azure Pipelines as CI Provider: https://docs.codecov.com/docs/supported-ci-providers

paulkagiri: Kindly note that this pipeline is already in place now - https://dev.azure.com/IdentityDivision/Engineering/_build?definitionId=1522&_a=summary

paulkagiri:
> Looks like coverage pipeline is not running e2e lab api tests.

It now does, @shahzaibj kindly re-review

paulkagiri:
> What are all these jacoco agent properties files?

I believe this is already addressed - that was a first attempt at fixing the issue which has already been done away with

msal/src/androidTest/resources/jacoco-agent.properties
```diff
@@ -0,0 +1 @@
+output=none
```
p3dr0rv, line 1: Nit: add extra new-line at EOF

p3dr0rv (approved):

.github/workflows/codecov.yml
shahzaibj: Does the report include coverage for code in common? Doesn't look like it. I'm asking because E2E tests in MSAL provide a lot of coverage for the code in common lib

paulkagiri: Common has its own code coverage pipeline which targets tests within it so is there still need to target the same in MSAL?

shahzaibj: We need to see what sort of coverage for common code are we getting for MSAL tests...so to answer your question, yes we need to do it.

shahzaibj (approved):
I'm going to approve this since it solves the original problem. 

I am however recommending that we include coverage for common code from MSAL tests (this was done when we initially implemented code coverage), however, you may choose to create a work item for it and handle it in a different PR.

shoatman (approved):

PR #1559, opened 2021-11-24
Msal get account crash in Account Matcher due to NPE because account id was missing

shahzaibj:
To resolve, fallback to set MSAL account id from homeAccountId in CacheRecord if …
…not present anywhere else.

Depends on common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1642

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/Account.java
+++ b/msal/src/main/java/com/microsoft/identity/client/Account.java
@@ -46,6 +46,7 @@ public class Account implements IAccount {
     private String mHomeOid;
     private String mHomeTenantId;
     private String mEnvironment;
+    private String mHomeAccountId;
 
     public Account(
             @Nullable final String clientInfo,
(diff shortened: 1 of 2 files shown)


msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -92,6 +94,24 @@ public String getId() {
...
+            id = com.microsoft.identity.common.java.util.StringUtil.getUIdFromHomeAccountId(mHomeAccountId);
+        }
+
+        if (StringUtil.isEmpty(id)) {
+            com.microsoft.identity.common.logging.Logger.warn(
+                    TAG + methodName,
+                    "Account ID is empty. Returning MISSING_FROM_THE_TOKEN_RESPONSE."
+            );
+
+            // must return something because the method is marked as non-null
```
AdamBJohnsonx, line 111: I think we should reconsider the fact that this method id non-null.  We should either accept the null return as a possibility, or deal with an exception, perhaps.

PR #1560, opened 2021-11-24
Fix msal get account crash in Account Matcher due to NPE because account id was missing

shahzaibj:
To resolve, fallback to set MSAL account id from homeAccountId in CacheRecord if …
…not present anywhere else.

Depends on common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1643

Complete description of issue available here: https://github.com/AzureAD/ad-accounts-for-android/pull/1785#issue-1062821554

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/Account.java
+++ b/msal/src/main/java/com/microsoft/identity/client/Account.java
@@ -95,6 +95,8 @@ public class Account implements IAccount {
         }
 
         if (StringUtil.isEmpty(id)) {
+            // This could happen because the ID token that we have for WPJ accounts may not contain
+            // oid claim because we used to not ask for PROFILE scope in that token. 
             com.microsoft.identity.common.logging.Logger.warn(
                     TAG + methodName,
                     "Unable to get account id from either ClientInfo or IdToken. Attempting to obtain from home account id."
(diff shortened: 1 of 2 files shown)


msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -92,6 +94,24 @@ public String getId() {
             id = mHomeOid;
         }
 
+        if (StringUtil.isEmpty(id)) {
+            com.microsoft.identity.common.logging.Logger.warn(
```
rpdome, line 98: can you elaborate more on why this happens?  Is the issue consistently reproductible?

rpdome: ah okay... this is a mitigation for this broker bug.

`Secondly, the ID token that we have for WPJ accounts may not contain oid claim, therefore, I've updated the PRT request to ask for PROFILE scope as well to ensure we get oid claim in ID Token. (this issue is in both dev and PROD`

rpdome: Could you please add a comment describing the issue above this line? Just so that future devs have a context.

rpdome (approved):

msal/src/main/java/com/microsoft/identity/client/Account.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1560/commits/254d3c2ce980c03b384fbf51d4009e99974d2bae

PR #1563, opened 2021-11-30
Fixes for Msal Broker UI Automation

iamgusain:
### What 
- Changing Germany to US cloud 
- Update MsalLogging rule to ignore the exception if external logger is set more than once. to avoid the automation app to crash

### Testing
Automation Run with these changes (along with changes in this [PR](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1645): 
https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=850900&view=results
Automation run before these changes : 
https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=850275&view=results

### DIFF
--- a/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/MsalLoggingRule.java
+++ b/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/MsalLoggingRule.java
@@ -88,14 +88,18 @@ public class MsalLoggingRule implements TestRule {
         final FileAppender msalFileLogAppender = new FileAppender(msalLogFileName, new LogcatLikeFormatter());
         Logger.getInstance().setLogLevel(Logger.LogLevel.VERBOSE);
         Logger.getInstance().setEnableLogcatLog(false);
-        Logger.getInstance().setExternalLogger(new ILoggerCallback() {
-            @Override
-            public void log(final String tag, final Logger.LogLevel logLevel,
-                            final String message, boolean containsPII) {
-                final LogLevel level = convertMsalLogLevelToInternalLogLevel(logLevel);
-                msalFileLogAppender.append(level, tag, message, null);
-            }
-        });
+        try {
+            Logger.getInstance().setExternalLogger(new ILoggerCallback() {
+                @Override
+                public void log(final String tag, final Logger.LogLevel logLevel,
+                                final String message, boolean containsPII) {
+                    final LogLevel level = convertMsalLogLevelToInternalLogLevel(logLevel);
+                    msalFileLogAppender.append(level, tag, message, null);
+                }
+            });
+        } catch (IllegalStateException ex) {
+            // External logger is already set,
+        }
 
         return msalFileLogAppender;
     }
--- a/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/MsalLoggingRule.java
+++ b/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/MsalLoggingRule.java
@@ -97,8 +97,9 @@ public class MsalLoggingRule implements TestRule {
                     msalFileLogAppender.append(level, tag, message, null);
                 }
             });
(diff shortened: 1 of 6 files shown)


shahzaibj:
The `MsalLoggingRule` here has nothing to do with Authenticator. This rule is for the automation app itself. We need to get MSAL logs from automation app that's consuming MSAL.

shahzaibj (changes requested):
Removal of MsalLoggingRule should do nothing to Authenticator as this logging rule is solely for automation app. With that said, I do know that resetting the external logger on the automation app can cause a crash in the automation app itself (not Authenticator). To mitigate this, I suggest we refactor the MsalLoggingRule to only set external logger if one was not already set (rather than removing logger altogether because that would mean we lose logs during automation).

iamgusain:
> Removal of MsalLoggingRule should do nothing to Authenticator as this logging rule is solely for automation app. With that said, I do know that resetting the external logger on the automation app can cause a crash in the automation app itself (not Authenticator). To mitigate this, I suggest we refactor the MsalLoggingRule to only set external logger if one was not already set (rather than removing logger altogether because that would mean we lose logs during automation).

Done,

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/TestCase850455.java
```diff
@@ -114,7 +114,7 @@ public String getTempUserType() {
 
     @Override
     public String getAuthority() {
-        return "https://login.microsoftonline.de/common";
+        return "https://login.microsoftonline.us/common";
```
AdamBJohnsonx, line 117: Is it possible to be getting this from the query we're running, rather than code the value?  Seems like it would be safer.

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/MsalLoggingRule.java
```diff
@@ -88,14 +88,18 @@ private FileAppender turnOnMsalLogging(final Description description) throws IOE
...
+            Logger.getInstance().setExternalLogger(new ILoggerCallback() {
+                @Override
+                public void log(final String tag, final Logger.LogLevel logLevel,
+                                final String message, boolean containsPII) {
+                    final LogLevel level = convertMsalLogLevelToInternalLogLevel(logLevel);
+                    msalFileLogAppender.append(level, tag, message, null);
+                }
+            });
+        } catch (IllegalStateException ex) {
+            // External logger is already set,
```
shahzaibj, line 101: Let's add a `System.out.printlin(ex.getMessage())` here

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/MsalLoggingRule.java
```diff
@@ -88,14 +88,18 @@ private FileAppender turnOnMsalLogging(final Description description) throws IOE
...
+        try {
+            Logger.getInstance().setExternalLogger(new ILoggerCallback() {
+                @Override
+                public void log(final String tag, final Logger.LogLevel logLevel,
+                                final String message, boolean containsPII) {
+                    final LogLevel level = convertMsalLogLevelToInternalLogLevel(logLevel);
+                    msalFileLogAppender.append(level, tag, message, null);
+                }
+            });
+        } catch (IllegalStateException ex) {
```
shahzaibj, line 100:
```suggestion
        } catch (final IllegalStateException ex) {
```

shahzaibj (approved):

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/MsalLoggingRule.java
AdamBJohnsonx, line 101: Log something?

AdamBJohnsonx (approved):

PR #1564, opened 2021-11-30
Add ability to provide common version using project property

iamgusain:
Add ability to provide common version using project property
Related PR on broker: https://github.com/AzureAD/ad-accounts-for-android/pull/1788

### DIFF
--- a/msal/build.gradle
+++ b/msal/build.gradle
@@ -169,6 +169,8 @@ task sourcesJar(type: Jar) {
     destinationDirectory = reporting.file("$project.buildDir/outputs/jar/")
 }
 
+String commonVersion = project.hasProperty("distCommonVersion") ? project.distCommonVersion : "0.0.+"
+
 dependencies {
     testImplementation project(path: ':common4j')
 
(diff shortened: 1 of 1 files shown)


shahzaibj:
So even during releases we have to now supply version from command line?

iamgusain:
> So even during releases we have to now supply version from command line?

For release branches we edit version "0.0.+" to the latest RC or final release version just to make it clear, same way as we do today
Command line is to mainly add support daily/dev or private builds

AdamBJohnsonx:
> > So even during releases we have to now supply version from command line?
> 
> For release branches we edit version "0.0.+" to the latest RC or final release version just to make it clear, same way as we do today Command line is to mainly add support daily/dev or private builds

Would be a nice thing to be able to automate that portion of the build.  I suppose we could use a GitHub action to edit the file in-place and check it in?

AdamBJohnsonx (approved):
Sure?  It's editing the property in once place instead of two...

shahzaibj (approved):

PR #1576, opened 2022-01-04
Added overloads to use List<String> in AcquireToken

fadidurah:
Added changes to address this bug https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20Android/Backlog%20items/?workitem=1104317. This is to make AcquireToken call consistent with other parts of the API. Also did this for AcquireTokenWithDeviceCode. Should this also be done for Sign-in methods in SingleAccountPCA and AcquireTokenSilent in both SingleAccountPCA and MultipleAccountPCA?

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/IMultipleAccountPublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/IMultipleAccountPublicClientApplication.java
@@ -110,6 +110,28 @@ public interface IMultipleAccountPublicClientApplication extends IPublicClientAp
      *                  {@link AuthenticationCallback#onError(MsalException)}.
      */
     void acquireToken(@NonNull final Activity activity,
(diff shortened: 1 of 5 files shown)


AdamBJohnsonx (commented):
Any reason not to weaken the contract to Iterable<String>?

shahzaibj:
Do we want to deprecate the APIs that use Arrays?

fadidurah:
Adam: I'm not entirely sure what the initial reason for using List was. The github issue for this was pointing out that the API was a bit inconsistent in how it asked for scopes, so this change is to let AT match the rest of the methods. We could follow up and consider if we want to change the type of the scope parameter type for all API methods.

Shazaib: That could make sense. I assume that if we went with that, we would want to include the actual logic of AT in this new List<String> method instead of just converting the scope parameter into a string array and calling the old method, which would deprecate.

PR #1604, opened 2022-03-16
Split MSAL (Global) and PCA-Specific Configuration

fadidurah:
**What**
Adding support for global configuration, where developers designate global fields that they'd want all PCA instances to have through a global configuration file. Refer to the design document PR below.
https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/4763

Relies on: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1726

| Global Fields   |      PCA-Specific      |  Unexposed (Default File) |
|-----------------|------------------------|---------------------------|
|minimum_required_broker_protocol_version|client_id|environment|
|web_view_zoom_controls_enabled|redirect_uri|http<br>- read_timeout<br>- connect_timeout|
|web_view_zoom_enabled|authorization_user_agent||
|power_opt_check_for_network_request_enabled|multiple_clouds_supported||
|handle_null_taskaffinity|broker_redirect_uri_registered||
|authorization_in_current_task|account_mode||
|telemetry|client_capabilities||
|logging<br>- pii_enabled<br>- log_level<br>- logcat_enabled|authorities||
|browser_safe_list<br>- browser_package_name<br>- browser_signature_hashes<br>- browser_version_lower_bound<br>- browser_version_upper_bound|||

**Why**
At the moment, every `PublicClientApplication` (PCA) needs a configuration file to explicitly set all of its configuration fields, including fields that are not designed to be different across multiple PCA instances within the same application. A developer would need to create a different configuration file if they wanted to exclude a field from a PCA's configuration, and reapply whatever fields they wanted to keep.

**How**
Creating a `GlobalSettings` class used to initialize the global configuration from a global configuration file that the developer passes in as a parameter. In this design, the global settings initialization is done before the PCA is created in the developer's app. Once initialization is done, the global settings will be available as a static singleton field, which can be used later by PCA instances to supply global fields as command parameters. These settings are called "Global" because as mentioned above, fields included in these files are not meant to be changed between different PCA instances within the same session.

**Testing**
Added unit testing in this PR through `GlobalSettingsPublicClientApplicationTest` to check current implementation is not broken and new global field reading is being sent to PCAConfiguration correctly. Testing in common (see other PR).


changelog
```diff
@@ -2,6 +2,7 @@ MSAL Wiki : https://github.com/AzureAD/microsoft-authentication-library-for-andr
 
 vNext
 ----------
+- [MAJOR] Adding support for Global Settings.
```
rpdome, line 5: nit - probably should be more detailed.

rpdome: nit - probably should be more detailed.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
```diff
@@ -0,0 +1,173 @@
...
+
+import java.io.File;
+
+/**
+ * Class used to initialize global configurations for the library.
+ */
+public class GlobalSettings {
+    private static final String TAG = GlobalSettings.class.getSimpleName();
+    public static final String NO_GLOBAL_SETTINGS_WARNING = "Global settings have not been initialized before the creation of this PCA Configuration.";
+    public static final String GLOBAL_INIT_AFTER_PCA = "pca_created_before_global";
```
rpdome: nit: GLOBAL_INIT_AFTER_PCA_ERROR_CODE

rpdome, line 47: nit: GLOBAL_INIT_AFTER_PCA_ERROR_CODE

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
```diff
@@ -0,0 +1,173 @@
...
+     * @param context Context of the app.
+     * @param configFileResourceId Resource Id for the configuration file.
+     * @param listener Handles success and error messages.
+     */
+    public static void loadGlobalConfigurationFile(@NonNull final Context context,
+                                                   final int configFileResourceId,
+                                                   @NonNull final GlobalSettingsListener listener) {
+        validateNonNullArgument(context, PublicClientApplication.NONNULL_CONSTANTS.CONTEXT);
+        validateNonNullArgument(listener, PublicClientApplication.NONNULL_CONSTANTS.LISTENER);
+        runOnBackground(new Runnable() {
```
rpdome: we probably want to lock when the method starts, and unlock when the background task finish.

rpdome, line 67: we probably want to lock when the method starts, and unlock when the background task finish.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
```diff
@@ -0,0 +1,173 @@
...
+public class GlobalSettings {
+    private static final String TAG = GlobalSettings.class.getSimpleName();
+    public static final String NO_GLOBAL_SETTINGS_WARNING = "Global settings have not been initialized before the creation of this PCA Configuration.";
+    public static final String GLOBAL_INIT_AFTER_PCA = "pca_created_before_global";
+    public static final String GLOBAL_INIT_AFTER_PCA_ERROR_MESSAGE = "Global initialization was attempted after a PublicClientApplicationConfiguration instance was already created. Please initialize global settings before any PublicClientApplicationConfiguration instance is created.";
+
+    private static GlobalSettingsConfiguration mGlobalSettingsConfiguration;
+    private static boolean mPCAConfigCreated = false;
+    private static boolean mGlobalSettingsInitialized = false;
+    private static Object mGlobalSettingsLock = new Object();
```
rpdome: ReadWriteLock?

rpdome: ReadWriteLock?

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
```diff
@@ -0,0 +1,173 @@
...
+            );
+
+            if (config.getAuthorities().size() > 1) {
+                throw new MsalClientException(SAPCA_USE_WITH_MULTI_POLICY_B2C);
+            }
+        }
+    }
+
+    protected static GlobalSettingsConfiguration getGlobalSettingsConfiguration() {
+        return mGlobalSettingsConfiguration;
```
rpdome: this should also be protected by the lock.

rpdome, line 143: this should also be protected by the lock.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
```diff
@@ -0,0 +1,173 @@
...
+import com.microsoft.identity.client.exception.MsalException;
+import com.microsoft.identity.common.java.authorities.AzureActiveDirectoryB2CAuthority;
+import com.microsoft.identity.common.logging.Logger;
+
+import java.io.File;
+
+/**
+ * Class used to initialize global configurations for the library.
+ */
+public class GlobalSettings {
```
rpdome, line 44: fyi: In general, i would prefer having a Singleton over a static class, since the former would allow me to test the class in isolation (and also allowing me to inject mock data/states, etc).

rpdome: fyi: In general, i would prefer having a Singleton over a static class, since the former would allow me to test the class in isolation 

This also allows me to inject mock data/states. i.e. if I want to test "READ", i could just inject a valid data and test it - without having to call "WRITE", because if you do that and WRITE is buggy, then your READ test will fail too.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
```diff
@@ -0,0 +1,173 @@
...
+     * Load the global configuration file using the context, resource id of the configuration file, and a listener.
+     *
+     * @param context Context of the app.
+     * @param configFileResourceId Resource Id for the configuration file.
+     * @param listener Handles success and error messages.
+     */
+    public static void loadGlobalConfigurationFile(@NonNull final Context context,
+                                                   final int configFileResourceId,
+                                                   @NonNull final GlobalSettingsListener listener) {
+        validateNonNullArgument(context, PublicClientApplication.NONNULL_CONSTANTS.CONTEXT);
```
rpdome: We probably don't need this... as Lombok NonNull will already be throwing an exception if a null object is passed.

(This method predates our lombok integration)

rpdome, line 65: We probably don't need this... as Lombok NonNull will already be throwing an exception if a null object is passed.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    private static void validateAccountModeConfiguration(@NonNull final GlobalSettingsConfiguration config) throws MsalClientException {
+        if (config.getAccountMode() == AccountMode.SINGLE
+                && null != config.getDefaultAuthority()
+                && config.getDefaultAuthority() instanceof AzureActiveDirectoryB2CAuthority) {
+            Logger.warn(
+                    TAG,
+                    "Warning! B2C applications should use MultipleAccountPublicClientApplication. "
+                            + "Use of SingleAccount mode with multiple IEF policies is unsupported."
```
rpdome: IEF?

msal/src/main/java/com/microsoft/identity/client/GlobalSettingsConfiguration.java
```diff
@@ -0,0 +1,675 @@
...
+    public String getRedirectUri() {
+        return this.mRedirectUri;
+    }
+
+    /**
+     * Sets the configured redirect uri for the PublicClientApplication.
+     *
+     * @param redirectUri The redirectUri to use.
+     */
+    public void setRedirectUri(@NonNull final String redirectUri) {
```
rpdome: you can use lombok's 

```Java
@Getter
@Setter
@Accessors(prefix = "m")
```

rpdome, line 254: you can use lombok's @ getter and @ setter

msal/src/main/java/com/microsoft/identity/client/GlobalSettingsConfiguration.java
```diff
@@ -0,0 +1,675 @@
...
+//            final boolean hasAuthority = !TextUtils.isEmpty(parsedRedirectUri.getAuthority());
+//            return hasScheme && hasAuthority;
+//        } catch (final NullPointerException e) {
+//            com.microsoft.identity.common.internal.logging.Logger.errorPII(TAG, INVALID_REDIRECT_MSG, e);
+//        }
+//
+//        return false;
+//    }
+
+//    private void validateAzureActiveDirectoryAuthority(@NonNull final AzureActiveDirectoryAuthority azureActiveDirectoryAuthority) {
```
rpdome: Do we still need these commented out code?

rpdome: Do we still need these commented out code?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfigurationFactory.java
```diff
@@ -93,12 +99,27 @@ private static PublicClientApplicationConfiguration initializeConfigurationInter
...
+
+    @WorkerThread
+    private static PublicClientApplicationConfiguration mergeConfigurationWithGlobal(final @NonNull PublicClientApplicationConfiguration developerConfig) {
+        if (!GlobalSettings.isGlobalSettingsInitialized()) {
+            Logger.warn(TAG + "mergeConfigurationWithGlobal",
+                    GlobalSettings.NO_GLOBAL_SETTINGS_WARNING);
+            return developerConfig;
+        }
+
+        developerConfig.mergeGlobalConfiguration(GlobalSettings.getGlobalSettingsConfiguration());
```
rpdome: suggestion: put `mergeGlobalConfiguration` (or this method) into `GlobalSettings`, and protect the method with a read lock. 

By doing this, we wouldn't have to expose the lock.

rpdome, line 120: suggestion: put `mergeGlobalConfiguration` (or this method) into `GlobalSettings`, and protect the method with a read lock. 

By doing this, we wouldn't have to expose the lock.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * Class used to initialize global configurations for the library.
+ */
+public class GlobalSettings {
+    private static final String TAG = GlobalSettings.class.getSimpleName();
+    public static final String NO_GLOBAL_SETTINGS_WARNING = "Global settings have not been initialized before the creation of this PCA Configuration.";
+    public static final String GLOBAL_INIT_AFTER_PCA = "pca_created_before_global";
+    public static final String GLOBAL_INIT_AFTER_PCA_ERROR_MESSAGE = "Global initialization was attempted after a PublicClientApplicationConfiguration instance was already created. Please initialize global settings before any PublicClientApplicationConfiguration instance is created.";
+
+    private static GlobalSettingsConfiguration mGlobalSettingsConfiguration;
```
rpdome: nit: javadoc for all variable - describing what they are and what do we need them for?

rpdome, line 50: nit: javadoc for all variable - describing what they are and what do we need them for?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfigurationFactory.java
rpdome: (and we can set mPCAConfigCreated in that method without having to worry about race condition)

rpdome (commented):
Left some suggestions. 

Please also add some unit tests, esp around the set up orders/potential race condition.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
fadidurah: You mean the name of the lock variable right?

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
fadidurah: Oh good to know. I thought it was an extra layer of security we were still using.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfigurationFactory.java
fadidurah: This would make sense, but I'd have to add setter methods in PublicClientApplicationConfiguration for all the fields that could be merged, since we wouldn't have direct access to the fields in the class.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
fadidurah: I'm not really sure abouyt this one. I copied this from PublicClientApplication. We may not need this check in global? I think it might be getting checked automatically in PCA anyway, after GlobalConfig is merged with PCAConfig.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * Load the global configuration file using the context, resource id of the configuration file, and a listener.
+     *
+     * @param context Context of the app.
+     * @param configFileResourceId Resource Id for the configuration file.
+     * @param listener Handles success and error messages.
+     */
+    @WorkerThread
+    public static void loadGlobalConfigurationFile(@NonNull final Context context,
```
rpdome: I think all static methods in this class can be made non-static.

We can still use the singleton pattern - the caller can access this global object via a "getInstance()" function. 

If we want extra testability for this class alone, we could add 
- a "protected" constructor , which we can use in test classes belonging in the same package. 
- a "static protected setInstance()" method that allows swapping in a new GlobalSettings object during tests.

(don't forget to make the class "final" - so that the protected constructor isn't misused).

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -480,6 +481,39 @@ void mergeConfiguration(PublicClientApplicationConfiguration config) {
         this.isAuthorizationInCurrentTask = config.isAuthorizationInCurrentTask == null ? this.isAuthorizationInCurrentTask : config.isAuthorizationInCurrentTask;
     }
 
+    void mergeGlobalConfiguration(GlobalSettingsConfiguration globalConfig) {
```
rpdome: nit: add private/final/nonnull

msal/src/test/java/com/microsoft/identity/client/GlobalSettingsTest.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    @Test
+    public void testGlobalFieldsOverrideSingleAccountPCAFields() throws InterruptedException {
+        final File globalConfigFile = new File(TEST_SPLIT_CONFIG_GLOBAL_CONFIG_FILE_PATH);
+        GlobalSettings.loadGlobalConfigurationFile(globalConfigFile, getSuccessGlobalListener());
+
+        CountDownLatch countDownLatch = new CountDownLatch(1);
+        createSAPCAFromConfigPath(SINGLE_ACCOUNT_MODE_AAD_CONFIG_FILE_PATH, countDownLatch);
+        countDownLatch.await();
```
rpdome: Use await with the timeout. Set timeout to maybe 5 seconds.

msal/src/test/java/com/microsoft/identity/client/GlobalSettingsTest.java
```diff
@@ -0,0 +1,300 @@
...
+        final File configFile = new File(configPath);
+        PublicClientApplication.createSingleAccountPublicClientApplication(mContext, configFile, new IPublicClientApplication.ISingleAccountApplicationCreatedListener() {
+            @Override
+            public void onCreated(ISingleAccountPublicClientApplication application) {
+                mSingleAccountApplication = application;
+                countDownLatch.countDown();
+            }
+            @Override
+            public void onError(MsalException exception) {
+                Assert.fail(exception.getMessage());
```
rpdome: you want to also countdown here too..

msal/src/test/java/com/microsoft/identity/client/GlobalSettingsTest.java
```diff
@@ -0,0 +1,300 @@
...
+            }
+        });
+
+        flushScheduler();
```
rpdome: still surprises me why we would need this as we start using countDownLatch.. @shahzaibj  any idea?

rpdome (approved):
left a few suggestions.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
fadidurah: I had 2 reasons for making `loadGlobalConfigurqationFile()` methods static:

First reason was to match the structure we have for creating PCA objects (PublicClientApplication.createSingleAccountPublicClientApplication).

Second reason was to limit access to the static instance, keep it internal to the our library. I felt this would be more secure than allowing developers to directly use the instance.

msal/src/test/java/com/microsoft/identity/client/GlobalSettingsTest.java
fadidurah: Yes, im still confused why I needed to use flushScheluder() here, if I don't have it then the countDownLatch.await() hangs forever. I don't think I tried it with countDownLatch.await(TIMEOUT), though.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
fadidurah: I made this protected, since it's being called in `GlobalSettings.java`

msal/src/test/java/com/microsoft/identity/client/GlobalSettingsTest.java
fadidurah: I tried this again with the timeout, getting this error:

`Main looper has queued unexecuted runnables. This might be the cause of the test failure. You might need a shadowOf(getMainLooper()).idle() call.`

This was appearing in an earlier PR for me, the only solution that worked for me in that case was adding back the flushScheduler().

shahzaibj: In my experience, countdown latches don't work with Robolectric at all and for that reason I've never used them. `flushScheduler` should be the way to go.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
```diff
@@ -0,0 +1,194 @@
...
+            developerConfig.validateConfiguration();
+            return developerConfig;
+        }
+    }
+
+    protected static GlobalSettings getInstance() {
+        return mGlobalSettingsSingleton;
+    }
+
+    protected static void resetInstance() {
```
shahzaibj: Why do we need this? Javadoc?

shahzaibj: Why do we need this? Javadoc?

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
```diff
@@ -0,0 +1,194 @@
...
+                return developerConfig;
+            }
+
+            developerConfig.mergeGlobalConfiguration(mGlobalSettingsSingleton.mGlobalSettingsConfiguration);
+            developerConfig.validateConfiguration();
+            return developerConfig;
+        }
+    }
+
+    protected static GlobalSettings getInstance() {
```
shahzaibj: Why protected? I think it can be public.

shahzaibj: Why protected? I think it can be public.

msal/src/main/java/com/microsoft/identity/client/GlobalSettingsConfiguration.java
```diff
@@ -0,0 +1,298 @@
...
+    @Setter
+    @SerializedName(WEB_VIEW_ZOOM_ENABLED)
+    private Boolean webViewZoomEnabled;
+
+    @Setter
+    @SerializedName(POWER_OPT_CHECK_FOR_NETWORK_REQUEST_ENABLED)
+    private Boolean powerOptCheckEnabled;
+
+    @SerializedName(HANDLE_TASKS_WITH_NULL_TASKAFFINITY)
+    private Boolean handleNullTaskAffinity;
```
shahzaibj: It seems like pretty much everything is becoming a Global Setting field. I don't think we want that. What's the reason behind things like `clientId` and `redirectUri` to be a global setting?

shahzaibj: It seems like pretty much everything is becoming a Global Setting field. I don't think we want that. What's the reason behind things like `clientId` and `redirectUri` to be a global setting?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfigurationFactory.java
```diff
@@ -93,12 +95,14 @@ private static PublicClientApplicationConfiguration initializeConfigurationInter
             config.validateConfiguration();
         }
 
+        final PublicClientApplicationConfiguration configWithGlobal = globalSettings.mergeConfigurationWithGlobal(config);
+
         //Initialize internal library configuration
-        final LibraryConfiguration libraryConfiguration = LibraryConfiguration.builder().authorizationInCurrentTask((config.authorizationInCurrentTask())).build();
+        final LibraryConfiguration libraryConfiguration = LibraryConfiguration.builder().authorizationInCurrentTask((configWithGlobal.authorizationInCurrentTask())).build();
```
shahzaibj: I believe the motivation behind [that work item](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1400670) was that GlobalSettings would replace the `LibraryConfiguration`.

shahzaibj: I believe the motivation behind [that work item](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1400670) was that GlobalSettings would replace the `LibraryConfiguration`.

msal/src/test/java/com/microsoft/identity/client/GlobalSettingsTest.java
```diff
@@ -0,0 +1,302 @@
+package com.microsoft.identity.client;
```
shahzaibj: Missing LICENSE

shahzaibj: Missing LICENSE

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
fadidurah: Should this be accessible to developers? I can't think of a reason why they should have access to it.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
fadidurah: This is a method i added to reset the static singleton instance during testing

fadidurah: And yes i'll add javadoc, good catch

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfigurationFactory.java
fadidurah: Hmmm, I actually didn't know this is what we wanted. I can take a look at that class.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
shahzaibj: I mean my point is that it probably needs to be either `public` or `package-private` or `private`. The one we pick depends on how we want the API to be accessed. `protected` is usually used if we intend to allow usage of the API by any classes that extend this class. I don't think there are currently any children of `GlobalSettings` so I'm not sure why we would want `protected`.

msal/src/main/java/com/microsoft/identity/client/GlobalSettingsConfiguration.java
fadidurah: Me and Shane started a discussion about this. I created a page in OneNote to decide which fields we would want to configure either in global, pca config, or the default msal config. We can set up some time today or tomorrow to discuss where we want to allow developers to configure fields.

msal/src/main/java/com/microsoft/identity/client/GlobalSettings.java
fadidurah: Ah! I see what you mean, I'm removing the access modifier, I really only want this usable in package, and I didn't have plans to make subclasses of GlobalSettings.

PR #1610, opened 2022-03-28
Single Tenant Application Test

paulokagiri:
Add test coverage for Single Tenant Application

**Why?:**
We recently introduced a bug in the broker that broke only single-tenant application.
Our test plans/test apps doesn't have coverage on it, otherwise it should have been caught.

[Single and multi-tenant apps in Azure AD - Microsoft identity platform | Microsoft Docs](https://docs.microsoft.com/en-us/azure/active-directory/develop/single-and-multi-tenant-apps)


shahzaibj:
I'm not sure we need this right now. Going to close this for now. cc/ @fadidurah

PR #1618, opened 2022-05-02
added keyboard to configChanges

melissaahn:
### Links to Dependent PRs
- Main PR in Common: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1729
- Broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1869
- Adding YubiKit Version to versions.gradle in android complete: https://github.com/AzureAD/android-complete/pull/135
## The Problem
Note: For a more detailed description of all changes being made to set up the smartcard EO work, please see the [main PR in common](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1729).

When the YubiKey was plugged in, the host app, common, and broker code reacted adversely. While trying to authenticate via WebView, the WebView would seem to refresh upon YubiKey plug in, and this would result in an authentication flow that didn't seem to complete despite providing a correct upn and certificate/password. Additionally, if the YubiKey was plugged in or removed while in AzureSample's fragments, the app would crash.

## Why This Was Happening
The issue was due to Android recognizing a YubiKey device as a keyboard, which I suspect has to do with the [YubiKey's Touch Triggered OTP feature](https://developers.yubico.com/Developer_Program/Guides/Touch_triggered_OTP.html#:~:text=The%20YubiKey%20has%20an%20integrated,input%20field%20or%20command%20prompt.). Android sees an external keyboard being plugged in as a configuration change [`configChanges`](https://developer.android.com/guide/topics/manifest/activity-element#config). The default behavior for `configChanges` is to restart the current Activity. This behavior checks out with the observed WebView behavior. In the case of the host app crashing, the account Spinner was attempting to reference a null context before the current fragment could get reattached.

## Changes Made
In all Manifests with Activities that could be exposed to a plugged in YubiKey, `keyboard` and `keyboardHidden` need to be present in the `configChanges` for each activity listed. This overrides the default behavior and allows the activities themselves to handle the configuration change (in this case, we do nothing).


msal/src/main/AndroidManifest.xml
```diff
@@ -9,20 +9,20 @@
...
-            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout"
+            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout|keyboard"
             android:exported="false"
             android:launchMode="standard" />
 
         <!-- Helper activity for displaying current broker redirect URI configuration -->
         <activity
             android:name="com.microsoft.identity.client.helper.BrokerHelperActivity"
-            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout"
+            android:configChanges="orientation|keyboardHidden|screenSize|smallestScreenSize|screenLayout|keyboard"
```
shahzaibj: Why do we need to add this to BrokerHelperActivity? Adding is probably harmless but I'm curios if it is actually needed?

shahzaibj: I mean I think we are not discovering USB devices on this activity, right?

melissaahn: I'll double check if the flag is needed for this activity

melissaahn: Based on my testing, wasn't needed. Removed it.

rpdome (approved):

shahzaibj (approved):

p3dr0rv (approved):

PR #1619, opened 2022-05-05
Shahzaibj/msal automation fixes

shahzaibj:


msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/crosscloud/TestCase1420484.java
```diff
@@ -83,7 +83,7 @@ public void test_1420484() throws Throwable {
                     PromptHandlerParameters.builder()
                     .prompt(PromptParameter.SELECT_ACCOUNT)
                     .loginHint(userName)
-                    .staySignedInPageExpected(true)
+                    .staySignedInPageExpected(false)
```
fadidurah, line 86: I've been seeing the stay signed in page still popping up on a lot of test cases. are we not expecting the stay signed in to come up anymore?

shahzaibj: It wasn't coming up when I ran these a month ago. May be ESTS keep changing the logic or may be the lab keeps changing something in the directory. I suggest let's create a global flag that indicates `IS_STAY_SIGN_IN_PAGE_EXPECTED`

and then let's pass that value to each of these `.staySignedInPageExpected()` in all the test cases. This way we can change the value in one place going forward as opposed to so many places.

fadidurah: Yes, I think that's a good move

shahzaibj:
This got merged in as part of other work/PRs created by Fadi for the repair work.

PR #1620, opened 2022-05-09
UI Automation test case for [JOINED][MSAL] Password change scenario

somalaya:
**What**
UI automation testing needs to be added for this scenario -   User completes Workplace joine and the password is changed. Using MSAL, we fetch do a silent fetch which fails followed by interactive sign in which should succeed

**How**
Added a test case - TestCase1561151

**Testing**
Automation UI testing was done on Local Pixel 4 Emulator with API 29.
The test case in msalAutomationApp - TestCase1561151 was passed successfully.


somalaya:
Closing as the one test case changes in this PR are included in another open PR instead

PR #1624, opened 2022-05-18
removed keyboard from configchanges

melissaahn:
Removing Keyboard flags from Activity configChanges for now, as this will cause errors for the upcoming broker release until a major release of MSAL can be completed.

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -43,7 +43,6 @@ ext {
     uiAutomatorVersion = "2.2.0"
     mseberaApacheHttpClientVersion = "4.5.8"
     msal4jVersion = "1.10.0"
-    yubikitAndroidVersion = "2.0.0"
 
     // TODO: adal automation test app.
     supportLibraryVersion = "27.1.+"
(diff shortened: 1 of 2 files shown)


shahzaibj (approved):

rpdome (approved):
would you mind adding the details of the error for future reference?

melissaahn:
This PR was created ahead of time in order to address errors that could occur should the Authenticator team decide to upgrade MSAL without broker (before smartcard CBA can be fully integrated). 
Manifest errors were brought up when Authenticator last tested a new build for broker due to mismatching `configChanges` sections between Common in the new broker build and MSAL in the current MSAL dependency (which was not being updated at the time). The updated `configChanges` flags are not needed until smartcard CBA is ready for integration. Thus, the quickest solution to avoid blocking releases is to merge this branch into the release branch.
It should be noted that the YubiKit dependency will also need to be taken out of Common in the case of a new MSAL build, so a separate branch doing this should be created and merged into the Common release branch as well.

PR #1636, opened 2022-06-10
Non-joined Test Cases on MSAL Side

fadidurah:
**What**: Adding Ui automation test cases for broker non-joined account that use msal test app. Test Case 1561152 is for password reset, so removing the standalone password reset test.

**Testing**: This PR adds new test cases to msal automation.

### DIFF
--- a/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/TestCase1561158.java
+++ b/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/TestCase1561158.java
@@ -1,0 +1,59 @@
+//  Copyright (c) Microsoft Corporation.
+//  All rights reserved.
+//
+//  This code is licensed under the MIT License.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files(the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions :
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.client.msal.automationapp.testpass.broker;
+
+import com.microsoft.identity.labapi.utilities.client.LabQuery;
+import com.microsoft.identity.labapi.utilities.constants.TempUserType;
+
+import org.junit.Ignore;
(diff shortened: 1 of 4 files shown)


msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/TestCase1600592.java
```diff
@@ -79,25 +82,19 @@ public void handleUserInteraction() {
...
                 .resource(mScopes[0])
                 .msalConfigResourceId(getConfigFileResourceId())
                 .build();
 
         final MsalAuthResult authResult2 = msalSdk.acquireTokenSilent(silentParams, TokenRequestTimeout.MEDIUM);
 
-        // Should get a failure due to password reset.
         authResult2.assertFailure();
-        Assert.assertTrue(authResult2.getException() instanceof MsalUiRequiredException);
+        Assert.assertTrue(authResult2.getException() instanceof MsalServiceException);
```
shahzaibj, line 97: I think this test case is going to fail right now? that's what I recall from the manual testing done last week...if this is failing then you can add the ignore annotation to this one as well

shahzaibj (approved):
lgtm

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/TestCase1600592.java
fadidurah: I got it to pass in automation, the silent call is returning an MsalServiceException about the application not being set up as multi tenant. This is the exception returned by the silent call.

AADSTS50194: Application 'a8c8bb57-55f3-4e95-9cf2-a3aac32f0d7a'(IDLABS_App_No_Admin_Consent) is not configured as a multi-tenant application. Usage of the /common endpoint is not supported for such applications created after '10/15/2018'. Use a tenant-specific endpoint or configure the application to be multi-tenant.
Trace ID: 79c48634-ce46-4953-bac9-e6bdffa43e00
Correlation ID: f7e03b76-ba82-4743-bb66-ca41e3a2bba2
Timestamp: 2022-06-10 04:18:35Z

shahzaibj: That's great. that means that something was wrong then in manual execution. cool.

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/TestCase1561158.java
```diff
@@ -0,0 +1,59 @@
...
+import com.microsoft.identity.labapi.utilities.client.LabQuery;
+import com.microsoft.identity.labapi.utilities.constants.TempUserType;
+
+import org.junit.Ignore;
+import org.junit.Test;
+
+// [Non-joined][MSAL] Prompt.CONSENT
+// https://identitydivision.visualstudio.com/DevEx/_workitems/edit/1561158
+@Ignore("Broker ignore prompt.CONSENT, just proceeds with silent")
+public class TestCase1561158 extends AbstractMsalBrokerTest {
```
somalaya, line 35: Fadi, I am curious what it means to add this new testcase and mark it ignore?

somalaya (approved):

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/TestCase1561158.java
fadidurah: So this test case is actually the prompt.CONSENT test case in non-joined flow, which we've currently marked as non-applicable because the broker ignores prompt.CONSENT and just proceeds with silent. I added an Ignore to it so it wouldn't run, but I wanted to go ahead and create the class for it so that I could implement it once we add support for prompt.CONSENT.

somalaya: Got it. Can we add a TO-DO to associate it with this bug ? https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1923085/

fadidurah: Added a todo to the top of the test case.

PR #1637, opened 2022-06-13
Added automation tests for AT-PoP scenarios

somalaya:
What: Adding Ui automation test cases for AT-PoP scenarios under the test plan for AT-PoP  folder in the following link https://identitydivision.visualstudio.com/Engineering/_testPlans/execute?planId=1905195&suiteId=1905209

Testing: Tested by running these testcases on Pixel 4 emulator

Related PR : This PR consists of some utility functions added to support AT-PoP scenarios https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1756

### DIFF
--- a/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922511.java
+++ b/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922511.java
@@ -22,8 +22,6 @@
 //  THE SOFTWARE.
 package com.microsoft.identity.client.msal.automationapp.testpass.atpop;
 
-import androidx.annotation.NonNull;
-
 import com.microsoft.identity.client.Prompt;
 import com.microsoft.identity.client.msal.automationapp.R;
 import com.microsoft.identity.client.msal.automationapp.sdk.MsalAuthResult;
@@ -36,17 +34,13 @@ import com.microsoft.identity.client.ui.automation.interaction.OnInteractionRequ
 import com.microsoft.identity.client.ui.automation.interaction.PromptHandlerParameters;
 import com.microsoft.identity.client.ui.automation.interaction.PromptParameter;
 import com.microsoft.identity.client.ui.automation.interaction.microsoftsts.AadPromptHandler;
-import com.microsoft.identity.common.java.exception.ServiceException;
-import com.microsoft.identity.common.java.providers.oauth2.IDToken;
 import com.microsoft.identity.labapi.utilities.client.LabQuery;
 import com.microsoft.identity.labapi.utilities.constants.AzureEnvironment;
 import com.microsoft.identity.labapi.utilities.constants.TempUserType;
 
-import org.junit.Assert;
 import org.junit.Test;
 
 import java.util.Arrays;
-import java.util.Map;
 
 // [Non-Joined] Acquire PoP token interactive followed by Silent
 // https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1922511
@@ -86,7 +80,7 @@ public class TestCase1922511 extends AbstractMsalBrokerTest {
         }, TokenRequestTimeout.MEDIUM);
 
         authResult.assertSuccess();
-        Assert.assertEquals(MsalAuthResult.atPopSuccessMsg, authResult.verifyATForPop(authResult.getAccessToken()));
+        MsalAuthResult.verifyATForPop(authResult.getAccessToken());
 
         // start silent token request in MSAL
         final MsalAuthTestParams authTestSilentParams = MsalAuthTestParams.builder()
@@ -100,7 +94,7 @@ public class TestCase1922511 extends AbstractMsalBrokerTest {
 
         final MsalAuthResult authSilentResult = msalSdk.acquireTokenSilent(authTestSilentParams, TokenRequestTimeout.MEDIUM);
         authSilentResult.assertSuccess();
-        Assert.assertEquals(MsalAuthResult.atPopSuccessMsg, authSilentResult.verifyATForPop(authSilentResult.getAccessToken()));
+        MsalAuthResult.verifyATForPop(authSilentResult.getAccessToken());
     }
 
     @Override
--- a/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922511.java
+++ b/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922511.java
@@ -86,7 +86,7 @@ public class TestCase1922511 extends AbstractMsalBrokerTest {
         }, TokenRequestTimeout.MEDIUM);
 
         authResult.assertSuccess();
-        verifyATForPop(authResult);
+        Assert.assertEquals(MsalAuthResult.atPopSuccessMsg, authResult.verifyATForPop(authResult.getAccessToken()));
 
         // start silent token request in MSAL
         final MsalAuthTestParams authTestSilentParams = MsalAuthTestParams.builder()
@@ -100,30 +100,7 @@ public class TestCase1922511 extends AbstractMsalBrokerTest {
 
         final MsalAuthResult authSilentResult = msalSdk.acquireTokenSilent(authTestSilentParams, TokenRequestTimeout.MEDIUM);
         authSilentResult.assertSuccess();
-        verifyATForPop(authSilentResult);
-    }
-
-    private void verifyATForPop(@NonNull final MsalAuthResult authResult) throws ServiceException {
-        String rawIdToken = authResult.getAccessToken();
-        Map<String, ?> tokens = IDToken.parseJWT(rawIdToken);
-        // check for url
-        if (tokens.containsKey("u")) {
-            Assert.assertEquals("signedhttprequest.azurewebsites.net", tokens.get("u"));
-        } else {
-            Assert.fail("decoded AccessToken does not contain the PoPResourceUri");
-        }
-        // check for path
-        if (tokens.containsKey("p")) {
-            Assert.assertEquals("/api/validateSHR", tokens.get("p"));
-        } else {
-            Assert.fail("decoded AccessToken does not contain the PoPResource path");
-        }
-        // check for the http method
-        if (tokens.containsKey("m")) {
-            Assert.assertEquals("GET", tokens.get("m"));
-        } else {
-            Assert.fail("decoded AccessToken does not contain the specified HTTP method");
-        }
+        Assert.assertEquals(MsalAuthResult.atPopSuccessMsg, authSilentResult.verifyATForPop(authSilentResult.getAccessToken()));
     }
 
     @Override
@@ -153,3 +130,4 @@ public class TestCase1922511 extends AbstractMsalBrokerTest {
         return R.raw.msal_config_default;
     }
 }
+
--- a/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922511.java
+++ b/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922511.java
@@ -130,4 +130,3 @@ public class TestCase1922511 extends AbstractMsalBrokerTest {
         return R.raw.msal_config_default;
     }
 }
-
--- a/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922511.java
+++ b/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922511.java
@@ -1,0 +1,153 @@
+//  Copyright (c) Microsoft Corporation.
+//  All rights reserved.
+//
+//  This code is licensed under the MIT License.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files(the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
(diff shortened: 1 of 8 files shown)


msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922511.java
```diff
@@ -0,0 +1,155 @@
...
+                        .expectingBrokerAccountChooserActivity(false)
+                        .build();
+
+                new AadPromptHandler(promptHandlerParameters)
+                        .handlePrompt(username, password);
+            }
+        }, TokenRequestTimeout.MEDIUM);
+
+        authResult.assertSuccess();
+        verifyATForPop(authResult);
```
fadidurah: Could we place this method in one of the super classes so it doesn't have to be defined in each case?

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922527.java
```diff
@@ -0,0 +1,115 @@
...
+                        .broker(mBroker)
+                        .expectingBrokerAccountChooserActivity(false)
+                        .build();
+
+                new AadPromptHandler(promptHandlerParameters)
+                        .handlePrompt(username, password);
+            }
+        }, TokenRequestTimeout.MEDIUM);
+
+        authResult.assertSuccess();
```
fadidurah, line 83: Call verifyATForPop next?

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/atpop/TestCase1922515.java
```diff
@@ -0,0 +1,117 @@
...
+                        .broker(mBroker)
+                        .expectingBrokerAccountChooserActivity(false)
+                        .build();
+
+                new AadPromptHandler(promptHandlerParameters)
+                        .handlePrompt(username, password);
+            }
+        }, TokenRequestTimeout.MEDIUM);
+
+        authResult.assertSuccess();
```
fadidurah, line 85: Call verifyATForPop next?

fadidurah (approved):
Looks good, some suggestions left.

msalautomationapp/src/main/java/com/microsoft/identity/client/msal/automationapp/sdk/MsalSdk.java
```diff
@@ -85,6 +89,17 @@ public MsalAuthResult acquireTokenInteractive(@NonNull MsalAuthTestParams authTe
             acquireTokenParametersBuilder.withClaims(authTestParams.getClaims());
         }
 
+        if (authTestParams.getAuthScheme() == AuthScheme.POP) {
```
shahzaibj, line 95: I think this is fine for now. Ultimately we might want to able to supply different values for HTTP Method, claims, URL etc to this API but I understand that would require more changes. So what we've done here is good for now.

msalautomationapp/src/main/java/com/microsoft/identity/client/msal/automationapp/sdk/MsalSdk.java
```diff
@@ -85,6 +89,17 @@ public MsalAuthResult acquireTokenInteractive(@NonNull MsalAuthTestParams authTe
...
+            acquireTokenParametersBuilder.withAuthenticationScheme(
+                    PoPAuthenticationScheme.builder()
+                            .withHttpMethod(HttpMethod.valueOf("GET"))
+                            .withClientClaims("")
```
shahzaibj: We can probably remove this line altogether I believe (for setting client claims). You can try removing and if nothing breaks then keep it removed.

shahzaibj: This feedback applies below to silent as well

msalautomationapp/src/main/java/com/microsoft/identity/client/msal/automationapp/sdk/MsalSdk.java
```diff
@@ -268,4 +290,49 @@ private IAccount getAccountForPolicyName(@NonNull final MultipleAccountPublicCli
...
+            }
+
+            @Override
+            public void onError(MsalException exception) {
+                future.setException(exception);
+            }
+
+        };
+
+        if (authTestParams.getAuthScheme() == AuthScheme.POP) {
```
shahzaibj: I think we don't necessarily need to check this here at all. We can just proceed to generateSignedHttpRequest regardless of it because just calling the `generateShr` method implies that scheme is PoP

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922511.java
shahzaibj: +1 to that. I'm actually thinking that we may want to put this directly on the `MsalAuthResult` class and then we can just call it like `authResult.verifyAtForPop()`

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/atpop/TestCase1922515.java
shahzaibj: +1

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922527.java
```diff
@@ -0,0 +1,115 @@
...
+
+        String shr = msalSdk.generateSHR(authTestParams, TokenRequestTimeout.SHORT);
+        Assert.assertNotNull(shr);
```
shahzaibj, line 87: we should also verify the claims in the SHR

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/atpop/TestCase1922515.java
```diff
@@ -0,0 +1,117 @@
...
+
+        String shr = msalSdk.generateSHR(authTestParams, TokenRequestTimeout.SHORT);
+        Assert.assertNotNull(shr);
```
shahzaibj, line 89: we should also verify the claims in the SHR

shahzaibj (commented):
Left some comments. I would encourage looking at the feedback from Fadi

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922511.java
somalaya: done, added this method to MsalAuthResult. I did not use asserts here as we have to add the dependency in this pkg for this purpose. Hence verified the negation of each expectation and returned the final result as string. If the string is returned as "success" -> it means that the AT-POP was successfully verified.

shahzaibj: I'm not sure I follow...did you push the commit already? (can't see it in GitHub)

msalautomationapp/src/main/java/com/microsoft/identity/client/msal/automationapp/sdk/MsalAuthResult.java
```diff
@@ -42,4 +45,20 @@ public MsalAuthResult(@NonNull final IAuthenticationResult authenticationResult)
     public MsalAuthResult(@NonNull final Exception exception) {
         super(exception);
     }
+
+    public String verifyATForPop(@NonNull final String rawIdToken) throws ServiceException {
```
shahzaibj: Okay, you're verifying accessToken (NOT IdToken); and I don't think you need to even accept AT as a param here...since you have access to the AT in this class itself..you can just invoke the `getAccessToken` method directly in this method to get the AT

msalautomationapp/src/main/java/com/microsoft/identity/client/msal/automationapp/sdk/MsalAuthResult.java
```diff
@@ -42,4 +45,20 @@ public MsalAuthResult(@NonNull final IAuthenticationResult authenticationResult)
...
+        if (!tokens.get("u").equals("signedhttprequest.azurewebsites.net"))
+            return "Decoded AccessToken does not contain the PoPResourceUri";
+
+        if (!tokens.get("p").equals("/api/validateSHR"))
+            return "Decoded AccessToken does not contain the PoPResource path";
+
+        if (!tokens.get("m").equals("GET"))
+            return "Decoded AccessToken does not contain the expected HTTP method";
+
+        return "successfully verified at-pop";
```
shahzaibj: Why wouldn't we just use assertions?

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922511.java
shahzaibj: In regards to the dependency, I think you can put the verifyAtPop method directly in the parent class i.e. AuthResult and that seems to already have the assertion dependency

msalautomationapp/src/main/java/com/microsoft/identity/client/msal/automationapp/sdk/MsalAuthResult.java
somalaya: we do not have access to SHR in this class. This method would be verifying both the authResult.getAccessToken and SHR

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/atpop/TestCase1922511.java
somalaya: I think even if i do this, it would expect a dependency to be added to use jwt Parser (IDToken.parseJWT). I wanted to avoid adding it

msalautomationapp/src/main/java/com/microsoft/identity/client/msal/automationapp/sdk/MsalAuthResult.java
shahzaibj: Okay sounds good. In that case then you can keep as is, however, just make 2 changes:

- Make this method a `static` method as opposed to `instance` method
- Rename the param from `rawIdToken` to `shr`

and then it should be good to be used for both AT / SHR. 

Also feel free to add the dependency for assertions to the automation app

somalaya: added the dependancy and used assertions

msalautomationapp/src/main/java/com/microsoft/identity/client/msal/automationapp/sdk/MsalAuthResult.java
```diff
@@ -42,4 +46,12 @@ public MsalAuthResult(@NonNull final IAuthenticationResult authenticationResult)
     public MsalAuthResult(@NonNull final Exception exception) {
         super(exception);
     }
+
+    public static void verifyATForPop(@NonNull final String shr) throws ServiceException {
+        Map<String, ?> tokens = IDToken.parseJWT(shr);
```
shahzaibj, line 58:
```suggestion
        final Map<String, ?> tokens = IDToken.parseJWT(shr);
```

msalautomationapp/src/main/java/com/microsoft/identity/client/msal/automationapp/sdk/MsalAuthResult.java
```diff
@@ -42,4 +46,12 @@ public MsalAuthResult(@NonNull final IAuthenticationResult authenticationResult)
...
+        // Verify if the url, path and http method are as expected
+        Assert.assertEquals("signedhttprequest.azurewebsites.net", tokens.get("u"));
+        Assert.assertEquals("/api/validateSHR", tokens.get("p"));
+        Assert.assertEquals("GET", tokens.get("m"));
```
shahzaibj: Let's define all the following in a constant file somewhere...since you're also using these in MsalSdk class as well:

- signedhttprequest.azurewebsites.net"
- /api/validateSHR
- GET

msalautomationapp/src/main/java/com/microsoft/identity/client/msal/automationapp/sdk/MsalSdk.java
```diff
@@ -145,6 +159,15 @@ public MsalAuthResult acquireTokenSilent(@NonNull MsalAuthTestParams authTestPar
...
         }
 
+        if (authTestParams.getAuthScheme() == AuthScheme.POP) {
+            acquireTokenParametersBuilder.withAuthenticationScheme(
+                    PoPAuthenticationScheme.builder()
+                            .withHttpMethod(HttpMethod.valueOf("GET"))
+                            .withUrl(new URL("https://signedhttprequest.azurewebsites.net/api/validateSHR"))
+                            .build()
+            );
+        }
```
shahzaibj: I see similar code above for interactive. It can probably move to a helper method in this class

`getAuthenticationScheme()`

shahzaibj (approved):
Left minor comments

msalautomationapp/src/main/java/com/microsoft/identity/client/msal/automationapp/sdk/MsalAuthResult.java
somalaya: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1643 This PR has the change suggested

PR #1669, opened 2022-08-04
Repackage automation tests

rpdome:
...as an attempt to consolidate / simplify the release process.

- Test plans are consolidated from 3 plans (MSAL/ADAL/broker) to 1, so that it's easier to keep track of.

- The package will now map 1:1 with the structure of the test suites, such that the release engineer can easily fill in the test results without having to search for individual tests in each subfolder to mark.

- The following tests are removed.

1. Fairfax and blackforest tests are removed.
2. 940393 is now redundant with 938365, removed 940393.
3. 796049 is now redundant with 938365, removed 796049.
4. 796048 is now redundant with 938383, removed 796048.

https://identitydivision.visualstudio.com/Engineering/_testPlans/define?planId=2007357&suiteId=2007405


fadidurah (approved):
LGTM, I was wanting to suggest removing the blackforest and fairfax tests so I'm definitely ok with this change.

PR #1672, opened 2022-08-08
Move MSAL Tests, bring back some tests too

rpdome:
Added back the previously removed tests. I thought it was redundant, but it's not.  (796048, 796049)
They're the same tests, but one is with broker, and the others are without.

After adding back, I moved MSAL-Only USGov tests to testpass/msalonly/usgov

Added 2016158 as we'll need that test on both brokered and non-brokered scenario.


fadidurah (approved):
Looks good to me

PR #1674, opened 2022-08-08
Add multi cloud DCF test to cover ICM

rpdome:
Adding 2 cases 

First Case
1. Init PCA in WW
2. Init PCA in USGov
3. both PCA should return URL from its own cloud.

Second Case (gets from the latest IcM update)
1. Init PCA in USGov
2. perform DCF
3. Init another PCA in USGov
4. the 2nd PCA should return USGov URI (but it returns WW. This is fixed in
 https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1820

----
https://portal.microsofticm.com/imp/v3/incidents/details/325344544/home

### DIFF
--- a/msal/src/test/java/com/microsoft/identity/client/e2e/tests/mocked/DeviceCodeFlowApiTest.java
+++ b/msal/src/test/java/com/microsoft/identity/client/e2e/tests/mocked/DeviceCodeFlowApiTest.java
@@ -32,6 +32,7 @@ import androidx.annotation.NonNull;
 import androidx.test.core.app.ApplicationProvider;
 
 import com.microsoft.identity.client.IAuthenticationResult;
+import com.microsoft.identity.client.IMultipleAccountPublicClientApplication;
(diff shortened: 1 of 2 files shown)


shahzaibj (approved):

moumighosh (approved):

msal/src/test/java/com/microsoft/identity/client/e2e/tests/mocked/DeviceCodeFlowApiTest.java
```diff
@@ -558,4 +559,106 @@ public void onError(@NonNull MsalException exception) {
         Assert.assertEquals(ww, "https://microsoft.com/devicelogin");
         Assert.assertEquals(usgov, "https://microsoft.com/deviceloginus");
     }
+
+    // The same device code url shall be the same for 2 PCA objects with the same configuration (pointing to USGov)
```
somalaya, line 563: nit : can remove same from here

somalaya (approved):

PR #1683, opened 2022-08-18
Implement Test-Sharding in UI Test Automation Pipelines to Decrease Runtime

fadidurah:
**What:**
This PR introduces test sharding to allow our UI Automation Pipelines to split up their test cases into multiple shards and run each shard on a separate device in firebase, allowing the test cases to run in parallel.

**Why:**
This addition will greatly reduce the runtime of the Automation pipelines as well as eliminate the possibility of running into the 45 minute execution time limit on firebase physical devices. Also, this opens the door to merge the MSAL automation pipelines since there is no need to split the test cases between two pipelines to shorten the run time.

**How:**
Test Sharding is implemented using [Flank](https://flank.github.io/flank/) Test Runner. Flank is compatible with gcloud, and was integrated into our pipeline YAML files by leveraging our existing gcloud settings/arguments and adding a few more Flank specific arguments in the flank.yml file included in this PR. This PR also includes the creation of a directory `uiautomation/templates/flank` to hold the flank related YAML scripts. This leaves the old pipeline scripts that did not incorporate sharding in case we want to run all tests on a single device.

**Testing:**
Ran pipelines multiple times to see that the flank implementation is still running all test cases, with no drop in consistency or pass rate.


shahzaibj (approved):
LGTM! Thanks @fadidurah

p3dr0rv (approved):
:shipit:

PR #1684, opened 2022-08-18
Adding intune app SDK, gradle plugin to MSAL project and updated MSAL…

shoatman:
First step in adding Intune App Protection SDK to the MSAL Test app is adding the references for it and the supporting gradle plugin to the MSAL root project and the MSAL test app project.  

To read more about how Intune App SDK is integrated with an Android App please refer to: https://docs.microsoft.com/en-us/mem/intune/developer/app-sdk-android

### DIFF
--- a/testapps/testapp/src/main/AndroidManifest.xml
+++ b/testapps/testapp/src/main/AndroidManifest.xml
@@ -38,7 +38,7 @@
 
     <application
         android:networkSecurityConfig="@xml/network_security_config"
-        android:name="com.microsoft.identity.client.testapp.MsalSampleApp"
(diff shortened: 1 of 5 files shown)


melissaahn (approved):

rpdome (approved):

testapps/testapp/src/main/AndroidManifest.xml
```diff
@@ -38,7 +38,7 @@
 
     <application
         android:networkSecurityConfig="@xml/network_security_config"
-        android:name="com.microsoft.identity.client.testapp.MsalSampleApp"
```
iamgusain, line 41: Do we need this change to the name?

PR #1693, opened 2022-09-12
Work to get 2 MSAL Automation pipelines to green

fadidurah:
Pipelines:
https://dev.azure.com/IdentityDivision/Engineering/_build?definitionId=1743&_a=summary
https://dev.azure.com/IdentityDivision/Engineering/_build?definitionId=1742&_a=summary

Takes in changes from Common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1840

This PR takes in the common changes to fix consent page automation, as well as adding RetryOnFailure annotation to flaky broker tests.  Also, made an addition to the pipeline yaml files that excludes tests with @Ignore annotation from being reported in azure (These tests were bringing down pass percentage even though they were being ignored).

There were some transient errors that were causing issues in the tests, those were resolved and the pipelines above have been all-green since the last change made in this PR. Will need to keep monitoring the pipelines are seeing if they are still green.

### DIFF
--- a/azure-pipelines/ui-automation/templates/flank/flank.yml
+++ b/azure-pipelines/ui-automation/templates/flank/flank.yml
@@ -2,7 +2,7 @@ flank:
    ## Max Test Shards
    # test shards - the amount of groups to split the test suite into
    # set to -1 to use one shard per test. default: 1
-   max-test-shards: 5
+   max-test-shards: 1
 
    ## Number of Test Runs
    # test runs - the amount of times to run the tests.
--- a/azure-pipelines/ui-automation/templates/flank/flank.yml
+++ b/azure-pipelines/ui-automation/templates/flank/flank.yml
@@ -24,4 +24,4 @@ flank:
    # 'multi' is used as the default. The output style 'multi' is not displayed correctly on consoles
    # which don't support ansi codes, to avoid corrupted output use single or verbose.
    # The output style `compact` is used to produce less detailed output, it prints just Args, test and matrix count, weblinks, cost, and result reports.
-    output-style: single
+   output-style: single
--- a/azure-pipelines/ui-automation/templates/flank/flank.yml
+++ b/azure-pipelines/ui-automation/templates/flank/flank.yml
@@ -1,19 +1,27 @@
 flank:
-  ## Max Test Shards
-  # test shards - the amount of groups to split the test suite into
-  # set to -1 to use one shard per test. default: 1
+   ## Max Test Shards
+   # test shards - the amount of groups to split the test suite into
+   # set to -1 to use one shard per test. default: 1
    max-test-shards: 5
 
-  ## Number of Test Runs
-  # test runs - the amount of times to run the tests.
-  # 1 runs the tests once. 10 runs all the tests 10x
+   ## Number of Test Runs
+   # test runs - the amount of times to run the tests.
+   # 1 runs the tests once. 10 runs all the tests 10x
    num-test-runs: 1
 
-  ## Use Average Test Time for New Tests flag
-  # Enable using average time from previous tests duration when using SmartShard and tests did not run before.
-  # Default: false
+   ## Use Average Test Time for New Tests flag
+   # Enable using average time from previous tests duration when using SmartShard and tests did not run before.
+   # Default: false
    use-average-test-time-for-new-tests: true
 
-  ## Local Results Directory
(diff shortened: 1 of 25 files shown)


azure-pipelines/ui-automation/templates/flank/flank.yml
```diff
@@ -1,19 +1,27 @@
...
+   # Local folder to store the test result. Folder is DELETED before each run to ensure only artifacts from the new run are saved.
+   local-result-dir: flank
+
+   ## Output Style flag
+   # Output style of execution status. May be one of [verbose, multi, single, compact].
+   # For runs with only one test execution the default value is 'verbose', in other cases
+   # 'multi' is used as the default. The output style 'multi' is not displayed correctly on consoles
+   # which don't support ansi codes, to avoid corrupted output use single or verbose.
+   # The output style `compact` is used to produce less detailed output, it prints just Args, test and matrix count, weblinks, cost, and result reports.
+   output-style: single
```
wainaina, line 27: Add new line at EOF

wainaina (approved):

azure-pipelines/ui-automation/templates/flank/flank.yml
shahzaibj, line 27: Was this also the default?

azure-pipelines/ui-automation/templates/flank/run-on-firebase-with-flank.yml
```diff
@@ -88,6 +91,7 @@ jobs:
               --environment-variables "clearPackageData=true" \
               --results-history-name "${{ parameters.resultsHistoryName }}" \
               --test-targets "${{ parameters.testTargetPackages }}" \
+              --test-targets "${{ parameters.ignoredTarget }} " \
```
shahzaibj: How does this work actually? Reading the line it seems the ignored are going to be included? but I believe they are actually left out?

azure-pipelines/ui-automation/templates/flank/run-on-firebase-with-flank.yml
```diff
@@ -6,6 +6,9 @@ parameters:
   - name: testTargetPackages
     displayName: Packages as Test Targets
     type: string
+  - name: ignoredTarget
```
shahzaibj: Okay, looking at usage, it seems this can be used in any form by the caller and not really meant to denote "ignored" tests. I mean anyone can supply a value to this that says "package aaa.bbb.ccc" and that means those are now "included". So this is really down to the caller using this correctly and supplying something along the lines of either "notPackage or notAnnotation". Otherwise it just wouldn't do what we think it does. Is that right?

azure-pipelines/ui-automation/templates/flank/run-on-firebase-with-flank.yml
fadidurah: So it's a little weird, but the parameter passed in this case is "notAnnotation" (found in the yml files that are utilizing this template), which would exclude classes with a particular annotation, I don't think you can do 
`--test-targets " not${{ parameters.ignoredTarget }} "`

azure-pipelines/ui-automation/templates/flank/run-on-firebase-with-flank.yml
fadidurah: Yes, that's right. I'm thinking of rolling this back because of that, and just hardcoding `notAnnotation org.junit.Ignore` within the `run-on-firebase-with-flank.yml` file, or include it at the end of the default value for `testTargetPackages` parameter

azure-pipelines/ui-automation/templates/flank/flank.yml
fadidurah: The default was `verbose` in this case, which was producing a lot of output that I felt was unnecessary.

azure-pipelines/ui-automation/templates/flank/run-on-firebase-with-flank.yml
fadidurah: Removed this, added it as part of default value for testTarget parameter.

azure-pipelines/ui-automation/templates/flank/flank.yml
fadidurah: There should be one already

shahzaibj (approved):

PR #1703, opened 2022-10-03
Add new exception type for broker protocol not supported exception during msal-broker handshake

rpdome:
When MSAL initialization fails due to a protocol handshake error, we will be returning an exception with enough info for the client app to show an actionable prompt to the customer. (i.e. update app XXXX)


mohitc1 (approved):
:shipit:

fadidurah (approved):
LGTM!

PR #1721, opened 2022-11-08
Remove Schedules from MSAL UI Pipeline YMLs, add Multi-API Testing Support.

fadidurah:
Creating this PR to remove the hardcoded schedules from the MSAL pipelines YML files, which are being replaced by the usage of the scheduled triggers feature in ADO. ADO Schedule feature does already override whatever schedule is set in the yml file, but removing this schedule will allow us to turn off scheduled runs for a pipeline by disabling the schedule trigger in ADO (Since the schedule in the yml is removed here, this will not leave any scheduled runs and we can avoid the cost of running broken pipelines).

Doing this for the following scheduled Pipelines:
Dev MSAL With Dev BrokerHost: [Pipelines - Runs for Dev MSAL with Dev BrokerHost (azure.com)](https://dev.azure.com/IdentityDivision/Engineering/_build?definitionId=1742)
Dev MSAL With Prod Broker: [Pipelines - Runs for Dev MSAL with Prod Broker (AuthApp, CP) (azure.com)](https://dev.azure.com/IdentityDivision/Engineering/_build?definitionId=1761)
Dev MSAL Without Broker: [Pipelines - Runs for Dev MSAL without Broker (azure.com)](https://dev.azure.com/IdentityDivision/Engineering/_build?definitionId=1743)
MSAL Release with Prod Broker: [Pipelines - Runs for MSAL Release Test with Prod Broker (azure.com)](https://dev.azure.com/IdentityDivision/Engineering/_build?definitionId=1763)

Also included a fix to B2C url, now using custom branded domain since global domain is broken at the moment.

### DIFF
--- a/azure-pipelines/ui-automation/msal-broker-BrokerHost-nightly.yml
+++ b/azure-pipelines/ui-automation/msal-broker-BrokerHost-nightly.yml
@@ -23,6 +23,10 @@ parameters:
     displayName: Firebase Device Android Version (Api 29-)
     type: number
     default: 28
+  - name: flankShards
+    displayName: Max Number of Flank Shards
+    type: number
+    default: 2
   - name: testTargetPackages
     displayName: Packages as Test Targets
     type: string
@@ -107,7 +111,7 @@ stages:
     dependsOn:
       - msalautomationapp
       - brokers
-    displayName: Running MSAL with Broker Test UI Test Suite (API 30+)
+    displayName: Running MSAL with Broker Test UI Test Suite (API ${{ parameters.firebaseDeviceAndroidVersionHigh }})
     jobs:
       - template: ./templates/flank/run-on-firebase-with-flank.yml
         parameters:
@@ -121,14 +125,15 @@ stages:
           resultsDir: "msal-BrokerHost-highapi-$(Build.BuildId)-$(Build.BuildNumber)"
           firebaseDeviceId: ${{ parameters.firebaseDeviceIdHigh }}
           firebaseDeviceAndroidVersion: ${{ parameters.firebaseDeviceAndroidVersionHigh }}
-          testRunTitle: "Broker(MSAL) UI Automation - Build (API 30+) # $(Build.BuildNumber)"
+          testRunTitle: "Broker(MSAL) UI Automation - Build (API ${{ parameters.firebaseDeviceAndroidVersionHigh }}) # $(Build.BuildNumber)"
           apiLevelTarget: "notAnnotation com.microsoft.identity.client.ui.automation.annotations.RunOnAPI29Minus"
+          flankShards: ${{ parameters.flankShards }}
   # MSAL with Broker Test Plan stage (API 29-)
   - stage: 'msal_with_broker_low_api'
     dependsOn:
       - msalautomationapp
       - brokers
-    displayName: Running MSAL with Broker Test UI Test Suite (API 29-)
+    displayName: Running MSAL with Broker Test UI Test Suite (API ${{ parameters.firebaseDeviceAndroidVersionLow }})
     jobs:
       - template: ./templates/run-on-firebase.yml
         parameters:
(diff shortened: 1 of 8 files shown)


shahzaibj (approved):

p3dr0rv (approved):

azure-pipelines/ui-automation/templates/run-on-firebase.yml
```diff
@@ -73,14 +53,14 @@ jobs:
               --app "${{ parameters.automationAppApkPath }}" \
               --test "${{ parameters.automationAppTestApkPath }}" \
               --device "model=${{ parameters.firebaseDeviceId }},version=${{ parameters.firebaseDeviceAndroidVersion }}" \
-              --timeout "$(firebaseTimeout)" \
+              --timeout "${{ parameters.firebaseTimeout }}" \
```
p3dr0rv, line 56: I do not see firebaseTimeout defined here

azure-pipelines/ui-automation/templates/flank/run-on-firebase-with-flank.yml
```diff
@@ -79,15 +59,15 @@ jobs:
               --test "${{ parameters.automationAppTestApkPath }}" \
               --auto-google-login \
               --record-video \
-              --other-files ${{ parameters.otherFiles }} \
               --device model=${{ parameters.firebaseDeviceId }},version=${{ parameters.firebaseDeviceAndroidVersion }},locale=en,orientation=portrait \
-              --timeout "$(firebaseTimeout)" \
+              --timeout "${{ parameters.firebaseTimeout }}" \
```
p3dr0rv, line 65: firebaseTimeout is not defined here

azure-pipelines/ui-automation/templates/flank/flank.yml
```diff
@@ -2,7 +2,7 @@ flank:
    ## Max Test Shards
    # test shards - the amount of groups to split the test suite into
    # set to -1 to use one shard per test. default: 1
-   max-test-shards: 3
+   max-test-shards: 2
```
p3dr0rv: can we make this a parameter?

azure-pipelines/ui-automation/msal-broker-BrokerHost-nightly.yml
```diff
@@ -122,12 +102,12 @@ stages:
           - publish: $(Build.ArtifactStagingDirectory)/oldBrokerHost
             displayName: 'Publish Broker Host old apk for later use'
             artifact: oldBrokerHost
-  # MSAL with Broker Test Plan stage
-  - stage: 'msal_with_broker'
```
p3dr0rv, line 126: this is hardcoded to API 30 but  it is based on **firebaseDeviceIdHigh** 
should we change it here and below?

(same with 29 and the other file)

azure-pipelines/ui-automation/templates/run-on-firebase.yml
fadidurah: it is defined in line 16

azure-pipelines/ui-automation/templates/flank/run-on-firebase-with-flank.yml
fadidurah: Also defined in line 16, default is 45m

azure-pipelines/ui-automation/msal-broker-BrokerHost-nightly.yml
fadidurah: I'm a little confused what you mean. Where is API 30 is hardcoded? API 30+ is meant to imply any API level above 29. the idea is firebaseDeviceIdHigh and firebaseDeivceAndroidVersionHigh are meant to point to a firebase device with API level 30 or above, same thing with firebaseDeviceLowHigh but for API 29 and below

azure-pipelines/ui-automation/templates/flank/flank.yml
fadidurah: I'm inclined to say yes. I originally did not because we can't pass parameters to this flank.yml file (it's not used as a template), which is where flank currently gets it's flank-specific parameters. I think we can pass this `max-test-shards` parameter at run time to the gcloud command itself, which has access to the yml parameters. I can take another look and wait on merging this PR

p3dr0rv: ok, Yeah, it does not have to be in this PR, but we can maybe think a way to do this if it is convenient.

azure-pipelines/ui-automation/msal-broker-BrokerHost-nightly.yml
p3dr0rv: OK I understood now. 
yeah, I was wondering if we should display which specific version it is running instead of API 30 +, or maybe both 
example something like this:
displayName: Running MSAL with Broker Test UI Test Suite (API 30+) {{parameters.firebaseDeviceIdHigh}}

fadidurah: Ohhhh I see what you mean. That's a good suggestion! I think that it is good to include

PR #1747, opened 2023-01-11
Wrap Runnable with Current Telemetry Context + Integrate Open Telemetry exporter in MSAL Test App

shahzaibj:
**Wrap Runnable with Current Telemetry Context** - This is to ensure that if spans or traces are started outside of a runnable, then we can carry the same span or trace forward into the runnable. This is particularly required for our partners that have integrated with our broker open telemetry work and have also decided to create their own spans wrapping around MSAL usage. In this case, we must carry the trace information forward so we can create child spans with correct trace ids.

**Note:** This PR also updates source and target compatibility of MSAL to Java 8, however, this is still NOT considered a MAJOR change for this particular PR because our common lib has been on Java 8 for some time and therefore the nature of this change in MSAL own build.gradle doesn't have any new effect more than what was introduced already for common several months back.

**Integrate Open Telemetry exporter in MSAL Test App** - This is to facilitate testing of Span Context Propagation between MSAL and Broker. So that we can test this directly with our MSAL Test app.

Related: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1956


testapps/testapp/build.gradle
```diff
@@ -50,10 +55,11 @@ android {
     defaultConfig {
         multiDexEnabled true
         applicationId "com.msft.identity.client.sample"
-        minSdkVersion rootProject.ext.minSdkVersion
+        minSdkVersion 24
```
fadidurah, line 58: Should we create a variable for this? I think Sowmya pushed a similar change recently and had created a variable with this 24 version, I don't remember if it's the same use case though

fadidurah (approved):

testapps/testapp/build.gradle
somalaya: Yes, I created this variable : brokerProjectMinSdkVersion. May be we create similar one for MSAL

msal/build.gradle
```diff
@@ -239,6 +239,8 @@ dependencies {
     String common4jVersion = project.hasProperty("distCommon4jVersion") ? project.distCommon4jVersion : "0.0.+"
     testDistImplementation(testFixtures("com.microsoft.identity:common4j:${common4jVersion}"))
 
+    implementation platform("io.opentelemetry:opentelemetry-bom:1.18.0")
```
iamgusain, line 240: nit: The version numbers can go in versions.gradle file, to make it easier to update in one place.

shahzaibj: I will handle all these version numbers in separate PR.

testapps/testapp/build.gradle
shahzaibj: Let me handle version numbers in a separate PR.

msal/build.gradle
```diff
@@ -45,7 +45,7 @@ android {
 //        }
 //    }
 
-    /*
+
     //Commenting out until the next major version of common/msal/etc...
```
iamgusain: remove the comment too?

shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1747/commits/d8cd7660f27322126e297f10612573a567ef5292

iamgusain (approved):

PR #1753, opened 2023-01-23
fixes to run unit test for dist flavor

iamgusain:
### What
Fix to run ut in dist flavor

### Why
Currently some of these only have :project dependencies, which don't resolve when we run tests in dist flavor from standalone repo.

### How
Adding dist flavor for dependencies for running tests.

### Testing
Ran UTs with dist flavor. Sample run
https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1042260&view=results

### DIFF
--- a/msal/build.gradle
+++ b/msal/build.gradle
@@ -184,8 +184,6 @@ task sourcesJar(type: Jar) {
 String commonVersion = project.hasProperty("distCommonVersion") ? project.distCommonVersion : "0.0.+"
 
 dependencies {
-    testImplementation project(path: ':common4j')
-
     //Please leave this in... desugaring is currently disabled by default; however it's required for running some tests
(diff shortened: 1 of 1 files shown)


somalaya (approved):

msal/build.gradle
```diff
@@ -184,8 +184,6 @@ task sourcesJar(type: Jar) {
 String commonVersion = project.hasProperty("distCommonVersion") ? project.distCommonVersion : "0.0.+"
 
 dependencies {
-    testImplementation project(path: ':common4j')
```
somalaya, line 187: why was there a testImplementation project(path: ':common4j') and testImplementation(testFixtures(project(":common4j"))) before? Now I see we have removed this.

shahzaibj (approved):

msal/build.gradle
iamgusain: We don't need testImplementation project(path: ':common4j'), as we are only using testFixtures from common4j.

PR #1791, opened 2023-04-04
LTW support work

fadidurah:
LTW support work

Supported so far:
LTW class for handling installation, updating, etc.
Custom broker installation (broker shuffling)


msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/AbstractMsalCustomBrokerInstallationTest.java
```diff
@@ -0,0 +1,95 @@
...
+    // Fields to access installed brokers
+    public BrokerHost mBrokerHost;
+    public BrokerMicrosoftAuthenticator mAuthenticator;
+    public BrokerCompanyPortal mCompanyPortal;
+    public BrokerLTW mLtw;
+
+    @Override
+    public void installOldBrokerHost() {
+        mBrokerHost = new BrokerHost(BrokerHost.OLD_BROKER_HOST_APK,
+                BrokerHost.BROKER_HOST_APK);
```
somalaya: do we still need to pass the new broker host apk name here? Or can we move this parameter to be passed in installBrokerHost method?

fadidurah: So, I took a look.

image

This is if we pass a singular apk name. Looks like it's used for both the initial install and the update file name, so I would think that the update call would be broken here. I'm going to leave it as is for now.

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/AbstractMsalCustomBrokerInstallationTest.java
```diff
@@ -0,0 +1,95 @@
...
+import com.microsoft.identity.client.ui.automation.broker.BrokerCompanyPortal;
+import com.microsoft.identity.client.ui.automation.broker.BrokerHost;
+import com.microsoft.identity.client.ui.automation.broker.BrokerLTW;
+import com.microsoft.identity.client.ui.automation.broker.BrokerMicrosoftAuthenticator;
+
+/**
+ * Msal UI Testing class that allows devs to easily install old/legacy and updated versions of brokers
+ * quickly to test complex update scenarios.
+ */
+public abstract class AbstractMsalCustomBrokerInstallationTest extends AbstractMsalUiTest implements ICustomBrokerInstallationTest {
```
shahzaibj, line 36: Should this whole class by moved to `uiautomationutilities`? It almost seems like it can easily move there

fadidurah: This would make sense. the reason I created two classes and placed one in each automation app is to extends AbstractMsalUiTest and AbstractAdalLocalTest, which allows it to leverage some more of the existing msal and adal base ui automation code. Both approaches will cause some duplicate code, no? I'd need to extend this class to make the syntax match existing MSAL and ADAL tests

fadidurah: Reworked this to use default methods in the interface

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/AbstractMsalCustomBrokerInstallationTest.java
```diff
@@ -0,0 +1,37 @@
...
```
shahzaibj, line 36: Could also consider removing this class and using the `extends AbstractMsalUiTest implements ICustomBrokerInstallationTest` directly in child classes, but I'm fine either way

shahzaibj (approved):

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/AbstractMsalCustomBrokerInstallationTest.java
fadidurah: I could go either way as well but I think this is simpler since devs just need to extend one class

rpdome (approved):

PR #1793, opened 2023-04-05
Changes to support MSA accounts in Broker

shahzaibj:
Two things done here:

- Let MSAL go to the Broker for MSA accounts (Related: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2003)
- Some improvements in MSAL Test App. I noticed that GetUsers fragment was always using default msal config resource id for loading accounts. So updated that to honor current request options. Also the correct place to load PCA is inside `onResume` so updated that too.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.java
@@ -22,34 +22,29 @@
 //  THE SOFTWARE.
 package com.microsoft.identity.client.internal.controllers;
 
+import static com.microsoft.identity.common.java.AuthenticationConstants.Broker.BROKER_ACCOUNT_TYPE;
+
 import android.accounts.AccountManager;
 import android.accounts.AuthenticatorDescription;
 import android.annotation.TargetApi;
-import android.app.Activity;
 import android.content.Context;
 import android.os.Build;
 import android.os.PowerManager;
 
 import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
 
 import com.microsoft.identity.client.PublicClientApplicationConfiguration;
 import com.microsoft.identity.client.exception.MsalClientException;
-import com.microsoft.identity.common.java.authorities.AnyPersonalAccount;
-import com.microsoft.identity.common.java.authorities.Authority;
-import com.microsoft.identity.common.java.authorities.AzureActiveDirectoryAuthority;
 import com.microsoft.identity.common.internal.broker.BrokerValidator;
-import com.microsoft.identity.common.java.controllers.BaseController;
 import com.microsoft.identity.common.internal.controllers.BrokerMsalController;
 import com.microsoft.identity.common.internal.controllers.LocalMSALController;
-import com.microsoft.identity.common.java.flighting.CommonFlight;
-import com.microsoft.identity.common.java.flighting.CommonFlightManager;
+import com.microsoft.identity.common.java.authorities.Authority;
+import com.microsoft.identity.common.java.authorities.AzureActiveDirectoryAuthority;
+import com.microsoft.identity.common.java.controllers.BaseController;
 import com.microsoft.identity.common.logging.Logger;
 
 import java.util.ArrayList;
 import java.util.List;
-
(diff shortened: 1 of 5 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.java
```diff
@@ -134,13 +136,15 @@ public static boolean brokerEligible(@NonNull final Context applicationContext,
             return false;
         }
 
-        //Do not use broker when the audience is MSA only (personal accounts / consumers tenant alias)
-        AzureActiveDirectoryAuthority azureActiveDirectoryAuthority = (AzureActiveDirectoryAuthority) authority;
+        if (!CommonFlightManager.isFlightEnabled(CommonFlight.SUPPORT_MSA_ACCOUNTS_IN_BROKER)) {
```
iamgusain: How is this flight going to be delivered in msal?
Is the idea that individual app team will provide the flight value?

shahzaibj: The idea here is to just use the default value. I just used the flight code since it was there as common flight manager so I thought perhaps the idea was that at some point we would let app teams control it. 

I can actually get rid of this whole code altogether since it doesn't do much.

iamgusain: Yep, app teams (or any consumer of common) can control the flights in common.

For this case I think we should let Broker control it, so we can either remove it from here, or expose canSupportMsaAccountsInBroker from Broker side.

shahzaibj: We already have the flight in Broker as well. This one here was just for double protection since this helps evaluate the flight early (as opposed to during broker handshake and hence I put it here). If this causes confusion then I can just remove this flight block and just let the flow proceed which would then be handled later during hello handshake.

iamgusain: I think better to remove it from here.

iamgusain (approved):
Approved with suggestion.

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
```diff
@@ -43,6 +43,7 @@
 import android.widget.Toast;
 import android.widget.ToggleButton;
 
+import androidx.annotation.NonNull;
```
melissaahn:
```suggestion
```

melissaahn (approved):

msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.java
shahzaibj: Addressed. Removed.

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
shahzaibj: Addressed here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1793/commits/6362f0b11c7a210897be32dffc4f690506c2f349

msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.java
```diff
@@ -42,6 +42,8 @@
 import com.microsoft.identity.common.java.controllers.BaseController;
 import com.microsoft.identity.common.internal.controllers.BrokerMsalController;
 import com.microsoft.identity.common.internal.controllers.LocalMSALController;
+import com.microsoft.identity.common.java.flighting.CommonFlight;
+import com.microsoft.identity.common.java.flighting.CommonFlightManager;
```
mohitc1: >CommonFlightManager [](http://example.com/codeflow?start=52&length=19)

nit:unused now

mohitc1 (approved):
:shipit:

PR #1826, opened 2023-05-18
Replacing SHA-1 used in broker validation with SHA-512

melissaahn:
See https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2019 for description.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,6 +1,7 @@
 MSAL Wiki : https://github.com/AzureAD/microsoft-authentication-library-for-android/wiki
 V.NEXT
 ----------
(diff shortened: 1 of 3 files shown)


rpdome (approved):

melissaahn:
Per Fadi, one of the CIAM test tenants is undergoing some changes, which is making a few tests fail. Temporarily adding in @Ignore for AcquireTokenCIAMTest so that PRs won't be blocked until the tenant can be fixed.

mohitc1 (approved):

PR #1827, opened 2023-05-18
Amritsaini/add teams whitlisting check for dcf api

Amritkrsaini1:
enabled the acquire token DCF with device id claims APIs for teams' application only.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -117,7 +117,11 @@ import com.microsoft.identity.common.java.exception.BaseException;
 import com.microsoft.identity.common.java.exception.ClientException;
 import com.microsoft.identity.common.java.exception.ErrorStrings;
 import com.microsoft.identity.common.java.exception.ServiceException;
+import com.microsoft.identity.common.java.opentelemetry.AttributeName;
+import com.microsoft.identity.common.java.opentelemetry.OTelUtility;
 import com.microsoft.identity.common.java.opentelemetry.OtelContextExtension;
+import com.microsoft.identity.common.java.opentelemetry.SpanExtension;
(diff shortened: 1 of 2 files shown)


somalaya:
Please provide more context for people to understand why this change was required

#1. Issue/Problem description
#2. And the fix

somalaya (approved):
LGTM. Please test with the min and max broker protocol versions to ensure that it does not break with different versions.

Thanks Amrit!

mohitc1 (approved):

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1839,48 +1841,56 @@ public void onError(MsalException exception) {
...
-//    }
+    public void acquireTokenWithDeviceCode(@NonNull List<String> scopes, @NonNull final DeviceCodeFlowCallback callback, @Nullable final ClaimsRequest claimsRequest, @Nullable final UUID correlationId) {
+        final Context context = mPublicClientConfiguration.getAppContext();
+        PackageHelper packageHelper = new PackageHelper(context);
+
+        // Currently this method is only supported for Teams app
+        if (!packageHelper.verifyIfValidTeamsPackage(context.getPackageName())) {
+            throw new UnsupportedOperationException(ERR_UNSUPPORTED_OPERATION);
+        } else {
+            DeviceCodeFlowParameters.Builder builder = new DeviceCodeFlowParameters.Builder();
```
somalaya: could we please add some telemetry here so that we know if anyone is requesting for this API?

Amritkrsaini1: Added

Amritkrsaini1:
Enabling the DCF with claims id API support only for teams app and for any other client we will throw unsupported exception

Amritkrsaini1:
Tested min max broker version scenarios , working as expected

PR #1835, opened 2023-06-01
Wire new Broker Discovery Client into MSAL - still disabled by default

rpdome:
MSAL counterpart of https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2057/

## Changes
1. Update MSALControllerFactory
    - Turn static method into non-static ones, so that we can reuse `IPlatformComponents` throughout the flow.
    - This will allow us to invoke broker discovery once per call as well (and use it throughout the flow`
3. Update the usage of MSALControllerFactory.


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -923,20 +923,10 @@ private static void create(@NonNull final PublicClientApplicationConfiguration c
...
-                    config);
-        } catch (MsalClientException e) {
-            listener.onError(e);
-            return;
-        }
-
         final GetDeviceModeCommand command = new GetDeviceModeCommand(
                 params,
-                controller,
+                new MSALControllerFactory(config).
```
melissaahn: nit: controllerfactory instantiation calls similar this one in this class are stylistically inconsistent; I personally like 1368-1369 the best (1213-1214 is similar, but feels like the parentheses are off?)

msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
```diff
@@ -122,36 +122,17 @@ private void getAccountsInternal(@NonNull final LoadAccountsCallback callback,
...
             @Override
             public void onMigrationFinished(int numberOfAccountsMigrated) {
-                final Handler handler;
-
-                if (null != Looper.myLooper() && Looper.getMainLooper() != Looper.myLooper()) {
-                    handler = new Handler(Looper.myLooper());
-                } else {
-                    handler = new Handler(Looper.getMainLooper());
-                }
-
```
shahzaibj, line 132: We no longer need the handler code?

msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
```diff
@@ -379,38 +353,31 @@ private void removeAccountInternal(@Nullable final IAccount account,
...
 
-            CommandDispatcher.submitSilent(removeAccountCommand);
+                    @Override
+                    public void onTaskCompleted(Boolean success) {
+                        callback.onRemoved();
+                    }
 
-        } catch (final MsalClientException e) {
-            callback.onError(e);
-        }
```
shahzaibj, line 413: We no longer need to catch this?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1881,32 +1862,15 @@ public void acquireTokenWithDeviceCode(@NonNull List<String> scopes, @NonNull fi
...
-                } catch (final MsalClientException e) {
-                    final MsalClientException clientException = new MsalClientException(
-                            UNKNOWN_ERROR,
-                            "Unexpected error while acquiring token with device code.",
-                            e
-                    );
-
-                    span.recordException(clientException);
-                    span.setStatus(StatusCode.ERROR);
-                    callback.onError(clientException);
```
shahzaibj, line 1908: No longer need to catch this?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -2335,14 +2299,14 @@ public boolean isValidBrokerPackage(@NonNull final Context context,
...
-        } catch (final ClientException e){
-            if (NOT_VALID_BROKER_FOUND.equalsIgnoreCase(e.getErrorCode())) {
-                return null;
-            }
+        final BrokerData activeBroker = BrokerDiscoveryClientFactory.getInstance(context,
+                        AndroidPlatformComponentsFactory.createFromContext(context))
+                .getActiveBroker(false);
 
-            throw MsalExceptionAdapter.msalExceptionFromBaseException(e);
+        if (activeBroker != null) {
```
shahzaibj: The method is marked Nullable, so you could just get rid of the `if` statement and just return `activeBroker.getPackageName()`. (then can also get rid of line 2310)

msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.kt
```diff
@@ -0,0 +1,179 @@
...
+        val activeBroker = getActiveBrokerPackageName()
+        val controllers: MutableList<BaseController> = ArrayList()
+        controllers.add(LocalMSALController())
+        if (!activeBroker.isNullOrEmpty() && brokerEligible(authority)) {
+            controllers.add(
+                BrokerMsalController(applicationContext, platformComponents, activeBroker)
+            )
+        }
+
+        return controllers
```
shahzaibj: Should we make the list `immutable` before returning it?

msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.kt
```diff
@@ -0,0 +1,179 @@
...
+    /**
+     * Returns true if the request is eligible to use broker (see [brokerEligible])
+     * AND if a valid broker is found.
+     **/
+    fun brokerEligibleAndInstalled(authority: Authority): Boolean {
+        return getActiveBrokerPackageName() != null && brokerEligible(authority)
+    }
+
+    /**
+     * Determine if request is eligible to use the broker
```
shahzaibj:
```suggestion
     * Determine if request is eligible to use the broker.
```

msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.kt
```diff
@@ -0,0 +1,179 @@
...
+        val methodTag = "$TAG:getActiveBrokerPackageName"
+        val activeBroker = discoveryClient.getActiveBroker(shouldSkipCache = false)
+        activeBroker?.let {
+            return it.packageName
+        }
+
+        Logger.info(methodTag,"Broker application is not installed.")
+        return null
+    }
+}
```
shahzaibj: nit: new line ending

msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
rpdome: yes - since we no longer throws exception from MSALControllerFactory. 

If MSALFactory hits any AccountManager error it will simply proceed as "Broker not found". (see LegacyBrokerDiscoveryClient.getActiveBroker())

The exception handler of the actual operation is passed via callback. no change there.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
rpdome: same answer as above.

msal/src/main/java/com/microsoft/identity/client/MultipleAccountPublicClientApplication.java
rpdome: same answer as below. No need to catch exception here, so no handler needed.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
rpdome: activeBroker itself could be null.

melissaahn (approved):

shahzaibj (approved):

PR #1836, opened 2023-06-01
Change applicationId of msalAutomationApp

somalaya:
Issue : We are unable to install MsalTestApp from UI Automation infrastructure. The msalAutomationApp crashes instantly when install command is executed without any stacktrace. This was the case only with MsalTestApp and not for any other test apps.

Cause : The applicationId of msalAutomationApp mentioned in build.gradle is same as the pkg name of MsalTestApp. So when msal test app install command runs, it crashes the msalAutomationApp as it is trying to install another app with same appId.

Fix : Changed the appId of msalAutomationApp to match the one mentioned in AndroidManifest.xml

Thanks @fadidurah for helping with the root cause.

### DIFF
--- a/msalautomationapp/build.gradle
+++ b/msalautomationapp/build.gradle
@@ -46,7 +46,7 @@ android {
 
     defaultConfig {
         multiDexEnabled true
-        applicationId "com.msft.identity.client.sample"
+        applicationId "com.microsoft.identity.client.msal.automationapp"
         minSdkVersion rootProject.ext.automationAppMinSDKVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode 1


iamgusain (approved):

shahzaibj:
Can you please run some UI Automation test cases locally (the ones in MSAL Automation App) to ensure they are continuing to work with this change?

somalaya:
> Can you please run some UI Automation test cases locally (the ones in MSAL Automation App) to ensure they are continuing to work with this change?

yes I will merge this PR only after the E2E tests in the pipeline runs successfully on my branch. Also ran few tests locally and those are succeeding

fadidurah:
> > Can you please run some UI Automation test cases locally (the ones in MSAL Automation App) to ensure they are continuing to work with this change?
> 
> yes I will merge this PR only after the E2E tests in the pipeline runs successfully on my branch. Also ran few tests locally and those are succeeding

I've updated the target branches to point to these branches with the new application ids, let's see how the run goes tomorrow.

fadidurah:
By the way, @somalaya did you also create a common branch to update what apps are uninstalled as part of initial rule pass? or is that waiting for the msaltestapp and adaltestapp implementations?

PR #1838, opened 2023-06-07
Add broker discovery support to test app

rpdome:
A picture is worth a thousand words

(This toggle switch is turned off by default, and not persisted in storage)

Screenshot 2023-06-07 at 5 12 51 PM


msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.kt
```diff
@@ -37,6 +38,7 @@ import com.microsoft.identity.common.java.controllers.BaseController
 import com.microsoft.identity.common.java.interfaces.IPlatformComponents
 import com.microsoft.identity.common.logging.Logger
 import com.microsoft.identity.msal.BuildConfig
+import kotlin.jvm.Throws
```
somalaya: do we need these imports?

somalaya (approved):

fadidurah (approved):

melissaahn (approved):

PR #1847, opened 2023-06-29
Fix exception in SDM 

somalaya:
Problem : We got an IcM (https://portal.microsofticm.com/imp/v3/incidents/details/390194576/home) in which a single user of yammer app reported that the app was crashing upon launch in SDM. 
Check the IcM sumamry for the stacktrace. The issue is that we are getting a null localAccountId from the cache. We could not figure out why this issue is happening on 1 particular device as we did not get more logs requested.

Fix :  To avoid NPE, we simply added a null check

Note : This does not fix the original problem of localAccountId being NULL but at least can help fix the app from crashing.


fadidurah (approved):

mohitc1:
>                 if (!acctHomeAccountId.contains(acctLocalAccountId)) {

do you want to patch here too?

---
Refers to: msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java:63 in 6a40709. [](commit_id = 6a40709351db6c1c66870875997aca84d893b0af, deletion_comment = False)

mohitc1 (approved):

PR #1848, opened 2023-06-29
Removing a flag entry in the cache for one cross cloud unit test

melissaahn:
## Summary
After [my PR was merged into common](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2058), one particular unit test in CrossCloudGuestAccountTest was failing. 
An assertion checks the size of the cache at one point, and the size expected is 4, while the actual value is 5. 5 is actually expected, because my PR added in another entry for a sha1-cleared flag. 
The next assertion also fails because the key for the flag doesn't follow the credential id convention. But since the flag isn't associated with a particular account, it shouldn't have an account id attached to it. 
This PR just removes the flag entry from the map before the assertions are run. As far as I know, there shouldn't be cases in prod where assumptions are made based just off of the size of the entire cache (the methods in SharedPreferencesAccountCredentialCache and SPACCWMC should only return parts of the cache that are specifically filtered to accounts, etc.). (but lmk if I've missed something)

### DIFF
--- a/msal/src/test/java/com/microsoft/identity/client/e2e/tests/mocked/CrossCloudGuestAccountTest.java
+++ b/msal/src/test/java/com/microsoft/identity/client/e2e/tests/mocked/CrossCloudGuestAccountTest.java
@@ -297,22 +297,17 @@ public class CrossCloudGuestAccountTest extends AcquireTokenAbstractTest {
         final IMultiTypeNameValueStorage sharedPreferences = mComponents.getStorageSupplier()
                 .getEncryptedFileStore(SHARED_PREFERENCES_NAME);
         final Map<String, ?> cacheValues = sharedPreferences.getAll();
+        //Getting rid of the SHA-1 cleared flag so it doesn't mess with the assertions below.
+        cacheValues.remove(SHA1_APPLICATION_IDENTIFIER_ACCESS_TOKEN_CLEARED);
 
-        if (cacheValues.size() == 5) {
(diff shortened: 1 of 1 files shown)


fadidurah (approved):

msal/src/test/java/com/microsoft/identity/client/e2e/tests/mocked/CrossCloudGuestAccountTest.java
```diff
@@ -296,6 +297,8 @@ public void onError(@NonNull MsalException exception) {
         final IMultiTypeNameValueStorage sharedPreferences = mComponents.getStorageSupplier()
                 .getEncryptedFileStore(SHARED_PREFERENCES_NAME);
         final Map<String, ?> cacheValues = sharedPreferences.getAll();
```
mohitc1, line 299: >final Map<String, ?> cacheValues = sharedPreferences.getAll(); [](http://example.com/codeflow?start=8&length=62)

nit: could use getAllFilteredByKey() if interested only on certain type or format of values. This way you're filtering in the entries you're interested in, and not filtering out later.

mohitc1 (approved):

msal/src/test/java/com/microsoft/identity/client/e2e/tests/mocked/CrossCloudGuestAccountTest.java
melissaahn: Good point. getAllFilteredByKey() does return an iterator though, so I'll have to rework the last parts of the test. In order to unblock other PRs, I'm going to keep the existing implementation and come back to this.

PR #1857, opened 2023-07-20
Added LTW annotation

somalaya:
All LTW tests should have @LTWTests annotation.  It was missing on TestCaseUpdateCompanyPortal, so added it.


fadidurah (approved):
Thanks for doing this!

melissaahn (approved):

PR #1863, opened 2023-07-25
change show pkg name toast to dialog

amoyixi:
change showMessage method from show Toast to show Dialog in order to process e2e test with MsalTestApp

### DIFF
--- a/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
+++ b/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
@@ -617,9 +617,14 @@ public class AcquireTokenFragment extends Fragment {
                 builder.setView(dialogView);
 
                 TextView dialogMessage = dialogView.findViewById(R.id.dialog_message);
+                Button okButton = dialogView.findViewById(R.id.dialog_ok_button);
                 dialogMessage.setText(msg);
 
                 AlertDialog dialog = builder.create();
+                okButton.setOnClickListener(v -> {
+                    dialog.dismiss();
+                });
+
                 dialog.show();
             }
         });
--- a/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
+++ b/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
@@ -25,6 +25,7 @@ package com.microsoft.identity.client.testapp;
 import static com.microsoft.identity.client.testapp.R.id.enablePII;
 
 import android.app.Activity;
+import android.app.AlertDialog;
 import android.content.ClipData;
 import android.content.ClipboardManager;
 import android.content.Context;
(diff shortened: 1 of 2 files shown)


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
```diff
@@ -603,11 +604,17 @@ public void showMessage(String message) {
 
     private void showMessage(final String msg) {
         new Handler(getActivity().getMainLooper()).post(new Runnable() {
-
             @Override
             public void run() {
-                Toast.makeText(getContext(), msg, Toast.LENGTH_LONG).show();
-                mStatus.setText(msg);
+                AlertDialog.Builder builder = new AlertDialog.Builder(getContext());
+                View dialogView = LayoutInflater.from(getContext()).inflate(R.layout.custom_dialog_layout, null);
```
somalaya, line 616: why did we create a custom layout? Can we not use default alert dialog? like in below image?

image

somalaya: Also, instead of changing showMessage method, can we add a new showDialog method in which you make this change.. and call showDialog method only in methods which are useful for your testing.

amoyixi: It's because as same as Toast, we can't obtain the Alert Dialog object by resource id. If we use alert dialog, we still need to obtain it by matching text, which means we still need to pass params in advance. That's why I create a custom dialog layout which has a TextView with resource id so that we can obtain the TextView Uiobject and then get the text in the TextView to compare with whatever we want in test cases.

common
somalaya, line 1: Why is this common submodule changes being pulled? Did you make any changes in common repo which is required in this PR?

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
somalaya: ok, could you please post a snapshot of how the dialog looks after your changes?

amoyixi: sure, it looks like this:
Screenshot 2023-08-02 at 15 20 21
Screenshot 2023-08-02 at 15 20 07

fadidurah (approved):

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
somalaya: How can we dismiss the dialog? I could not see any ok button in the snpshot

amoyixi: we can call `msalTestApp.handleBackButton();` to go back to the main page

somalaya: no, manually how can you dismiss it?

amoyixi: We can click anywhere outside the dialog to close the dialog

somalaya: It would be good to have an ok button. Is it a simple thing to add?

amoyixi: Added the OK button to the dialog. When the button is clicked the dialog would disappear. 
Screenshot 2023-08-09 at 14 14 21

somalaya (approved):

PR #1873, opened 2023-08-02
make changes to test app to validate Broker DCF flow

rpdome:
related PR: https://github.com/AzureAD/ad-accounts-for-android/pull/2457

Currently - the logic is wired to the local version of acquireTokenWithDeviceCode(), so i wire it with the new one and pass claimsrequest to it.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -137,6 +137,7 @@ import com.microsoft.identity.msal.BuildConfig;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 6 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -137,6 +137,7 @@
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
```
fadidurah: Added by accident?

fadidurah (approved):

mohitc1 (approved):
:shipit:

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
rpdome: ah yes.

PR #1882, opened 2023-08-15
Target, compile SDK, AGP and gradle versions increased

somalaya:
- Target sdk version changed to 33
- Compile sdk version changed to 34
- AGP version changed to 7.4.2
- Gradle distribution url changed to 7.5

After changing above versions, I had to update other dependencies for compatibility

- Powerlift version changed to 1.0.3
- Bouncy castle version changed to 1.69
- robolectric version to 4.9
- Also, with new gradle versions, some of the properties used in gradle were deprecated (naming conventions changed). So, made those changes
- With gradle version changes, we should not be having a package name in AndroidManifest.xml.. instead, should add a namespace property in build.gradle. https://developer.android.com/build/publish-library/prep-lib-release#choose-namespace
- For apps targeting Android 14, Android restricts apps from sending implicit intents to internal app components. So , added exported attribute. https://developer.android.com/about/versions/14/behavior-changes-14#safer-intents

### DIFF
--- a/azure-pipelines/pull-request-validation/pr-msal.yml
+++ b/azure-pipelines/pull-request-validation/pr-msal.yml
@@ -7,13 +7,13 @@ name: $(date:yyyyMMdd)$(rev:.r)
 
 variables:
 - name: BuildParameters.jdkVersion
-  value: 1.8
+  value: 1.11
 - name: BuildParameters.jdkArchitecture
   value: x64
 - name: BuildParameters.javaHomeSelection
   value: JDKVersion
 - name: robolectricSdkVersion
-  value: 28
+  value: 33
 
 trigger:
   branches:
@@ -44,8 +44,9 @@ jobs:
     persistCredentials: True
   - template: azure-pipelines/templates/steps/automation-cert.yml@common
   - task: JavaToolInstaller@0
-    displayName: Use Java 8
+    displayName: Use Java 11
     inputs:
+      versionSpec: '11'
       jdkArchitectureOption: x64
       jdkSourceOption: PreInstalled
   - task: Gradle@2
@@ -88,11 +89,12 @@ jobs:
     inputs:
       filename: echo
       arguments: '##vso[task.setvariable variable=ENV_VSTS_MVN_ANDROID_MSAL_ACCESSTOKEN]$(System.AccessToken)'
-  - task: Gradle@1
+  - task: Gradle@3
     displayName: Lint Local debug
     inputs:
       tasks: clean msal:lintLocalDebug
       publishJUnitResults: false
+      jdkVersion: 1.11
 - job: codeql
   displayName: CodeQL
   cancelTimeoutInMinutes: 1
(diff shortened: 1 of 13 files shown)


melissaahn (approved):

azure-pipelines/pull-request-validation/pr-msal.yml
```diff
@@ -7,13 +7,13 @@ name: $(date:yyyyMMdd)$(rev:.r)
...
 - name: BuildParameters.jdkVersion
-  value: 1.8
+  value: 1.11
 - name: BuildParameters.jdkArchitecture
   value: x64
 - name: BuildParameters.javaHomeSelection
   value: JDKVersion
 - name: robolectricSdkVersion
-  value: 28
+  value: 33
```
iamgusain, line 16: We have pipeline variable to set robolectric version (see- ProbolectricSdkVersion), please check if that needs to be updated as well.

testapps/testapp/src/main/AndroidManifest.xml
```diff
@@ -51,7 +51,8 @@
             android:name="com.microsoft.identity.client.testapp.StartActivity"
             android:taskAffinity=""
             android:theme="@style/AppTheme.NoActionBar"
-            android:windowSoftInputMode="stateHidden">
+            android:windowSoftInputMode="stateHidden"
+            android:exported="true">
```
iamgusain, line 55: does the calling app need to make any changes for these exported activities?

msal/build.gradle
```diff
@@ -161,7 +161,8 @@ android {
 task javadoc(type: Javadoc) {
     failOnError false
     source = android.sourceSets.main.java.srcDirs
-    classpath += configurations.compile
+    configurations.api.setCanBeResolved(true)
```
iamgusain, line 164: What does this do?

azure-pipelines/pull-request-validation/pr-msal.yml
iamgusain: oops.. this the pipeline variable itself that you are modifying :)

iamgusain (approved):

testapps/testapp/src/main/AndroidManifest.xml
somalaya: There is no action item on calling apps when they update MSAL with my changes. But, when they update the compileSDK version to 34 in their app, they need to ensure that the BrowserTabActivity is exported and is true. Otherwise, we cannot get back the token.

msal/build.gradle
somalaya: This defines whether the configuration can be resolved or not. If set to true, the dependencies of the task also cannot be resolved.

This is what I found online:

" A configuration that can be resolved is a configuration for which we can compute a dependency graph, because it contains all the necessary information for resolution to happen. That is to say we’re going to compute a dependency graph, resolve the components in the graph, and eventually get artifacts. A configuration which has canBeResolved set to false is not meant to be resolved. Such a configuration is there only to declare dependencies. The reason is that depending on the usage (compile classpath, runtime classpath), it can resolve to different graphs. It is an error to try to resolve a configuration which has canBeResolved set to false. To some extent, this is similar to an abstract class (canBeResolved=false) which is not supposed to be instantiated, and a concrete class extending the abstract class (canBeResolved=true). A resolvable configuration will extend at least one non-resolvable configuration (and may extend more than one)."

PR #1885, opened 2023-08-17
Add UI (in testapp) for setting active broker in cache

rpdome:
This will allow us to force set the active broker cache.

With this option, this will make us much easier to test the passthrough mode. 
(e.g. if Active broker is LTW, I can now force MSAL to make a request to AuthApp. AuthApp will trigger passthrough and forward the request to LTW to process).

Screenshot 2023-08-17 at 6 12 36 PM
Screenshot 2023-08-17 at 6 12 41 PM

### DIFF
--- a/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
+++ b/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
@@ -58,7 +58,7 @@ import com.microsoft.identity.common.components.AndroidPlatformComponentsFactory
 import com.microsoft.identity.common.internal.activebrokerdiscovery.BrokerDiscoveryClientFactory;
 import com.microsoft.identity.common.internal.broker.BrokerData;
 import com.microsoft.identity.common.internal.cache.ClientActiveBrokerCache;
-import com.microsoft.identity.common.internal.cache.IActiveBrokerCache;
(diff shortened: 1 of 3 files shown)


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
```diff
@@ -203,6 +205,23 @@ public void onClick(View v) {
             }
         });
 
+        mKnownBrokerApps = view.findViewById(R.id.spinner_knownBrokerApps);
+        mSetCachedActiveBrokerButton = view.findViewById(R.id.btn_setCachedActiveBroker);
+
+        final List<BrokerData> debugBrokers = new ArrayList<>(BrokerData.getDebugBrokers());
```
melissaahn, line 211: Seems like all the brokers, regardless of what's currently installed, are added to the spinner. Are there specific cases where you'd want to force set an active broker which is not installed onto the device?

melissaahn (approved):

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
rpdome: Uninstallation scenario 
(SDK cached a broker app that is already uninstalled)

fadidurah (approved):

PR #1886, opened 2023-08-18
Bump daily library versions to 1.0.X

fadidurah:
Bump dependencies that use 0.0.+ for our library versions to now use 1.0.+

Remove Ymls for pipelines that have been deleted


iamgusain (approved):

changelog
```diff
@@ -2,6 +2,7 @@ MSAL Wiki : https://github.com/AzureAD/microsoft-authentication-library-for-andr
 V.NEXT
 ----------
 - [MINOR] MSA UI tests for Brokered Auth (#1856)
+- [MINOR] Daily dev version dependency now uses 1.0.+ (#1886)
```
shahzaibj: Probably doesn't need a changelog since this doesn't ship to production

shahzaibj (approved):

changelog
fadidurah: Yeah... I figured in case partners read this and they were using our daily versions... But probably doesn't make sense for customers to see this. I'll remove it

PR #1892, opened 2023-08-31
Use Active Broker cache specifically for Client SDK

rpdome:
Related: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2164

### DIFF
--- a/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
+++ b/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
@@ -180,7 +180,7 @@ public class AcquireTokenFragment extends Fragment {
         mDebugBrokers.setTextOn("Debug Brokers");
         mDebugBrokers.setChecked(BrokerData.getShouldTrustDebugBrokers());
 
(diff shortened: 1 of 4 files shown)


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
```diff
@@ -180,7 +180,7 @@ public void onClick(View v) {
         mDebugBrokers.setTextOn("Debug Brokers");
         mDebugBrokers.setChecked(BrokerData.getShouldTrustDebugBrokers());
 
-        mCache = ClientActiveBrokerCache.Companion.getCache(
+        mCache = ClientActiveBrokerCache.getClientSdkCache(
```
somalaya, line 183: would it not be ClientActiveBrokerCache.Companion.getClientSdkCache ? given that the new methods are also inside Companion  block?

rpdome: Any companion method marked with JvmStatic doesn't need companion when invoked.

(In kotlin, companion ~= static)

somalaya (approved):

mohitc1 (approved):

PR #1895, opened 2023-09-07
Adding webauthn_capable property option for MSAL config files

melissaahn:
### Summary
This PR adds logic to handle an MSAL config file property `webauthn_capable` with a boolean value.
When the property is set to true, MSAL adds the `webauthn=1` parameter to the extra query string parameter list for interactive token calls. Once the server side implementation handling the parameter is in prod, adding this extra query parameter will make a passkey option appear on the sign in page in browser and WebView. 
The property is set to false by default, as host apps (only 1P apps) should set this to true once they can confirm the future passkey feature works with their app. 

### Testing
I used msalTestApp and tested with a test flight of ESTS and some injected JavaScript code (which mimics the work that still needs to be done on the ESTS UX side). I set the config property to "true" in the WEBVIEW test config file, and when using this config file, the passkey option correctly appeared. The option does not appear when this config property isn't set. 
Note that this PR does not include the logic for the test flight nor the JavaScript code, so this property will add the query string parameter, but the server side does nothing about it at the moment. 

### Related PRs
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2171
- https://github.com/AzureAD/ad-accounts-for-android/pull/2527

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
@@ -418,7 +418,7 @@ public class PublicClientApplicationConfiguration {
     }
 
     public Boolean isWebauthnCapable() {
-        return webauthnCapable != null ? webauthnCapable : false;
+        return webauthnCapable != null && webauthnCapable;
     }
 
     public Authority getDefaultAuthority() {
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
@@ -418,7 +418,7 @@ public class PublicClientApplicationConfiguration {
     }
 
     public Boolean isWebauthnCapable() {
-        return webauthnCapable != null && webauthnCapable;
+        return webauthnCapable == Boolean.TRUE;
     }
 
     public Authority getDefaultAuthority() {
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
@@ -42,7 +42,6 @@ import com.microsoft.identity.client.configuration.AccountMode;
 import com.microsoft.identity.client.configuration.HttpConfiguration;
 import com.microsoft.identity.client.configuration.LoggerConfiguration;
 import com.microsoft.identity.client.exception.MsalClientException;
-import com.microsoft.identity.client.internal.MsalUtils;
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.adal.internal.AuthenticationSettings;
 import com.microsoft.identity.common.java.authorities.Authority;
@@ -82,6 +81,7 @@ import static com.microsoft.identity.client.PublicClientApplicationConfiguration
 import static com.microsoft.identity.client.PublicClientApplicationConfiguration.SerializedNames.REQUIRED_BROKER_PROTOCOL_VERSION;
 import static com.microsoft.identity.client.PublicClientApplicationConfiguration.SerializedNames.TELEMETRY;
 import static com.microsoft.identity.client.PublicClientApplicationConfiguration.SerializedNames.USE_BROKER;
+import static com.microsoft.identity.client.PublicClientApplicationConfiguration.SerializedNames.WEBAUTHN_CAPABLE;
 import static com.microsoft.identity.client.PublicClientApplicationConfiguration.SerializedNames.WEB_VIEW_ZOOM_CONTROLS_ENABLED;
 import static com.microsoft.identity.client.PublicClientApplicationConfiguration.SerializedNames.WEB_VIEW_ZOOM_ENABLED;
 import static com.microsoft.identity.client.exception.MsalClientException.APP_MANIFEST_VALIDATION_ERROR;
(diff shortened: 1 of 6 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -486,4 +492,28 @@ public static GenerateShrCommandParameters createGenerateShrCommandParameters(
...
+            @NonNull final PublicClientApplicationConfiguration configuration) {
+        if (configuration.isWebauthnCapable()) {
+            final Map.Entry<String, String> webauthnExtraParameter = new AbstractMap.SimpleEntry<>(
+                    FidoConstants.WEBAUTHN_QUERY_PARAMETER_FIELD,
+                    FidoConstants.WEBAUTHN_QUERY_PARAMETER_VALUE);
+            if (queryStringParameters == null) {
+                return Collections.singletonList(webauthnExtraParameter);
+            }
+            if (!queryStringParameters.contains(webauthnExtraParameter)) {
+                queryStringParameters.add(webauthnExtraParameter);
```
mohitc1: >queryStringParameters [](http://example.com/codeflow?start=16&length=21)

If the input list is singletonList() then .add() would fail. Safer to create a new list here.

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -486,4 +492,28 @@ public static GenerateShrCommandParameters createGenerateShrCommandParameters(
...
     }
+
+    /**
+     * Adds additional query string parameters to original list.
+     * @param queryStringParameters extra query string parameters inputted.
+     * @param configuration configuration created from MSAL config file.
+     * @return combined list of query string parameters.
+     */
+    @Nullable
+    private static List<Map.Entry<String, String>> getExtraQueryStringParameters(
```
rpdome: nit: appendWebAuthNToExtraQueryParameters?

rpdome (approved):

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
rpdome: ~~i think it should be fine with the null check + early return above.~~ for some reason i thought null was mentioned above....

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
melissaahn: what about appendToExtraQueryParametersIfWebAuthnCapable? Cause we only append it if the property is set.

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
melissaahn: Hmm just checked AcquireTokenParameters, and looks like mExtraQueryStringParameters is a List, meaning a developer could pass a singletonList in with an entry not related to WebAuthn... and that would lead to an exception thrown here.
Just to be safe, I'll create a new ArrayList here, and I'll also return an ArrayList instead of a SingletonList in the null check in case there are any `add` calls applied to it in the future.

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -486,4 +492,34 @@ public static GenerateShrCommandParameters createGenerateShrCommandParameters(
...
+            @Nullable final List<Map.Entry<String, String>> queryStringParameters,
+            @NonNull final PublicClientApplicationConfiguration configuration) {
+        if (configuration.isWebauthnCapable()) {
+            final Map.Entry<String, String> webauthnExtraParameter = new AbstractMap.SimpleEntry<>(
+                    FidoConstants.WEBAUTHN_QUERY_PARAMETER_FIELD,
+                    FidoConstants.WEBAUTHN_QUERY_PARAMETER_VALUE);
+            if (queryStringParameters == null) {
+                return new ArrayList<>(Collections.singletonList(webauthnExtraParameter));
+            }
+            if (!queryStringParameters.contains(webauthnExtraParameter)) {
```
mohitc1, line 513: >!queryStringParameters.contains(webauthnExtraParameter) [](http://example.com/codeflow?start=16&length=55)

Q: Can input query parameters contain webauthn extra parameter as valid contract to pass the webauthn param value? 
Technically, developer can pass it. If they do, do you want to prefer that over config set (if set)?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -409,6 +417,10 @@ public Boolean authorizationInCurrentTask() {
         return isAuthorizationInCurrentTask;
     }
 
+    public Boolean isWebauthnCapable() {
+        return webauthnCapable != null && webauthnCapable;
```
shahzaibj: Could probably just do `webauthnCapable == Boolean.TRUE`

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
melissaahn: Yes, it's valid for the developer to pass in the WebAuthn param value into extraQP in AcquireTokenParameters. Though unless they have a good reason to only pass through the parameters (like their own flighting maybe), it'll be recommended to set the property in MSAL config file so that the developer doesn't have to repeatedly add the query parameter on all acquireToken calls.
Like you said, since an app can decide to set both, the above conditional just leaves the List as is in that case.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
melissaahn: Done: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1895/commits/0945c6818094fa5624bcfc7129deeb9249ef4d55

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
melissaahn: See this commit: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1895/commits/fa3f57091d9cca77998282688006513205068b5a Added some unit tests as well.

mohitc1 (approved):

PR #1896, opened 2023-09-07
Fix Simple Update Tests

fadidurah:
Solution in this PR https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1894 seemed ineffective, issue was actually with how the tests were written.

Success run: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1169229&view=ms.vss-test-web.build-test-results-tab

image

### DIFF
--- a/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/msalonly/update/TestCaseUpdateCompanyPortal.kt
+++ b/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/msalonly/update/TestCaseUpdateCompanyPortal.kt
@@ -47,13 +47,13 @@ import java.util.*
 @LTWTests
 class TestCaseUpdateCompanyPortal : AbstractMsalCustomBrokerInstallationTest() {
 
-    private val mCompanyPortal: BrokerCompanyPortal = installOldCompanyPortal()
-
     @Test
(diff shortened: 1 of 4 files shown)


shahzaibj (approved):

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/msalonly/update/TestCaseUpdateCompanyPortal.kt
```diff
@@ -47,14 +47,14 @@ import java.util.*
 @LTWTests
 class TestCaseUpdateCompanyPortal : AbstractMsalCustomBrokerInstallationTest() {
 
-    private val mCompanyPortal: BrokerCompanyPortal = installOldCompanyPortal()
```
somalaya, line 50: was it not installing the old CP apk correctly when this step executes?

fadidurah: It was installing it before the test rules would run, so if these tests were ran before any others, the copy file rule would not have been ran at any point, and they would fail because the files are not where they need to be.

somalaya (approved):

PR #1897, opened 2023-09-08
Base work for the CIAM features

SaurabhMSFT:
The code in this pull request is for the base work needed to support Signup, SignIn and SSPR features for Native Auth. Any class that is used by at least two of the features has been included in this pull request.

This pull request introduces kotlin co routines libraries as dependencies for msal and its associated test project. Further mockk,  mockito  libraries have been added as dependencies for test projects.

A companion PR in the MSAL Common repository is https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2166


msal/build.gradle
```diff
@@ -221,11 +226,20 @@ dependencies {
...
         exclude module: 'common'
         exclude group: 'com.microsoft.identity', module: 'LabApiUtilities'
     }
+    testImplementation "org.powermock:powermock-module-junit4:$rootProject.ext.powerMockVersion"
+    testImplementation project(':testutils')
+    testImplementation("org.mockito.kotlin:mockito-kotlin:$rootProject.ext.mockitoKotlinVersion") {
+        exclude group: 'org.jetbrains.kotlin', module: 'kotlin-stdlib'
+    }
+
+    testImplementation "io.mockk:mockk:1.11.0"
```
SammyO: version nr in versions.gradle?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -969,10 +1165,6 @@ private static void validateAccountModeConfiguration(@NonNull final PublicClient
                     "Warning! B2C applications should use MultipleAccountPublicClientApplication. "
                             + "Use of SingleAccount mode with multiple IEF policies is unsupported."
             );
-
-            if (config.getAuthorities().size() > 1) {
```
SammyO: I don't think we should have removed this. I just looked through the code and I can't find a reason why we would have. Let's put it back.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -2265,10 +2465,24 @@ void performMigration(@NonNull final TokenMigrationCallback callback) {
         }
     }
 
-    private static void runOnBackground(@NonNull final Runnable runnable) {
+    protected static void runOnBackground(@NonNull final Runnable runnable) {
```
SammyO: you can revert this - it's no longer used

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -2265,10 +2465,24 @@ void performMigration(@NonNull final TokenMigrationCallback callback) {
...
         new Thread(runnable).start();
     }
 
+    protected static void runOnBackground(@NonNull final Runnable runnable, INativeAuthApplicationCreatedListener listener) {
```
SammyO: you can remove this - it's not used anymore

msal/src/test/java/com/microsoft/identity/client/utils/MockApi.kt
```diff
@@ -0,0 +1,59 @@
+package com.microsoft.identity.client.utils
```
fadidurah: License

msal/src/test/java/com/microsoft/identity/client/MainDispatcherRule.kt
```diff
@@ -0,0 +1,21 @@
+package com.microsoft.identity.client
```
fadidurah: License

msal/src/test/java/com/microsoft/identity/client/utils/MockApiUtils.kt
```diff
@@ -0,0 +1,63 @@
+package com.microsoft.identity.client.utils
```
fadidurah: License

msal/src/test/java/com/microsoft/identity/client/utils/MockApiResponseType.kt
```diff
@@ -0,0 +1,49 @@
+package com.microsoft.identity.client.utils
```
fadidurah: License

msal/src/test/java/com/microsoft/identity/client/utils/MockApiEndpointType.kt
```diff
@@ -0,0 +1,15 @@
+package com.microsoft.identity.client.utils
```
fadidurah: License

msal/build.gradle
```diff
@@ -191,6 +191,9 @@ task sourcesJar(type: Jar) {
 String commonVersion = project.hasProperty("distCommonVersion") ? project.distCommonVersion : "1.0.+"
 
 dependencies {
+    //implementation "androidx.core:core-ktx:$rootProject.ext.coreKtxVersion"
+    //testImplementation project(path: ':common4j')
```
fadidurah: Let's remove these comments

msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -129,7 +129,7 @@ public String getTenantId() {
     }
 
     @NonNull
-    String getHomeAccountId() {
+    public String getHomeAccountId() {
```
fadidurah, line 132: Is this necessary? for this and `getEnvironment`

msal/src/main/java/com/microsoft/identity/client/INativeAuthPublicClientApplication.kt
```diff
@@ -0,0 +1,32 @@
...
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.client
+
+
+/**
+ * An interface that contains list of operations that are available when MSAL is in Native Auth
+ * mode.
+ */
+interface INativeAuthPublicClientApplication : IPublicClientApplication {
```
fadidurah, line 31: We should probably populate this, seems unnecessary like this.

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
```diff
@@ -260,6 +261,21 @@ interface IMultipleAccountApplicationCreatedListener {
         void onError(final MsalException exception);
     }
 
+    /**
+     * Listener callback for asynchronous initialization of INativeAuthPublicClientApplication object.
+     */
+    interface INativeAuthApplicationCreatedListener {
```
fadidurah: Wondering if this could be moved to `NativeAuthPublicClientApplication` or `INativeAuthPublicClientApplication`, nothing binding, feel free to ignore this

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
```diff
@@ -0,0 +1,111 @@
...
+import com.microsoft.identity.common.java.providers.microsoft.azureactivedirectory.AzureActiveDirectory
+import com.microsoft.identity.common.java.util.ResultFuture
+import com.microsoft.identity.common.java.util.checkAndWrapCommandResultType
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+class NativeAuthPublicClientApplication(
```
fadidurah, line 65: This class should have documentation

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
```diff
@@ -0,0 +1,111 @@
...
+    @Throws(MsalClientException::class)
+    private fun initializeApplication() {
+        val context = nativeAuthConfig.appContext
+
+        setupTelemetry(context, nativeAuthConfig)
+        AzureActiveDirectory.setEnvironment(nativeAuthConfig.environment)
+        Authority.addKnownAuthorities(nativeAuthConfig.authorities)
+        initializeLoggerSettings(nativeAuthConfig.loggerConfiguration)
+
+        // TODO initializeTokenSharingLibrary(); ?
```
fadidurah: I'm not sure if TSL is in scope, if not, we should remove this TODO

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    @SerializedName(NativeAuthSerializedNames.CHALLENGE_TYPES)
+    private var challengeTypes: List<String>? = null
+
+    @SerializedName(NativeAuthSerializedNames.USE_REAL_AUTHORITY)
+    var useRealAuthority: Boolean? = null
+
+    @SerializedName(NativeAuthSerializedNames.DC)
+    var dc: String? = null
```
fadidurah, line 72: Maybe documentation would be good for these three fields.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
```diff
@@ -0,0 +1,204 @@
...
+        // Call super.mergeConfiguration to handle base configuration fields
+        super.mergeConfiguration(config)
+
+        // Native-auth specific account mode check. If dev passes MULTIPLE, we still want to set it so
+        // we can throw an exception to clarify that MULTIPLE account mode can't be used with Native
+        // Auth.
+        accountMode = if (config.accountMode != null) config.accountMode else accountMode
+
+        // Handle Native Auth specific fields
+        challengeTypes = if (config.challengeTypes == null) challengeTypes else config.challengeTypes
```
fadidurah, line 92: since useRealAuthority and dc were added, we should probably add them to the mergeConfiguration method

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    /**
+     * Override base validateConfiguration() method as native auth validation is different
+     * Had to make this function private as companion object in
```
fadidurah:
```suggestion
     * Had to make this function public as companion object in
```

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
```diff
@@ -0,0 +1,204 @@
...
+                MsalClientException.NATIVE_AUTH_USE_WITHOUT_CLIENT_ID_ERROR_MESSAGE
+            )
+        }
+
+        // If we are given a redirectUri, we should run the superclass validation method to make sure
+        // we can use web auth
+        if (redirectUri != null) {
+            super.validateConfiguration()
+        } else {
+            LogSession.log(TAG, Logger.LogLevel.WARN, "No redirect URI was passed. You cannot use browser-based auth flow to handle BrowserRequired error.")
```
fadidurah: I'm not sure if this log message needs to be updated, mainly because we're referencing BrowserRequired.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -335,6 +334,15 @@ public AccountMode getAccountMode() {
         return this.mAccountMode;
     }
 
+    /**
+     * Sets the account mode id for the PublicClientApplication.
```
fadidurah: account mode id?

msal/src/test/java/com/microsoft/identity/client/MainDispatcherRule.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.test.TestDispatcher
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
+import kotlinx.coroutines.test.resetMain
+import kotlinx.coroutines.test.setMain
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+
+class MainDispatcherRule(
```
fadidurah: Let's add a comment here for clarity

msal/build.gradle
```diff
@@ -208,6 +211,8 @@ dependencies {
     }
     implementation "cz.msebera.android:httpclient:$rootProject.ext.mseberaApacheHttpClientVersion"
     implementation "androidx.constraintlayout:constraintlayout:$rootProject.ext.constraintLayoutVersion"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$rootProject.ext.kotlinXCoroutinesVersion"
```
shahzaibj, line 211: Please add PR description and in the description also mention why you are introducing new dependencies to MSAL Android.

msal/src/main/java/com/microsoft/identity/client/Account.java
shahzaibj: Please describe the reasons for making such changes in the PR description

msal/src/main/java/com/microsoft/identity/client/AuthenticationResultAdapter.java
```diff
@@ -35,11 +33,13 @@
...
 
-class AuthenticationResultAdapter {
+import androidx.annotation.NonNull;
+
+public class AuthenticationResultAdapter {
 
     private static final String TAG = AuthenticationResultAdapter.class.getName();
 
-    static IAuthenticationResult adapt(@NonNull final ILocalAuthenticationResult localAuthenticationResult) {
+    public static IAuthenticationResult adapt(@NonNull final ILocalAuthenticationResult localAuthenticationResult) {
```
shahzaibj, line 46: Why?

msal/src/main/java/com/microsoft/identity/client/INativeAuthPublicClientApplication.kt
```diff
@@ -0,0 +1,32 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.client
+
+
+/**
+ * An interface that contains list of operations that are available when MSAL is in Native Auth
+ * mode.
```
shahzaibj: Is Native Auth actually a mode? How does one configure MSAL to be in that mode?

msal/src/main/java/com/microsoft/identity/client/INativeAuthPublicClientApplication.kt
```diff
@@ -0,0 +1,32 @@
...
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.client
+
```
shahzaibj, line 24: nit: extra line

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
shahzaibj: It really should

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
```diff
@@ -260,6 +261,21 @@ interface IMultipleAccountApplicationCreatedListener {
...
+        /**
+         * Called once an IMultipleAccountPublicClientApplication is successfully created.
```
shahzaibj:
```suggestion
         * Called once an INativeAuthPublicClientApplication is successfully created.
```

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
```diff
@@ -260,6 +261,21 @@ interface IMultipleAccountApplicationCreatedListener {
...
+         */
+        void onCreated(final @NonNull INativeAuthPublicClientApplication application);
+
+        /**
+         * Called once IMultipleAccountPublicClientApplication can't be created.
```
shahzaibj:
```suggestion
         * Called once INativeAuthPublicClientApplication can't be created.
```

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
```diff
@@ -0,0 +1,111 @@
...
+    companion object {
+        /**
+         * Name of the shared preference cache for storing NativeAuthPublicClientApplication data.
+         */
+        private const val NATIVE_AUTH_CREDENTIAL_SHARED_PREFERENCES =
+            "com.microsoft.identity.client.native_auth_credential_cache"
+
+        internal val TAG = NativeAuthPublicClientApplication::class.java.toString()
+
+        val pcaScope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
```
shahzaibj, line 88: comments

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Throws(MsalClientException::class)
+    private fun initializeApplication() {
+        val context = nativeAuthConfig.appContext
+
+        setupTelemetry(context, nativeAuthConfig)
```
shahzaibj: Do we have a conclusion on telemetry?

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
```diff
@@ -0,0 +1,111 @@
...
+
+class NativeAuthPublicClientApplication(
+    private val nativeAuthConfig: NativeAuthPublicClientApplicationConfiguration
+) : INativeAuthPublicClientApplication, PublicClientApplication(nativeAuthConfig) {
+
+    private lateinit var sharedPreferencesFileManager: SharedPreferencesFileManager
+
+    init {
+        initializeApplication()
+        initializeSharedPreferenceFileManager(nativeAuthConfig.appContext)
```
shahzaibj, line 73: I suspect you need this because you want to operate in SINGLE account mode

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
```diff
@@ -0,0 +1,204 @@
...
+ * app implementation.
+ */
+@Accessors(prefix = ["m"])
+@Getter
+class NativeAuthPublicClientApplicationConfiguration :
+    PublicClientApplicationConfiguration(),
+    Serializable {
+    companion object {
+        private val TAG = NativeAuthPublicClientApplicationConfiguration::class.java.simpleName
+        private val VALID_CHALLENGE_TYPES = listOf("password", "oob", "redirect")
```
shahzaibj: Each challenge type needs to be defined as a constant somewhere....I think I left similar feedback in the common PR too. 

Once you've defined them there then you can just reference those consts here

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    private object NativeAuthSerializedNames {
+        const val CHALLENGE_TYPES = "challenge_types"
+        const val USE_REAL_AUTHORITY = "useRealAuthority"
```
shahzaibj: Let's follow one naming pattern for the values of Serialized Names (underscores vs camel case). I think underscore is used in most places so let's stick to that.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
shahzaibj, line 72: I'm not sure if we need to expose DC here at the config level. I believe this info is already present in the Authority that is part of the config

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
```diff
@@ -0,0 +1,204 @@
...
+    fun setChallengeTypes(challengeTypes: List<String>?) {
+        this.challengeTypes = challengeTypes
+    }
+
+    fun mergeConfiguration(config: NativeAuthPublicClientApplicationConfiguration) {
+        // Call super.mergeConfiguration to handle base configuration fields
+        super.mergeConfiguration(config)
+
+        // Native-auth specific account mode check. If dev passes MULTIPLE, we still want to set it so
+        // we can throw an exception to clarify that MULTIPLE account mode can't be used with Native
```
shahzaibj, line 87: I guess that exception is being thrown at a different layer?

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
```diff
@@ -0,0 +1,204 @@
...
+            throw MsalClientException(
+                MsalClientException.NATIVE_AUTH_INVALID_ACCOUNT_MODE_CONFIG_ERROR_CODE,
+                MsalClientException.NATIVE_AUTH_INVALID_ACCOUNT_MODE_CONFIG_ERROR_MESSAGE
+            )
+        }
+
+        // We only allow one authority being passed
+        if (authorities == null || authorities.size != 1) {
+            // Throw is no authority passed
+            if (authorities == null || authorities.size == 0) {
```
shahzaibj: You just checked for null on previous line. There is some optimization you can do on these if blocks to remove redundant checks

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
```diff
@@ -0,0 +1,204 @@
...
+                // If not, throw exception
+                throw MsalClientException(
+                    MsalClientException.NATIVE_AUTH_INVALID_CIAM_AUTHORITY_ERROR_CODE,
+                    MsalClientException.NATIVE_AUTH_INVALID_CIAM_AUTHORITY_ERROR_MESSAGE
+                )
+            }
+        }
+
+        // Enforce that shared device mode is disabled
+        if (isSharedDevice) {
```
shahzaibj, line 165: This is checking a value in the config JSON? or actually checking whether the device is in shared mode?

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
```diff
@@ -0,0 +1,204 @@
...
+            }
+        }
+    }
+
+    /**
+     * Overriding this method to add a check for redirect uri. If no uri was passed, we don't need to check this.
+     */
+    @Throws(MsalClientException::class)
+    override fun checkIntentFilterAddedToAppManifestForBrokerFlow() {
+        // TODO wonder if we really even need this, as broker is not used so we might also pass the check anyway.
```
shahzaibj: You still need this. The word "Broker" in that method name is a bit misleading. That method ensures that your app can handle redirection from Browser to your app for authorization code grant flow and you need to do that check if you want to support Browser fallback

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfigurationFactory.kt
```diff
@@ -0,0 +1,262 @@
...
+import com.microsoft.identity.common.java.providers.microsoft.microsoftsts.MicrosoftStsOAuth2Strategy
+import com.microsoft.identity.common.java.providers.microsoft.microsoftsts.MicrosoftStsTokenResponse
+import com.microsoft.identity.msal.R
+import java.io.File
+import java.io.FileInputStream
+import java.io.FileNotFoundException
+import java.io.IOException
+import java.io.InputStream
+
+class NativeAuthPublicClientApplicationConfigurationFactory :
```
shahzaibj, line 56: javadoc

msal/src/main/java/com/microsoft/identity/client/Account.java
SaurabhMSFT: All of the current usages are from files in the same package. We will use it from BaseStates.kt file (will be added in SiginIn PR). That file is not in the same package so we have to change the access modifier from `package` to `public`.

msal/src/main/java/com/microsoft/identity/client/INativeAuthPublicClientApplication.kt
SaurabhMSFT: We will be adding methods to this interface as part of SignIn, SignUp and SSPR PRs. The reason I didn't add the methods here as it would mean adding empty stub methods in `NativeAuthPublicClientApplication` class

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
SaurabhMSFT: For NativeAuth this value is read from the config.

msal/src/main/java/com/microsoft/identity/client/AuthenticationResultAdapter.java
SaurabhMSFT: We are going to use it in SignIn code from a file in different package. We need to the change access modifier from `package` to `public`.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
SaurabhMSFT: @SammyO

SammyO: @shahzaibj thanks for flagging this, this is something I had wanted to discuss with you. 
So this SharedPreferencesFileManager is only used for multi-account flows? We don't need it if native auth is operating in a single account flow?

msal/src/main/java/com/microsoft/identity/client/INativeAuthPublicClientApplication.kt
SaurabhMSFT: I have updated the javadoc

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
SaurabhMSFT: It is thrown in the `validateConfiguration` method .

msal/src/main/java/com/microsoft/identity/client/INativeAuthPublicClientApplication.kt
SaurabhMSFT: Addressed

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
SaurabhMSFT: Addressed

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
SaurabhMSFT: Addressed

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
SaurabhMSFT: @SammyO

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
SaurabhMSFT: Addressed

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfigurationFactory.kt
SaurabhMSFT: Addressed

msal/src/test/java/com/microsoft/identity/client/utils/MockApi.kt
SaurabhMSFT: Addressed

msal/src/test/java/com/microsoft/identity/client/utils/MockApiEndpointType.kt
SaurabhMSFT: Addressed

msal/src/test/java/com/microsoft/identity/client/utils/MockApiResponseType.kt
SaurabhMSFT: Addressed

msal/src/test/java/com/microsoft/identity/client/utils/MockApiUtils.kt
SaurabhMSFT: Addressed

msal/src/main/java/com/microsoft/identity/client/INativeAuthPublicClientApplication.kt
fadidurah: Ok, that's fine

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
fadidurah: Ping on this

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
SammyO: We should remove this, as we're not doing anything with this telemetry right now. We'll add this or something like this to our public preview release.
@SaurabhMSFT  you can remove this call.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
SaurabhMSFT: Addressed

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
SaurabhMSFT: I have removed the TODO

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
SaurabhMSFT: Addressed by removing this function call.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
SaurabhMSFT: Addressed

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
SaurabhMSFT: The DC value is not present in `NativeAuthCIAMAuthority` class. Did you mean any other authority?

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
shahzaibj: @SammyO That shared preference file is for SINGLE account mode (you won't need it if you operate in Multiple mode)

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
shahzaibj: It is present on the parent class i.e. `Authority.java`

msal/build.gradle
shahzaibj: Your PR description says you are only adding new dependencies in tests but that's not what I'm seeing here. Please specify why you need to add this in MSAL.

msal/build.gradle
```diff
@@ -221,11 +223,20 @@ dependencies {
     testImplementation "org.mockito:mockito-inline:$rootProject.ext.mockitoCoreVersion"
     testImplementation ("org.robolectric:robolectric:$rootProject.ext.robolectricVersion")
     testImplementation "androidx.test:core:$rootProject.ext.androidxTestCoreVersion"
+    testImplementation "org.powermock:powermock-module-junit4:$rootProject.ext.powerMockVersion"
```
shahzaibj: Mentioned in other places that you could've considered using Robolectric

msal/src/main/java/com/microsoft/identity/client/AuthenticationResultAdapter.java
```diff
@@ -35,11 +33,13 @@
 import java.util.Locale;
 import java.util.Set;
 
-class AuthenticationResultAdapter {
+import androidx.annotation.NonNull;
+
+public class AuthenticationResultAdapter {
```
shahzaibj, line 42: Javadoc for all public classes

msal/src/main/java/com/microsoft/identity/client/INativeAuthPublicClientApplication.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.microsoft.identity.client.exception.MsalException
+
+/**
+ * INativeAuthPublicClientApplication provides top level interface that is used by dpp developers
+ * to use Native Auth methods
```
shahzaibj:
```suggestion
 * to use Native Auth methods.
```

Periods at the end of each Javadoc

msal/src/main/java/com/microsoft/identity/client/INativeAuthPublicClientApplication.kt
```diff
@@ -0,0 +1,46 @@
...
+ */
+interface INativeAuthPublicClientApplication : IPublicClientApplication {
+    /**
+     * Listener callback for asynchronous initialization of INativeAuthPublicClientApplication object.
+     */
+    interface INativeAuthApplicationCreatedListener {
```
shahzaibj, line 35: Is there a reason you couldn't just use the default `ApplicationCreatedListener`? Seems like a lot of potential for code re-use there and avoid code duplication.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    companion object {
+        /**
+         * Name of the shared preference cache for storing NativeAuthPublicClientApplication data.
+         */
+        private const val NATIVE_AUTH_CREDENTIAL_SHARED_PREFERENCES =
```
shahzaibj, line 80: I guess you could've considered extending your class from SingleAccountPublicClientApplication and that would've allowed you to leverage a lot of the single account functionality that is already built-in into that class.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
```diff
@@ -0,0 +1,214 @@
...
+
+    //List of challenge types supported by the client.
+    //For a complete list of challenge types see [NativeAuthConstants.GrantType]
+    @SerializedName(NativeAuthSerializedNames.CHALLENGE_TYPES)
+    private var challengeTypes: List<String>? = null
+
+    //The mock API authority used for testing will be rejected by validation logic run on
+    // instantiation. This flag is used to bypass those checks in various points in the application
+    @SerializedName(NativeAuthSerializedNames.USE_REAL_AUTHORITY)
+    var useRealAuthority: Boolean? = null
```
shahzaibj, line 67: I don't think this should be exposed in the configuration here.

shahzaibj: If you are already using BuildConfig then you certainly don't need it on the configuration.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
```diff
@@ -0,0 +1,214 @@
...
+                MsalClientException.NATIVE_AUTH_SHARED_DEVICE_MODE_ERROR_CODE,
+                MsalClientException.NATIVE_AUTH_SHARED_DEVICE_MODE_ERROR_MESSAGE
+            )
+        }
+
+        // Enforce that dev cannot set "broker_redirect_uri_registered = true"
+        // This check is technically not necessary, since Native Auth configuration only loads with a
+        // CIAM authority, which does not result in creating a broker controller
+        // Adding it for clarity with devs
+        if (useBroker) {
```
shahzaibj: Checking this really doesn't matter because the default config for MSAL already has set this to true, and you want that if you want to fallback to Browser anyway. You may choose to entirely remove this check.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfigurationFactory.kt
```diff
@@ -0,0 +1,266 @@
...
+import com.microsoft.identity.common.java.providers.microsoft.microsoftsts.MicrosoftStsTokenResponse
+import com.microsoft.identity.msal.R
+import java.io.File
+import java.io.FileInputStream
+import java.io.FileNotFoundException
+import java.io.IOException
+import java.io.InputStream
+
+/**
+ * NativeAuthPublicClientApplicationConfigurationFactory manages the initialisation and loads the
```
shahzaibj:
```suggestion
 * NativeAuthPublicClientApplicationConfigurationFactory manages the initialization and loads the
```

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfigurationFactory.kt
```diff
@@ -0,0 +1,266 @@
...
+                AndroidPlatformComponentsFactory.createFromContext(context)
+            )
+
+            // Set build values if present in configuration
+            initializeBuildValues(developerConfig)
+
+            return config
+        }
+
+        private fun initializeBuildValues(developerConfig: NativeAuthPublicClientApplicationConfiguration?) {
```
shahzaibj, line 132: I think you should consider removing the BuildValues class entirely. I don't think it is needed given that similar data already exists in BuildConfig, Authority, and AadSliceInfo classes

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfigurationFactory.kt
```diff
@@ -0,0 +1,266 @@
...
+                buffer = ByteArray(configStream.available())
+                configStream.read(buffer)
+            } catch (e: IOException) {
+                if (isDefaultConfiguration) {
+                    throw IllegalStateException("Unable to open default native auth configuration file.", e)
+                } else {
+                    throw IllegalArgumentException("Unable to open provided native auth configuration file.", e)
+                }
+            } finally {
+                try {
```
shahzaibj: You may want to use `try-with-resources`. The Kotlin equivalent I believe is called `use`

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfigurationFactory.kt
```diff
@@ -0,0 +1,266 @@
...
+            > {
+            val methodName = ":createCache"
+            Logger.verbose(
+                TAG + methodName,
+                "Creating MsalOAuth2TokenCache"
+            )
+            return MsalOAuth2TokenCache.create(components)
+        }
+
+        private fun getGsonForLoadingConfiguration(): Gson {
```
shahzaibj, line 237: If such a method already exists in the codebase then can you just leverage that?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -791,6 +795,158 @@ public static ISingleAccountPublicClientApplication createSingleAccountPublicCli
...
+     *                    the running activity's context, which could potentially make the sdk hold a
+     *                    strong reference to the activity, thus preventing correct garbage
+     *                    collection and causing bugs.
+     *                    </p>
+     * @param clientId    The application client id. Cannot be null.
+     * @param authority   The default authority to be used for the authority. If this is null, the default authority will be used.
+     * @param redirectUri The redirect URI of the application.
+     * @return An instance of INativeAuthPublicClientApplication.
+     */
+    public static INativeAuthPublicClientApplication createNativeAuthPublicClientApplication(
```
shahzaibj, line 922: I see you are adding a lot of overloads but do you actually need them? I think following the same pattern as existing MSAL code sounds good but you should consider whether you really have a need to provide all these? Any code you write is something you will have to support for a long time so you should consider that since you are offering a new feature then can you choose to provide just one or max 2 constructors so that you only have to supported those limited number of constructors in the future.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -959,6 +1119,43 @@ public void onCancel() {
         CommandDispatcher.submitSilent(command);
     }
 
+    private static NativeAuthPublicClientApplication createNativeAuthApplication(@NonNull final NativeAuthPublicClientApplicationConfiguration config,
```
shahzaibj, line 1122: Left comment above about these constructor overloads

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
```diff
@@ -335,6 +334,15 @@ public AccountMode getAccountMode() {
         return this.mAccountMode;
     }
 
+    /**
+     * Sets the account mode for the PublicClientApplication.
+     *
+     * @param accountMode the account mode.
+     */
+    public void setAccountMode(final AccountMode accountMode) {
```
shahzaibj, line 342: Why do we need this method? This seems new addition to existing MSAL product?

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
```diff
@@ -505,4 +507,10 @@ public static void throwOnMainThread(final String methodName) {
             throw new IllegalStateException("method: " + methodName + " may not be called from main thread.");
         }
     }
+
+    public static void throwOnMainThread(final BaseException exception) {
```
shahzaibj, line 517: Javadoc for all public methods

msal/src/main/java/com/microsoft/identity/client/statemachine/Error.kt
```diff
@@ -0,0 +1,99 @@
...
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+package com.microsoft.identity.client.statemachine
+
+sealed class Error(
```
shahzaibj, line 26: Javadoc for all classes

msal/src/main/java/com/microsoft/identity/client/statemachine/Error.kt
```diff
@@ -0,0 +1,99 @@
...
+    open val error: String? = null,
+    open val errorMessage: String?,
+    open val correlationId: String,
+    open var exception: Exception? = null,
+    open val errorCodes: List<Int>? = null
+)
+
+class GeneralError(
```
shahzaibj: Javadoc here and everywhere

msal/src/main/java/com/microsoft/identity/client/statemachine/Error.kt
shahzaibj, line 26: Are we always sending back Error objects for error in Native Auth or are we still throwing Exceptions in certain cases? Either way, why?

msal/src/main/res/raw/msal_native_auth_default_config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "broker_redirect_uri_registered": false,
```
shahzaibj, line 2: Interesting. I think this is needed to be **true** for proper redirect uri validation.

msal/src/test/java/com/microsoft/identity/client/utils/MockApi.kt
```diff
@@ -0,0 +1,81 @@
...
+
+import com.google.gson.annotations.SerializedName
+import com.microsoft.identity.common.java.net.HttpConstants
+import com.microsoft.identity.common.java.net.UrlConnectionHttpClient
+import com.microsoft.identity.common.java.util.ObjectMapper
+import org.junit.Assert.assertTrue
+import java.net.URL
+import java.util.TreeMap
+
+class MockApi private constructor(
```
shahzaibj: Javadoc

shahzaibj: I saw either the exact or very similar class in common PR too I believe

msal/src/test/java/com/microsoft/identity/client/utils/MockApiEndpointType.kt
```diff
@@ -0,0 +1,37 @@
...
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.client.utils
+
+enum class MockApiEndpointType(val stringValue: String) {
```
shahzaibj: Duplicated in common

msal/src/test/java/com/microsoft/identity/client/utils/MockApiResponseType.kt
```diff
@@ -0,0 +1,71 @@
...
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.client.utils
+
+enum class MockApiResponseType(val stringValue: String) {
```
shahzaibj: Duplicated in common

msal/src/test/java/com/microsoft/identity/client/utils/MockApiUtils.kt
```diff
@@ -0,0 +1,85 @@
...
+//  THE SOFTWARE.
+package com.microsoft.identity.client.utils
+
+import com.microsoft.identity.common.java.logging.DiagnosticContext
+import com.microsoft.identity.common.java.logging.IRequestContext
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+import org.powermock.reflect.Whitebox
+
+class MockApiUtils {
```
shahzaibj: Possibly Duplicated in common

msal/src/test/res/raw/native_auth_native_only_test_config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "client_id" : "0e565df2-c350-47f2-9cd4-c6c7a464b8bf",
+  "authorities" : [
+    {
+      "type": "CIAM",
```
shahzaibj, line 5: Do you want AAD_NA instead?

msal/src/test/res/raw/native_auth_native_only_test_config.json
```diff
@@ -0,0 +1,10 @@
...
+      "authority_url": "https://ciamtestandroid.ciamlogin.com/ciamtestandroid.onmicrosoft.com"
+    }
+  ],
+  "challenge_types" : ["oob", "password"]
+}
```
shahzaibj: nit: new line ending

shahzaibj (changes requested):
Left comments

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
SammyO: @shahzaibj I think this relates to a conversation we had before: native auth will be in single mode only for now, but there is no use of broker. Is this needed then? Practically in our code currently we don't use it, so I'm inclined to remove it, but I'm not sure if I'm overlooking something.
What does normal MSAL use this for in single mode?

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
SammyO: This is a flag that should be settable by the calling application (e.g. our sample app makes use of the mock API flows), and AFAIK there is no way for a calling app to change the library's BuildConfig

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
SammyO: No, because our native auth version of `getCurrentAccount()` (part of a next PR) would conflict with the existing `getCurrentAccount()` (and we can't reuse the existing one, as the logic in ours is different). There would have been limited to no reusability of code benefit if we had extended `SingleAccountPublicClientApplication` instead of what we do now.

fadidurah (approved):
My comments are approved

msal/src/main/java/com/microsoft/identity/client/AuthenticationResultAdapter.java
SaurabhMSFT: Addressed

msal/src/main/java/com/microsoft/identity/client/INativeAuthPublicClientApplication.kt
SaurabhMSFT: Addressed

msal/src/main/java/com/microsoft/identity/client/INativeAuthPublicClientApplication.kt
SaurabhMSFT: I will add method to `INativeAuthPublicClientApplication` class in subsequent pull requests. `ApplicationCreatedListener` makes a callback with `IPublicClientApplication` which will have to cast every time. That's why we have avoided using that interface.

msal/src/test/res/raw/native_auth_native_only_test_config.json
SaurabhMSFT: New line added

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
SaurabhMSFT: I have removed the check and improved the comment.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfigurationFactory.kt
SaurabhMSFT: Fixed

msal/src/test/java/com/microsoft/identity/client/utils/MockApi.kt
SaurabhMSFT: This file has been removed.

SaurabhMSFT: This file has been removed.

msal/src/test/java/com/microsoft/identity/client/utils/MockApiEndpointType.kt
SaurabhMSFT: This file has been removed.

msal/src/test/java/com/microsoft/identity/client/utils/MockApiResponseType.kt
SaurabhMSFT: This file has been removed.

msal/src/test/java/com/microsoft/identity/client/utils/MockApiUtils.kt
SaurabhMSFT: This file has been removed.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfigurationFactory.kt
SaurabhMSFT: I have used `use`

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfigurationFactory.kt
SaurabhMSFT: This style of code is present in multiple places where it's used with different custom deserializers. There is no existing method that can be used.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfigurationFactory.kt
SaurabhMSFT: I have created a separate ticket for it.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
SaurabhMSFT: I have created a separate ticket for removing BuildValues file. I will take this as part of that ticket.

msal/build.gradle
SaurabhMSFT: I will have to see how long it will take to change the framework. The test cases itself will be part of later pull requests. I will remove this if we don't end up using it.

msal/build.gradle
SaurabhMSFT: I have fixed the description.

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
SaurabhMSFT: fixed

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
SaurabhMSFT: fixed

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplicationConfiguration.kt
SaurabhMSFT: fixed

msal/src/main/java/com/microsoft/identity/client/internal/MsalUtils.java
SaurabhMSFT: fixed

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
SaurabhMSFT: Our aim is to provide multiple options to the 3rd party developers. We will be okay to maintain these three static constructor methods.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
SaurabhMSFT: Responded above

msal/src/main/java/com/microsoft/identity/client/PublicClientApplicationConfiguration.java
SaurabhMSFT: It's used in the mergeConfiguration.

msal/src/main/res/raw/msal_native_auth_default_config.json
SaurabhMSFT: I have made the change. @SammyO  FYI

msal/src/main/java/com/microsoft/identity/client/statemachine/Error.kt
SaurabhMSFT: added

msal/src/main/java/com/microsoft/identity/client/statemachine/Error.kt
SaurabhMSFT: added

msal/src/main/java/com/microsoft/identity/client/statemachine/Error.kt
SaurabhMSFT: we are sending back error objects. @SammyO  Can you provide the reason for this?

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shahzaibj: Is there a reason for that aim? Do you have customer demand for these overloads?

msal/src/main/res/raw/msal_native_auth_default_config.json
SammyO: Hold on @SaurabhMSFT. Redirect URI should be optional, so wouldn't making this true make it in effect also required? @shahzaibj

msal/src/main/java/com/microsoft/identity/client/statemachine/Error.kt
SammyO: We're doing both. Any expected error will be returned as an Error object (e.g. PasswordInvalid, GeneralError). If an exception happens in one of the underlying SDK layers, it might also be thrown back to the developer. I.e. we don't have a try-catch in the SDK interface that catches all exceptions and turns them in to Errors.

msal/src/test/res/raw/native_auth_native_only_test_config.json
shahzaibj: Is this addressed?

msal/src/main/res/raw/msal_native_auth_default_config.json
shahzaibj: @SammyO That depends on how you evaluate these things when you do your own validation. For Android MSAL, redirect URI is always required. 

This particular parameter is just checking if redirect URI is of the correct format that support brokered auth.

msal/src/main/java/com/microsoft/identity/client/statemachine/Error.kt
shahzaibj: This looks like putting too much work on the developers. They have to be on the look out for both Errors and Exceptions. I'll rather prefer having to worry about only one of these but its up to you.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
shahzaibj: Can you elaborate more on that? I mean it seems you need Single Account mode but you can't use SingeAccountModePCA - I'm not sure I understand why.

msal/src/main/java/com/microsoft/identity/client/INativeAuthPublicClientApplication.kt
shahzaibj: This is probably something we could've solved with some refactoring and use of generics.

msal/src/test/res/raw/native_auth_native_only_test_config.json
SaurabhMSFT: This must be CIAM. I checked this in the documentation for private preview.

msal/src/main/res/raw/msal_native_auth_default_config.json
SammyO: @SaurabhMSFT  in that case you can't simply make the change. We'll have to evaluate whether we can or not, as redirect URI should stay optional.

SaurabhMSFT: I have reverted it to false.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
SammyO: What would you like me to expand on? We need a method called `getCurrentAccount()` which already exists in `SingleAccountPublicClientApplication`. It's not ideal that we're not extending this class, but without changing the existing MSAL code and SDK interface I don't see how this can be solved.

msal/src/main/java/com/microsoft/identity/client/INativeAuthPublicClientApplication.kt
SaurabhMSFT: I don't think generics would work here. The user of this callback would need the type of the first parameter in the callback but due to type erasure at compile time, the type information wouldn't be available.

shahzaibj (approved):

PR #1929, opened 2023-10-24
Adding null checks and logging to filter methods in AccountAdapter

melissaahn:
### Summary
There was an [incident](https://portal.microsofticm.com/imp/v3/incidents/incident/434102603/summary) opened up that reported some crashes for devices in SDM due to a NPE thrown within the AccountAdapter class. 
This PR adds handling and logging for any NPE errors thrown in the filter methods. Note that this is not a root cause fix for the incident, but will provide a mitigation.

Update: We also received a PR from a customer who noted they are also seeing crash reports due to the same NPE: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1927

### Related PR
- https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1847/files

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
@@ -54,14 +54,24 @@ class AccountAdapter {
 
         @Override
         public List<ICacheRecord> filter(@NonNull List<ICacheRecord> records) {
+            final String methodTag = GuestAccountFilter.class.getSimpleName() + ":filter";
+
             final List<ICacheRecord> result = new ArrayList<>();
 
             for (final ICacheRecord cacheRecord : records) {
                 final String acctHomeAccountId = cacheRecord.getAccount().getHomeAccountId();
                 final String acctLocalAccountId = cacheRecord.getAccount().getLocalAccountId();
-
-                if (!acctHomeAccountId.contains(acctLocalAccountId)) {
-                    result.add(cacheRecord);
+                try {
+                    if (!acctHomeAccountId.contains(acctLocalAccountId)) {
+                        result.add(cacheRecord);
+                    }
+                } catch (final NullPointerException e) {
+                    if (acctHomeAccountId == null) {
+                        Logger.warn(methodTag, "Home account id is null.");
+                    }
+                    if (acctLocalAccountId == null) {
+                        Logger.warn(methodTag, "Local account id is null.");
+                    }
                 }
             }
 
@@ -77,13 +87,23 @@ class AccountAdapter {
 
         @Override
         public List<ICacheRecord> filter(@NonNull final List<ICacheRecord> records) {
+            final String methodTag = HomeAccountFilter.class.getSimpleName() + ":filter";
             final List<ICacheRecord> result = new ArrayList<>();
 
             for (final ICacheRecord cacheRecord : records) {
                 final String acctHomeAccountId = cacheRecord.getAccount().getHomeAccountId();
                 final String acctLocalAccountId = cacheRecord.getAccount().getLocalAccountId();
-                if (acctLocalAccountId != null && acctHomeAccountId.contains(acctLocalAccountId)) {
-                    result.add(cacheRecord);
+                try {
+                    if (acctHomeAccountId.contains(acctLocalAccountId)) {
+                        result.add(cacheRecord);
+                    }
+                } catch (final NullPointerException e) {
+                    if (acctHomeAccountId == null) {
+                        Logger.warn(methodTag, "Home account id is null.");
+                    }
+                    if (acctLocalAccountId == null) {
+                        Logger.warn(methodTag, "Local account id is null.");
+                    }
                 }
             }
 
@@ -98,8 +118,13 @@ class AccountAdapter {
 
         private boolean hasNoCorrespondingHomeAccount(@NonNull final ICacheRecord guestRecord,
                                                       @NonNull final List<ICacheRecord> homeRecords) {
+            final String methodTag = CacheRecordFilter.class.getSimpleName() + ":hasNoCorrespondingHomeAccount";
             // Init our sought value
             final String guestAccountHomeAccountId = guestRecord.getAccount().getHomeAccountId();
+            if (guestAccountHomeAccountId == null){
+                Logger.warn(methodTag, "Guest account home account id is null.");
+                return true;
+            }
 
             // Create a List of home_account_ids from the homeRecords...
             final List<String> homeAccountIds = new ArrayList<String>() {{
--- a/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
@@ -62,16 +62,16 @@ class AccountAdapter {
             for (final ICacheRecord cacheRecord : records) {
                 final String acctHomeAccountId = cacheRecord.getAccount().getHomeAccountId();
                 final String acctLocalAccountId = cacheRecord.getAccount().getLocalAccountId();
-                boolean notNullorEmpty = true;
+                boolean isNullorEmpty = false;
                 if (StringUtil.isNullOrEmpty(acctHomeAccountId)) {
                     Logger.warn(methodTag, "Home account id is null or empty.");
-                    notNullorEmpty = false;
+                    isNullorEmpty = true;
                 }
                 if (StringUtil.isNullOrEmpty(acctLocalAccountId)) {
                     Logger.warn(methodTag, "Local account id is null or empty.");
-                    notNullorEmpty = false;
+                    isNullorEmpty = true;
                 }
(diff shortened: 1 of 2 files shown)


raffij:
@melissaahn is this taking our pr (https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1927) and using it as an issue report?

melissaahn:
> @melissaahn is this taking our pr (#1927) and using it as an issue report?

No, I am referring to a separate issue reported internally. I wasn't aware of your team's PR initially, so I will link your PR in the description of this one to show that your team has also noted this issue.

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
```diff
@@ -77,13 +87,23 @@ private static class HomeAccountFilter implements CacheRecordFilter {
...
             for (final ICacheRecord cacheRecord : records) {
                 final String acctHomeAccountId = cacheRecord.getAccount().getHomeAccountId();
                 final String acctLocalAccountId = cacheRecord.getAccount().getLocalAccountId();
-                if (acctLocalAccountId != null && acctHomeAccountId.contains(acctLocalAccountId)) {
-                    result.add(cacheRecord);
+                try {
+                    if (acctHomeAccountId.contains(acctLocalAccountId)) {
+                        result.add(cacheRecord);
+                    }
+                } catch (final NullPointerException e) {
```
somalaya: why are we catching the NPE instead of checking if acctHomeAccountId/acctLocalAccountId are null directly?

melissaahn: I thought it could be cleaner if we want to add logging as well.
If we were to check if home account id and local account id are null in the conditional, then we would have to check them individually again to provide the logging statements (since it could be possible that both are null). I also could have moved the lines in the catch block to the beginning of the loop and implemented some sort of boolean flag (to avoid checking for null again) to be checked in the actual conditional, but I opted for a try/catch block because I thought it was better condensed.

melissaahn: Here's the flag option I was thinking about. Is this preferred:
```               
 boolean notNull = true;
  if (acctHomeAccountId == null) {
      Logger.warn(methodTag, "Home account id is null.");
      notNull = false;
  }
  if (acctLocalAccountId == null) {
      Logger.warn(methodTag, "Local account id is null.");
      notNull= false;
  }
  if (notNull && !acctHomeAccountId.contains(acctLocalAccountId)) {
      result.add(cacheRecord);
  }

File: msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java [hunk 2]
```diff
@@ -54,14 +54,24 @@ private static class GuestAccountFilter implements CacheRecordFilter {
...
-                    result.add(cacheRecord);
+                try {
+                    if (!acctHomeAccountId.contains(acctLocalAccountId)) {
+                        result.add(cacheRecord);
+                    }
+                } catch (final NullPointerException e) {
+                    if (acctHomeAccountId == null) {
+                        Logger.warn(methodTag, "Home account id is null.");
+                    }
+                    if (acctLocalAccountId == null) {
```
shahzaibj: Given your code, I think you will NOT get NPE if local account id is null. Only the field on which method is invoked can cause NPE if null (in this case that's home account id)

shahzaibj: I suggest you re-write your code to avoid NPE in the first place (as opposed to causing NPE and then catching it since this particular NPE is easy to prevent)

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
```diff
@@ -98,8 +118,13 @@ public List<ICacheRecord> filter(@NonNull final List<ICacheRecord> records) {
 
         private boolean hasNoCorrespondingHomeAccount(@NonNull final ICacheRecord guestRecord,
                                                       @NonNull final List<ICacheRecord> homeRecords) {
+            final String methodTag = CacheRecordFilter.class.getSimpleName() + ":hasNoCorrespondingHomeAccount";
```
shahzaibj: We usually define a class level TAG constant and then append that to method name to create a method tag (rather than doing it inline here)

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
```diff
@@ -98,8 +118,13 @@ public List<ICacheRecord> filter(@NonNull final List<ICacheRecord> records) {
...
             // Init our sought value
             final String guestAccountHomeAccountId = guestRecord.getAccount().getHomeAccountId();
+            if (guestAccountHomeAccountId == null){
```
shahzaibj: Do you also want to check for empty string here?

shahzaibj (changes requested):
Left comments

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
melissaahn: addressed... added TAG to each inner class of this AccountAdapter class

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
melissaahn: Addressed

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
melissaahn: addressed

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
```diff
@@ -51,16 +51,27 @@ private interface CacheRecordFilter {
...
             final List<ICacheRecord> result = new ArrayList<>();
 
             for (final ICacheRecord cacheRecord : records) {
                 final String acctHomeAccountId = cacheRecord.getAccount().getHomeAccountId();
                 final String acctLocalAccountId = cacheRecord.getAccount().getLocalAccountId();
-
-                if (!acctHomeAccountId.contains(acctLocalAccountId)) {
+                boolean notNullorEmpty = true;
+                if (StringUtil.isNullOrEmpty(acctHomeAccountId)) {
+                    Logger.warn(methodTag, "Home account id is null or empty.");
```
shahzaibj, line 67: In practice, how often does this happen? I'm just asking because these are `warn` level and I don't want to flood the logs with too much of these statements in cases when these being null is expected behavior

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
```diff
@@ -96,10 +118,17 @@ public List<ICacheRecord> filter(@NonNull final List<ICacheRecord> records) {
...
 
+        final String TAG = CacheRecordFilter.class.getSimpleName();
+
         private boolean hasNoCorrespondingHomeAccount(@NonNull final ICacheRecord guestRecord,
                                                       @NonNull final List<ICacheRecord> homeRecords) {
+            final String methodTag = TAG + ":hasNoCorrespondingHomeAccount";
             // Init our sought value
             final String guestAccountHomeAccountId = guestRecord.getAccount().getHomeAccountId();
+            if (StringUtil.isNullOrEmpty(guestAccountHomeAccountId)){
+                Logger.warn(methodTag, "Guest account home account id is null or empty.");
```
shahzaibj, line 129: In what cases do this happen?

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
somalaya: yes, how about boolean isNullOrEmpty = false?

shahzaibj (approved):

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
melissaahn: We don't know yet. Based on how the original code was written, it doesn't seem that either home id or local account id were expected to be null, but I also think that it's possible different scenarios have been created since then. The issue is still being actively investigated, so when I get more info, I'll make another PR to remove the logging if we need to consider null as expected.

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
melissaahn: Line 140 shows another case where a NPE can occur, so I'm putting a check here to be precautionary.

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
melissaahn: Addressed: f17222b57bd615f9654573a530aba9b5cd6aaadb

melissaahn:
@rpdome made a targeted fix for this issue here: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1933 . Going to close this PR.

PR #1930, opened 2023-10-24
October UI Maintenance

fadidurah:
October UI Maintenance:

- Take in changes from common
- Include a short wait in 833514 to increase consistency
- Retry on failrue for api 29 and below result in intent failure (Could not launch intent)
- Make 99267 more consistent

Common: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2217

### DIFF
--- a/msalautomationapp/build.gradle
+++ b/msalautomationapp/build.gradle
@@ -71,7 +71,7 @@ android {
 
     testOptions {
         execution 'ANDROIDX_TEST_ORCHESTRATOR'
-        animationsDisabled = true
+        animationsDisabled = false
     }
 
     buildTypes {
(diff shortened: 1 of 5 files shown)


melissaahn (approved):

somalaya:
What exactly is the issue with RetryOnFailure with API 29?

fadidurah:
> What exactly is the issue with RetryOnFailure with API 29?

For whatever reason, we can't succesfully retry on api 29, we get the intent failure error when we try. So having a retry actually just masks whatever the actual failure exception is, because we just see the intent failure

tanmaymanolkar1 (approved):

PR #1931, opened 2023-10-27
isQrPinAvailable API

p3dr0rv:
## Why?
To determine if QR + PIN authorization is available for MDM devices, Teams app will call this API.
if the result is true: Teams will show a button to login using QR + PIN, 
if the result is false: Teams will not show this button.

## How does this API works?
The QR + PIN feature will be configured by the administrator through app configuration policies for managed Android Enterprise devices. https://learn.microsoft.com/en-us/mem/intune/apps/app-configuration-policies-use-android
Where the administrator can enable/disable this type of authorization for all managed devices, by seating this key "isQrPinAvailable" to true in the Authenticator app. 

On Android we have 3 apps which can be the broker, Authenticator, CP (Company Portal) and LTW (Link to Windows). which makes things extremely complex.  
To keep the configuration simple for the tenant's administrator, we propose to use the managed configurations of the Authenticator app to read/set the admin flag. Meaning that authenticator app must be installed on Android devices to use the QR + PIN feature. 

qrpinapi drawio2

[Client design doc](https://microsoft-my.sharepoint-df.com/:w:/p/mipetriu/EWbcvu99HRxEgy-BC9jilKoBnqIr3yCJaNPZMlGSG5c3ow?e=aP0ey1)
## Changes

- In this PR we modify the msal test app to add button and check the functionality of the API
- We add the new function to the PublicClientAPI

## Related PR
https://github.com/AzureAD/ad-accounts-for-android/pull/2587
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2219


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1261,6 +1267,68 @@ public boolean isSharedDevice() {
...
+        final ResultFuture<CommandResult> future = CommandDispatcher.submitSilentReturningFuture(command);
+
+        try {
+            final CommandResult commandResult = future.get();
+            switch (commandResult.getStatus()) {
+                case COMPLETED:
+                    Logger.info(methodTag, "is QR + PIN available? " + commandResult.getResult());
+                    return (Boolean) commandResult.getResult();
+                case ERROR:
+                    Logger.error(methodTag, "Unexpected error on isQRPinAvailable", (Throwable) commandResult.getResult());
```
rpdome: This is throwing me off :/ 

We should be using something similar to kotlin's Result class instead. (separate method for result and exception).

rpdome (approved):

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
p3dr0rv: I'll address this as a separate PR, I would like to keep it simple, and this change will affect other thing no related to QR+PIN.

fadidurah (approved):

PR #1939, opened 2023-11-06
Rapong/release/4.9.1

rpdome:
Hotfix.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
@@ -82,7 +82,7 @@ class AccountAdapter {
             for (final ICacheRecord cacheRecord : records) {
                 final String acctHomeAccountId = cacheRecord.getAccount().getHomeAccountId();
                 final String acctLocalAccountId = cacheRecord.getAccount().getLocalAccountId();
-                if (acctLocalAccountId != null && acctHomeAccountId.contains(acctLocalAccountId)) {
+                if (acctHomeAccountId.contains(acctLocalAccountId)) {
                     result.add(cacheRecord);
                 }
             }
--- a/changelog
+++ b/changelog
@@ -1,5 +1,5 @@
 MSAL Wiki : https://github.com/AzureAD/microsoft-authentication-library-for-android/wiki
-vNext
+Version 4.9.1
 ----------
 - [PATCH] Fix NPE in SingleAccountPublicClientApplication.getPersistedCurrentAccount (#1933)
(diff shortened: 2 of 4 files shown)


msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
```diff
@@ -82,7 +82,7 @@ public List<ICacheRecord> filter(@NonNull final List<ICacheRecord> records) {
             for (final ICacheRecord cacheRecord : records) {
                 final String acctHomeAccountId = cacheRecord.getAccount().getHomeAccountId();
                 final String acctLocalAccountId = cacheRecord.getAccount().getLocalAccountId();
-                if (acctLocalAccountId != null && acctHomeAccountId.contains(acctLocalAccountId)) {
```
shahzaibj, line 85: Why removing this null check? Isn't this going to actually cause NPE now?

rpdome: @shahzaibj This was actually a band-aid fix of the same issue.

account local id is not supposed to be null. If we're getting null, it's likely that the data is malformed.

the fix I made targeted the malformed data in a specific flow. We probably shouldn't add a null check to the layer which is used throughout the code base.

see melissa's suggestion in this PR
https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1933

shahzaibj: I looked into the code, the `getLocalAccountId` method on AccountRecord is not marked NonNull - so that means it is Nullable and that means that caller should perform the null check before using the value

rpdome: @shahzaibj the class AccountRecord was written in 2018. IIRC, that was before we introduced Nullable/NonNull  (lombok) into the code.

Therefore, I'm not sure if we can conclude that every field in that class is meant to be nullable. 

My concern is that suppressing the missing value in that layer could lead to unexpected behavior in other flows. (This is why I switched to have a flow-specific exception handler in https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1933 instead)

The PR that introduced this null check was https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1847, which had the same stack trace as https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1933.  We never had/needed a null check here before then.

shahzaibj (approved):

melissaahn (approved):

PR #1949, opened 2023-11-10
Signup for NativeAuth

SaurabhMSFT:
This PR includes the work for Signup related features of Native Auth. No new dependencies have been added as part of this PR.

A technical overview of Native Auth is available at https://dev.azure.com/IdentityDivision/DevEx/_git/AuthLibrariesApiReview?path=/%5BAndroid%5D%20Native%20authentication/technical_overview.md&version=GBsammy/native-authentication&_a=preview

A companion PR in MSAL Common is https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2229


msal/src/main/java/com/microsoft/identity/client/RequiredUserAttribute.kt
```diff
@@ -0,0 +1,55 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.client
+
+import com.microsoft.identity.common.java.providers.nativeauth.responses.UserAttributeApiResult
+
+/**
+ * RequiredUserAttribute represents details about the account attributes required by the server.
```
shahzaibj, line 28: I'm not sure what we mean by these being required. There is a boolean flag here..so if require boolean was false then will these still be considered required?

msal/src/main/java/com/microsoft/identity/client/UserAttributes.kt
```diff
@@ -0,0 +1,132 @@
+package com.microsoft.identity.client
```
shahzaibj, line 23: LICENSE

msal/src/main/java/com/microsoft/identity/client/statemachine/states/SignUpStates.kt
```diff
@@ -0,0 +1,632 @@
...
+                    )
+                }
+            }
+        }
+    }
+}
+
+/**
+ * Native Auth uses a state machine to denote state and transitions for a user.
+ * SignInAfterSignUpState class represents a state where the user must signin after successful
```
shahzaibj, line 596: So we don't automatically sign the user in ourself as part of sign up?

SaurabhMSFT: From user's perspective we do. The signup operation will provide a short lived token to the sign in process and that will complete the sign in. It will be completely transparent to the user.

msal/src/main/java/com/microsoft/identity/client/UserAttributes.kt
SaurabhMSFT: Added

msal/src/main/java/com/microsoft/identity/client/RequiredUserAttribute.kt
SaurabhMSFT: These attributes are requested (required) by the server so they are called `RequiredUserAttribute`. The required field denotes whether it is mandatory for the attribute to be sent. The class names reflect the attribute names we get from the REST APIs.

msal/src/main/java/com/microsoft/identity/client/statemachine/states/SignUpStates.kt
shahzaibj: So the client app developer doesn't have to invoke the SignInAfterSignupState?

msal/src/main/java/com/microsoft/identity/client/RequiredUserAttribute.kt
shahzaibj: So the boolean here will always be true. Honestly the naming here is too confusing. It sounds like the required boolean means something else but I'm not sure what it means

msal/src/main/java/com/microsoft/identity/client/statemachine/states/SignUpStates.kt
SaurabhMSFT: Yes, the client doesn't have to invoke SignInAfterSignupState

msal/src/main/java/com/microsoft/identity/client/RequiredUserAttribute.kt
SaurabhMSFT: I agree the naming is a bit confusing but it represents the attribute names from the REST API. I will talk to API folks to see if we can rename it. That change will be outside of this PR.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
```diff
@@ -524,6 +536,402 @@ class NativeAuthPublicClientApplication(
...
+                        "Unexpected result $result"
+                    )
+                    SignUpResult.UnexpectedError(
+                        error = GeneralError(
+                            errorMessage = "Unexpected state",
+                            error = "unexpected_state",
+                            correlationId = result.correlationId
+                        )
+                    )
+                }
```
fadidurah, line 928: is there anything different in how signup and signupwithpassword handle the command results? Feel like most if not all of this can be pulled out into a shared helper method.

msal/src/main/java/com/microsoft/identity/client/statemachine/states/SignUpStates.kt
```diff
@@ -0,0 +1,632 @@
...
+class SignInAfterSignUpState internal constructor(
+    override val signInVerificationCode: String?,
+    override val username: String,
+    private val config: NativeAuthPublicClientApplicationConfiguration
+) : SignInAfterSignUpBaseState(signInVerificationCode, username, config) {
+    private val TAG: String = SignInAfterSignUpState::class.java.simpleName
+    interface SignInAfterSignUpCallback : SignInAfterSignUpBaseState.SignInAfterSignUpCallback
+
+    /**
+     * Signs in with the sign-in-after-sign-up verification code; callback variant.
```
fadidurah, line 611: is the sign-in-after-sign-up verification code different from the slt token you get after signing up?

msal/src/main/java/com/microsoft/identity/client/statemachine/states/SignUpStates.kt
fadidurah: I believe we also make sure documentation in customer-facing methods doesn't refer to the SLT token directly, Seems we refere to it as verification code here?

fadidurah: I think they would still have to invoke the sign in method from signInAfterSignupState, how else would they pass scopes when signing in.

msal/src/main/java/com/microsoft/identity/client/statemachine/states/SignUpStates.kt
SaurabhMSFT: Yes, it is the same. We are going to change name of all the tokens to `continuation_token` in a PR next week. Do you want me to change the variable names here?

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
SaurabhMSFT: I tried to create it but it cannot be done easily. The return types for these methods are different classes `SignUpResult` and `SignUpUsingPasswordResult`. The `SignUpUsingPasswordResult` does not derive from `SignUpResult` so I cannot find a common type that will be suitable for the helper method.

msal/src/main/java/com/microsoft/identity/client/statemachine/states/SignUpStates.kt
SaurabhMSFT: The token names are being changed to continuation_token. I was wrong in the previous comment. The developer has to call signin method after successful signup operation.

msal/src/main/java/com/microsoft/identity/client/NativeAuthPublicClientApplication.kt
SaurabhMSFT: I looked into today to see if we can change the hierarchies of `SignUpResult ` and `SignUpUsingPasswordResult` classes to have a common method. We cannot do it without making wholesale changes to class hierarchies.

msal/src/main/java/com/microsoft/identity/client/statemachine/states/SignUpStates.kt
fadidurah: Next PR is fine

fadidurah (approved):

msal/src/main/java/com/microsoft/identity/client/RequiredUserAttribute.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    //Attribute value should match the constraints
+    val options: RequiredUserAttributeOptions?
+)
+
+internal fun List<UserAttributeApiResult>.toListOfRequiredUserAttribute(): List<RequiredUserAttribute> {
+    return this.map { it.toRequiredUserAttribute() }
+}
+
+internal fun UserAttributeApiResult.toRequiredUserAttribute(): RequiredUserAttribute {
```
shahzaibj: Javadoc on these two methods

msal/src/main/java/com/microsoft/identity/client/UserAttributes.kt
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        fun build(): UserAttributes {
+            return UserAttributes(userAttributes)
+        }
+    }
+}
+
+internal fun UserAttributes.toMap(): Map<String, String> {
+    return ObjectMapper.constructMapFromObject(userAttributes)
```
shahzaibj, line 153: wasn't `userAttributes` already declared as a Map?

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -357,6 +364,220 @@ public static DeviceCodeFlowCommandParameters createDeviceCodeFlowCommandParamet
         return commandParameters;
     }
 
+    /**
+     * Creates command parameter for [SignUpStartCommand] of Native Auth.
```
shahzaibj, line 368: You can use `{@link SignUpStartCommand}` to actually link to that file from within the javadoc....here and everywhere

msal/src/main/java/com/microsoft/identity/client/statemachine/results/SignUpResult.kt
```diff
@@ -0,0 +1,233 @@
...
+import com.microsoft.identity.client.statemachine.InvalidEmailError
+import com.microsoft.identity.client.statemachine.InvalidPasswordError
+import com.microsoft.identity.client.statemachine.UserAlreadyExistsError
+import com.microsoft.identity.client.statemachine.states.SignInAfterSignUpState
+import com.microsoft.identity.client.statemachine.states.SignUpAttributesRequiredState
+import com.microsoft.identity.client.statemachine.states.SignUpCodeRequiredState
+import com.microsoft.identity.client.statemachine.states.SignUpPasswordRequiredState
+
+/**
+ * Sign up start result, produced by
```
shahzaibj: Sign up start result or just sign up result?

shahzaibj (approved):
Approved with comments

msal/src/main/java/com/microsoft/identity/client/RequiredUserAttribute.kt
SaurabhMSFT: Updated

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
SaurabhMSFT: Updated

msal/src/main/java/com/microsoft/identity/client/statemachine/results/SignUpResult.kt
SaurabhMSFT: Updated

PR #1964, opened 2023-11-30
Add PreferredAuthMethod to interactive token flow

p3dr0rv:
## Why
With the introduction of the QR + PIN Auth method a new query string parameter was added.
Preferred authorization method = pc.
This parameter will be passed to the authorization endpoint to indicate the preferred authorization method for a request. (for QR + PIN Auth the code is 18)
If this parameter is present ESTS will read the code and will return a UX according to the specified pc. 
If this parameter is not present the behavior will be the same, we have now.

## Changes

* Add PreferredAuthMethod to AcquireTokenParameters.
* Pass the data on the CommandParametersAdapter.
* Add button on MSALTestApp to test this flow.

## Related PR's
https://github.com/AzureAD/ad-accounts-for-android/pull/2630
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2245


fadidurah (approved):

msal/src/main/java/com/microsoft/identity/client/AcquireTokenParameters.java
```diff
@@ -86,6 +91,10 @@ void setLoginHint(String loginHint) {
         this.mLoginHint = loginHint;
     }
 
+    public PreferredAuthMethod getPreferredAuthMethod() {
```
rpdome, line 100: nit: add comments.

rpdome: and mark nullable on top of this as well.

msal/src/main/java/com/microsoft/identity/client/AcquireTokenParameters.java
```diff
@@ -157,6 +168,11 @@ public AcquireTokenParameters.Builder withLoginHint(String loginHint) {
             return self();
         }
 
+        public AcquireTokenParameters.Builder withPreferredAuthMethod(final PreferredAuthMethod preferredAuthMethod) {
```
rpdome, line 178: nit: nonnull? nullable?

rpdome (approved):

PR #1965, opened 2023-11-30
Fixing label in CommandParameters tests

melissaahn:
### Summary
Removing the Ignore modifier I accidently put on the CommandParametersTest class, and putting all the passkey specific classes behind a flag (makes it easier to find them later).

### DIFF
--- a/msal/src/test/java/com/microsoft/identity/client/CommandParametersTest.java
+++ b/msal/src/test/java/com/microsoft/identity/client/CommandParametersTest.java
@@ -44,6 +44,7 @@ import com.microsoft.identity.common.java.providers.oauth2.OAuth2TokenCache;
 import com.microsoft.identity.common.java.exception.ClientException;
 
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


msal/src/test/java/com/microsoft/identity/client/CommandParametersTest.java
```diff
@@ -181,94 +180,110 @@ public void testDeviceCodeFlowOperationWithoutClaims() throws ClientException {
 
     @Test
     public void testAppendToExtraQueryParametersIfWebAuthnCapable_UnsetPropertyAndNullInput() {
-        final List<Map.Entry<String, String>> combinedQueryParameters = CommandParametersAdapter.appendToExtraQueryParametersIfWebAuthnCapable(
-                null,
-                getConfiguration(AAD_NONE_CONFIG_FILE)
-        );
-        Assert.assertNull(combinedQueryParameters);
+        if (FidoConstants.IS_PASSKEY_SUPPORT_READY) {
```
shahzaibj: So these tests will only run if passkey support is ready? 

You may want to use JUnit Assume and assume the value of `IS_PASSKEY_SUPPORT_READY` - this way you can halt (ignore) the test if passkey support is not ready

https://junit.org/junit4/javadoc/4.12/org/junit/Assume.html#assumeTrue(boolean)

shahzaibj (approved):

rpdome (approved):

msal/src/test/java/com/microsoft/identity/client/CommandParametersTest.java
melissaahn: Thanks for the suggestion; changes here: 773e61e49efec38645c1d19513c3d78e6991c5b8

PR #1982, opened 2023-12-18
Implement updates of the Native Auth API

GBakolas:
Summary of changes:

- Convert all "flow" tokens to continuation token
- Use continuation token for sign in after sign up instead of SLT

Companion PR in common repo: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2261


msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInStates.kt
```diff
@@ -384,15 +385,15 @@ class SignInPasswordRequiredState(
...
  * successfull signup
  * in the signin flow.
- * @property signInVerificationCode: Short lived token from signup APIS
+ * @property continuationToken: Continuation token from signup APIS
  * @property username: Username of the user
  * @property config Configuration used by Native Auth
  */
 abstract class SignInAfterSignUpBaseState(
-    internal open val signInVerificationCode: String?,
+    override val continuationToken: String?,
```
SammyO, line 393: could you double check that having `internal` in the base class keeps it internal in extending classesI? E.g. in this case, `SignInAfterSignUpBaseState.continuationToken` should not be accessible by the sample app.

SammyO (approved):

SaurabhMSFT (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInStates.kt
GBakolas: Tried to access the continuationToken from the sample app and got a compilation error stating it is internal.

PR #1990, opened 2024-01-02
Remove dependency on testutils project

SaurabhMSFT:
Testutils project is not required by MSAL so removing it.


iamgusain:
@SaurabhMSFT I have already made this change with PR #1988 , you can close the PR.

PR #1993, opened 2024-01-05
Fixed a nonce testcase - TestCase1561652

somalaya:
Issue : The testcase was failing after PRTV3 because the response format seems to have changed. Order of the fields refreshToken and nonce is changed. Earlier we were seeing RT followed by nonce. Now we are seeing nonce first and then RT.
Fix : Corrected the testcase to run irrespective of the order.

### DIFF
--- a/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/brokerapi/TestCase1561652.java
+++ b/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/brokerapi/TestCase1561652.java
@@ -104,14 +104,7 @@ public class TestCase1561652 extends AbstractMsalBrokerTest {
                     .append('\n');
         }
         final String decodedToken = sb.toString();
-        if (decodedToken.contains("request_nonce")) {
-            final String[] str = decodedToken.split("request_nonce => ");
-            if (str.length > 1) {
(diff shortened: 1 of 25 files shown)


msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/brokerapi/TestCase1561652.java
```diff
@@ -104,14 +104,7 @@ public void decodeSSOTokenAndVerifyNonce(@NonNull final String ssoToken,
...
         final String decodedToken = sb.toString();
-        if (decodedToken.contains("request_nonce")) {
-            final String[] str = decodedToken.split("request_nonce => ");
-            if (str.length > 1) {
-                Assert.assertEquals(str[1].trim(), nonce);
-            } else {
-                Assert.fail("decoded token does not contain correct nonce");
-            }
-        } else {
+        if (!decodedToken.contains("request_nonce")) {
```
iamgusain: Should also verify that it contains the <nonce?>

somalaya: yes, now it is only looking for nonce

iamgusain (approved):

PR #1996, opened 2024-01-08
Use Parcelable instead of Serializable for NativeAuth statemachine classes

SaurabhMSFT:
The four state files (SignInStates, SignUpStates, ResetPasswordStates and AccountState have been refactored. They no longer inherit from Serializable but do so from the Parcelable interface. Parcelable interface is more Android specific and thats why its been used. One another benefit of Parcelable is that it doesn't support serialisation to a file. NativeAuthPublicClientApplicationConfiguration has not been modified as it's not exposed directly as part of State classes

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
+++ b/msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
@@ -101,6 +101,7 @@ class NativeAuthPublicClientApplication(
     init {
         initializeApplication()
         initializeSharedPreferenceFileManager(nativeAuthConfig.appContext)
+        applicationConfig = nativeAuthConfig
     }
 
     companion object {
@@ -116,6 +117,8 @@ class NativeAuthPublicClientApplication(
         //  To avoid duplicating the code, callback methods are routed through their
         //  coroutine-equivalent through this CoroutineScope.
         val pcaScope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
+
+        var applicationConfig:NativeAuthPublicClientApplicationConfiguration = NativeAuthPublicClientApplicationConfiguration()
 
         fun getCurrentAccountInternal(config: NativeAuthPublicClientApplicationConfiguration): IAccount? {
             LogSession.logMethodCall(TAG, "${TAG}.getCurrentAccountInternal")
--- a/msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
+++ b/msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
@@ -309,7 +309,7 @@ class NativeAuthPublicClientApplication(
                 is SignInCommandResult.CodeRequired -> {
                     SignInResult.CodeRequired(
                         nextState = SignInCodeRequiredState(
-                            flowToken = result.credentialToken,
+                            continuationToken = result.continuationToken,
                             scopes = scopes,
                             config = nativeAuthConfig
                         ),
@@ -321,7 +321,7 @@ class NativeAuthPublicClientApplication(
                 is SignInCommandResult.PasswordRequired -> {
                     SignInResult.PasswordRequired(
                         nextState = SignInPasswordRequiredState(
-                            flowToken = result.credentialToken,
+                            continuationToken = result.continuationToken,
                             scopes = scopes,
                             config = nativeAuthConfig
                         )
@@ -470,7 +470,7 @@ class NativeAuthPublicClientApplication(
                         )
                         SignInResult.CodeRequired(
                             nextState = SignInCodeRequiredState(
-                                flowToken = result.credentialToken,
+                                continuationToken = result.continuationToken,
                                 scopes = scopes,
                                 config = nativeAuthConfig
                             ),
@@ -610,7 +610,7 @@ class NativeAuthPublicClientApplication(
                     is SignUpCommandResult.Complete -> {
                         SignUpResult.Complete(
                             nextState = SignInAfterSignUpState(
-                                signInVerificationCode = result.signInSLT,
+                                continuationToken = result.continuationToken,
                                 username = username,
                                 config = nativeAuthConfig
                             )
@@ -620,7 +620,7 @@ class NativeAuthPublicClientApplication(
                     is SignUpCommandResult.AttributesRequired -> {
                         SignUpResult.AttributesRequired(
                             nextState = SignUpAttributesRequiredState(
-                                flowToken = result.signupToken,
+                                continuationToken = result.continuationToken,
                                 username = username,
                                 config = nativeAuthConfig
                             ),
@@ -631,7 +631,7 @@ class NativeAuthPublicClientApplication(
                     is SignUpCommandResult.CodeRequired -> {
                         SignUpResult.CodeRequired(
                             nextState = SignUpCodeRequiredState(
-                                flowToken = result.signupToken,
+                                continuationToken = result.continuationToken,
                                 username = username,
                                 config = nativeAuthConfig
                             ),
@@ -792,7 +792,7 @@ class NativeAuthPublicClientApplication(
                 is SignUpCommandResult.Complete -> {
                     SignUpResult.Complete(
                         nextState = SignInAfterSignUpState(
-                            signInVerificationCode = result.signInSLT,
+                            continuationToken = result.continuationToken,
                             username = username,
                             config = nativeAuthConfig
                         )
@@ -802,7 +802,7 @@ class NativeAuthPublicClientApplication(
                 is SignUpCommandResult.AttributesRequired -> {
                     SignUpResult.AttributesRequired(
                         nextState = SignUpAttributesRequiredState(
-                            flowToken = result.signupToken,
+                            continuationToken = result.continuationToken,
                             username = username,
                             config = nativeAuthConfig
                         ),
@@ -813,7 +813,7 @@ class NativeAuthPublicClientApplication(
                 is SignUpCommandResult.CodeRequired -> {
                     SignUpResult.CodeRequired(
                         nextState = SignUpCodeRequiredState(
-                            flowToken = result.signupToken,
+                            continuationToken = result.continuationToken,
                             username = username,
                             config = nativeAuthConfig
                         ),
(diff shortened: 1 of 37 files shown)


SammyO:
@SaurabhMSFT can you attach the corresponding sample app PR?

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -117,6 +118,8 @@ class NativeAuthPublicClientApplication(
         //  coroutine-equivalent through this CoroutineScope.
         val pcaScope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
 
+        var applicationConfig:NativeAuthPublicClientApplicationConfiguration = NativeAuthPublicClientApplicationConfiguration()
```
SammyO: This is basically a singleton, correct? We can't do this, as developers need to be able to have multiple SDK instances running in the same application (e.g. using different tenant applications and user flows).

SammyO: Note: I tried this approach a few months ago, but faced the above limitation. Which is why I refactored it into passing the configuration object to every state object initialisation.

SammyO: Why did you make this change in the first place? As I see you're still passing the configuration object to the state constructors.

SammyO:
@SaurabhMSFT can you expand on the need for this refactoring (either here or in the ticket)? As I'm seeing existing MSAL classes (e.g. `IAccount`) that implement `Serializable` currently.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
```diff
@@ -65,10 +66,15 @@ import java.io.Serializable
...
 class ResetPasswordCodeRequiredState internal constructor(
     override val flowToken: String,
-    private val config: NativeAuthPublicClientApplicationConfiguration
-) : BaseState(flowToken), State, Serializable {
+    private val config: NativeAuthPublicClientApplicationConfiguration = NativeAuthPublicClientApplication.applicationConfig
+) : BaseState(flowToken), State, Parcelable {
     private val TAG: String = ResetPasswordCodeRequiredState::class.java.simpleName
 
+    constructor(parcel: Parcel) : this(
+        parcel.readString()!!
```
SammyO: Can we make this (and in other instances) a bit more elegant? The application shouldn't crash if the String returned from `readString()` is null.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignUpStates.kt
```diff
@@ -288,6 +297,25 @@ class SignUpCodeRequiredState internal constructor(
...
+    override fun describeContents(): Int {
+        return 0
+    }
+
+    companion object CREATOR : Parcelable.Creator<SignUpCodeRequiredState> {
+        override fun createFromParcel(parcel: Parcel): SignUpCodeRequiredState {
+            return SignUpCodeRequiredState(parcel)
+        }
+
+        override fun newArray(size: Int): Array<SignUpCodeRequiredState?> {
```
SammyO, line 316: Why is this nullable?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
```diff
@@ -256,9 +280,15 @@ class ResetPasswordCodeRequiredState internal constructor(
 class ResetPasswordPasswordRequiredState internal constructor(
     override val flowToken: String,
     private val config: NativeAuthPublicClientApplicationConfiguration
-) : BaseState(flowToken), State, Serializable {
+) : BaseState(flowToken), State, Parcelable {
     private val TAG: String = ResetPasswordPasswordRequiredState::class.java.simpleName
 
+    constructor(parcel: Parcel) : this(
+        parcel.readString()!!,
+        parcel.readSerializable() as NativeAuthPublicClientApplicationConfiguration
```
SammyO, line 291: This is inconsistent with other instances of this constructor. E.g. in the Parcel constructor in `ResetPasswordCodeRequiredState` you're not reading (and setting) the configuration object.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
```diff
@@ -361,4 +391,22 @@ class ResetPasswordPasswordRequiredState internal constructor(
             }
         }
     }
+
+    override fun writeToParcel(parcel: Parcel, flags: Int) {
+        parcel.writeString(flowToken)
```
SammyO: You're setting the config object in the constructor in line 288, but not writing it to the Parcel here.

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
SaurabhMSFT: I have updated the description with this info.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignUpStates.kt
SaurabhMSFT: This is implementing the CREATOR interface and the signature comes from that interface.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
SaurabhMSFT: We should not write and read config objects as it could lead to stale configuration.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
SaurabhMSFT: yes, config is not being written on purpose.

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
SaurabhMSFT: If multiple SDK instances are allowed then we need to change the signature of the APIs.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
Yuki-YuXin: Could we write like this if NativeAuthPublicClientApplicationConfiguration is not a singleton pattern?

```
constructor(parcel: Parcel) : this(
        parcel.readString() ?: "",
        NativeAuthPublicClientApplicationConfiguration()
}
```

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
SaurabhMSFT: This change has been removed.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
SaurabhMSFT: Done

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
SaurabhMSFT: Done

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
SaurabhMSFT: Updated to write config to parcel.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -55,43 +57,23 @@ import com.microsoft.identity.nativeauth.statemachine.results.SignOutResult
...
-            config: NativeAuthPublicClientApplicationConfiguration
-        ): AccountState {
-            return AccountState(
-                account = account,
-                config = config
-            )
-        }
+    constructor(parcel: Parcel) : this(
+        parcel.readSerializable() as IAccount,
+        parcel.readSerializable() as NativeAuthPublicClientApplicationConfiguration
```
Yuki-YuXin, line 73: NativeAuthPublicClientApplicationConfiguration class does not implement Serializable. Can we use parcel.readSerializable() to read it from a Parcel?

SaurabhMSFT:
> @SaurabhMSFT can you expand on the need for this refactoring (either here or in the ticket)? As I'm seeing existing MSAL classes (e.g. `IAccount`) that implement `Serializable` currently.

https://github.com/Azure-Samples/ms-identity-ciam-native-auth-android-sample/pull/7

SaurabhMSFT:
> @SaurabhMSFT can you attach the corresponding sample app PR?

https://github.com/Azure-Samples/ms-identity-ciam-native-auth-android-sample/pull/7

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignUpStates.kt
```diff
@@ -586,8 +667,16 @@ class SignUpAttributesRequiredState internal constructor(
 class SignInAfterSignUpState internal constructor(
     override val continuationToken: String?,
     override val username: String,
-    private val config: NativeAuthPublicClientApplicationConfiguration
+    val config: NativeAuthPublicClientApplicationConfiguration
 ) : SignInAfterSignUpBaseState(continuationToken, username, config) {
+
+    constructor(parcel: Parcel) : this(
+        parcel.readString(),
```
SammyO, line 674: This and line 675 should be swapped, right? The first parcel.read entry should handle the first constructor parameter, which is nullable. The second handles a non-null parameter.

SaurabhMSFT: No, the current code is correct. The continuation token parameter can be nullable and the method `parcel.readString()` can return a nullable value.

SammyO: Oh right, I see 👍

SammyO (approved):

Yuki-YuXin (approved):

PR #2004, opened 2024-01-15
Update Native Auth password reset response to include sign in state

rmccahill:
Summary of changes:

* Updates the reset password completion state to return a sign in state with the required continuation token.
* Adds scenario tests for signing in after completing the reset password flow

Related PR in common repo: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2281
PR for the sample application: https://github.com/Azure-Samples/ms-identity-ciam-native-auth-android-sample/pull/9

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
+++ b/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
@@ -65,6 +65,7 @@ import java.io.Serializable
  */
 class ResetPasswordCodeRequiredState internal constructor(
     override val continuationToken: String,
+    private val username: String,
     private val config: NativeAuthPublicClientApplicationConfiguration
 ) : BaseState(continuationToken), State, Serializable {
     private val TAG: String = ResetPasswordCodeRequiredState::class.java.simpleName
@@ -121,6 +122,7 @@ class ResetPasswordCodeRequiredState internal constructor(
                     ResetPasswordSubmitCodeResult.PasswordRequired(
                         nextState = ResetPasswordPasswordRequiredState(
                             continuationToken = result.continuationToken,
+                            username = username,
                             config = config
                         )
                     )
@@ -213,6 +215,7 @@ class ResetPasswordCodeRequiredState internal constructor(
                     ResetPasswordResendCodeResult.Success(
                         nextState = ResetPasswordCodeRequiredState(
                             continuationToken = result.continuationToken,
+                            username = username,
                             config = config
                         ),
                         codeLength = result.codeLength,
@@ -256,6 +259,7 @@ class ResetPasswordCodeRequiredState internal constructor(
  */
 class ResetPasswordPasswordRequiredState internal constructor(
     override val continuationToken: String,
+    private val username: String,
     private val config: NativeAuthPublicClientApplicationConfiguration
 ) : BaseState(continuationToken), State, Serializable {
     private val TAG: String = ResetPasswordPasswordRequiredState::class.java.simpleName
@@ -311,7 +315,13 @@ class ResetPasswordPasswordRequiredState internal constructor(
                 return@withContext when (val result =
                     rawCommandResult.checkAndWrapCommandResultType<ResetPasswordSubmitNewPasswordCommandResult>()) {
                     is ResetPasswordCommandResult.Complete -> {
-                        ResetPasswordResult.Complete
+                        ResetPasswordResult.Complete(
+                            nextState = SignInAfterSignUpState(
+                                continuationToken = result.continuationToken,
+                                username = username,
+                                config = config
+                            )
+                        )
                     }
 
                     is ResetPasswordCommandResult.PasswordNotAccepted -> {
--- a/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
+++ b/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
@@ -317,7 +317,7 @@ class ResetPasswordPasswordRequiredState internal constructor(
                     rawCommandResult.checkAndWrapCommandResultType<ResetPasswordSubmitNewPasswordCommandResult>()) {
                     is ResetPasswordCommandResult.Complete -> {
                         ResetPasswordResult.Complete(
-                            nextState = SignInAfterPasswordResetState(
+                            nextState = SignInContinuationState(
                                 continuationToken = result.continuationToken,
                                 username = username,
                                 config = config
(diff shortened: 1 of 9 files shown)


SammyO:
@rmccahill could you attach the sample app PR as well?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
```diff
@@ -311,7 +315,13 @@ class ResetPasswordPasswordRequiredState internal constructor(
                 return@withContext when (val result =
                     rawCommandResult.checkAndWrapCommandResultType<ResetPasswordSubmitNewPasswordCommandResult>()) {
                     is ResetPasswordCommandResult.Complete -> {
-                        ResetPasswordResult.Complete
+                        ResetPasswordResult.Complete(
+                            nextState = SignInAfterSignUpState(
```
SammyO: If we're going to reuse `SignInAfterSignUpState` state class (which I'd be in favour of), we need to rename it. The current name doesn't fit the SSPR flow.

SammyO (changes requested):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
rmccahill: Good point - any naming suggestions? The most appropriate I can think of is `SignInAfterFlowCompletionState`.

SammyO: There was an idea a few months ago of not introducing a new state for this at all, but using the existing SignIn state (and adding new parameters, functions, etc.). Could that work?

rmccahill: Not sure what that would look like - from what I know, we initiate the sign in flow by calling signIn from the public client application itself, then the sign in states we have inherit from BaseState and reflect the potential results from the initial sign in call, eg: `SignInCodeRequiredState`, `SignInPasswordRequiredState`, etc. So I don't see how we could extend or re-use those states for this case.

For now, it looks like there was an abstract class called `SIgnInAfterSignUpBaseState` that contains the logic for running the sign in call, so I've re-named that to `SignInAfterFlowCompletionState`, and created a `SignInAfterResetPasswordState` that extends that abstract class. Let me know what you think.

SammyO: What does the current setup (base class with 2 classes extending it) have that we couldn't do with just 1 class; the base class? I.e. why do we need `SignInAfterResetPasswordState` and `SignInAfterSignUpState`?

rmccahill: Naming, mainly. Returning their respective names from the completion result (in my opinion) is more clear compared to the base name of `SignInAfterFlowCompletionState`.

SammyO: Can we rename the base class to make it useful for every flow? I'm questioning whether:
``` 
val result: ResetPassword.Complete
val nextState: SignInAfterResetPasswordState = result.nextState
nextState.signIn()
```

is much clearer then
```
val result: ResetPassword.Complete
val nextState: SignInAfterFlowCompletionState = result.nextState
nextState.signIn()
```

Developer should be checking the result class as against a list of expected results. The name of the `nextState` class is much less relevant (in many cases developers might not even see it). 
I do agree that `SignInAfterFlowCompletionState` is not a great name. Can we come up with something more developer friendly, something simpler? `QuickSignInState`?

rmccahill: After discussing the naming, the new name will be `SignInContinuationState`. The `SignInAfterFlowCompletionState` is no longer abstract and is now used for both the Sign Up and Reset Password flows. `SignInAfterSignUpState` and `SignInAfterPasswordState` are both removed.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInStates.kt
```diff
@@ -382,55 +382,54 @@ class SignInPasswordRequiredState(
...
 
     /**
-     * Submits the sign-in-after-sign-up verification code to the server; Kotlin coroutines variant.
+     * Submits the sign-in-continuation verification code to the server; Kotlin coroutines variant.
      *
      * @param scopes (Optional) The scopes to request.
      * @return The results of the sign-in-after-sign-up action.
      */
-    suspend fun signInAfterSignUp(scopes: List<String>? = null): SignInResult {
+    suspend fun signIn(scopes: List<String>? = null): SignInResult {
```
Yuki-YuXin, line 465: For those SignInCommandResult.CodeRequired & SignInCommandResult.PasswordRequired & INativeAuthCommandResult.Redirect, should we group them togerther into the error? @GBakolas is proposing a SignInCompletionError class with no utility methods. I'm guessing this is another PR after this one.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInStates.kt
```diff
@@ -382,55 +382,54 @@ class SignInPasswordRequiredState(
...
      *
      * @param scopes (Optional) The scopes to request.
-     * @param callback [com.microsoft.identity.nativeauth.statemachine.states.SignInAfterSignUpBaseState.SignInAfterSignUpCallback] to receive the result on.
-     * @return The results of the sign-in-after-sign-up action.
+     * @param callback [com.microsoft.identity.nativeauth.statemachine.states.SignInContinuationState.SignInContinuationCallback] to receive the result on.
+     * @return The results of the sign-in-continuation action.
      */
-    fun signInAfterSignUp(scopes: List<String>? = null, callback: SignInAfterSignUpCallback) {
-        LogSession.logMethodCall(TAG, "${TAG}.signInAfterSignUp")
+    fun signIn(scopes: List<String>? = null, callback: SignInContinuationCallback) {
```
Yuki-YuXin, line 445: Do we still need scopes in the case of a continuation token signIn?

rmccahill: I think so - I can see scenarios where developers would want to request specific scopes when signing in the user after signing up. Otherwise they would have to start the sign in flow from scratch to request specific scopes.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInStates.kt
rmccahill: Not sure what you mean by "group them together into the error" - which error are you referring to? Each error type will return a different state, so I don't see how we can combine them.

Yuki-YuXin: When signing in with a continuation token, I don't think the user is required to provide a password / code. My initial thought was if we could do something here instead of having the sample app treat them as invalid states. But you're right, the results come with different state/error parameters and we can't combine them.

Yuki-YuXin (approved):

SammyO (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
```diff
@@ -406,4 +408,23 @@ class ResetPasswordPasswordRequiredState internal constructor(
             }
         }
     }
-}
\ No newline at end of file
+
+    override fun writeToParcel(parcel: Parcel, flags: Int) {
+        parcel.writeString(continuationToken)
+        parcel.writeSerializable(config)
```
SaurabhMSFT, line 416: We should write username here

SaurabhMSFT (changes requested):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
```diff
@@ -66,11 +67,13 @@ import kotlinx.coroutines.withContext
  */
 class ResetPasswordCodeRequiredState internal constructor(
     override val continuationToken: String,
+    private val username: String,
     private val config: NativeAuthPublicClientApplicationConfiguration
 ) : BaseState(continuationToken), State, Parcelable {
     private val TAG: String = ResetPasswordCodeRequiredState::class.java.simpleName
 
     constructor(parcel: Parcel) : this(
+        parcel.readString() ?: "",
```
SaurabhMSFT, line 76: The `writeToParcel` method for this class should be updated.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
```diff
@@ -282,11 +287,13 @@ class ResetPasswordCodeRequiredState internal constructor(
  */
 class ResetPasswordPasswordRequiredState internal constructor(
     override val continuationToken: String,
+    private val username: String,
```
SaurabhMSFT, line 291: The writeToParcel method for this class should be updated.

SaurabhMSFT (changes requested):

SaurabhMSFT (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
rmccahill: Good catch - updated

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
rmccahill: Done

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
rmccahill: Done

PR #2007, opened 2024-01-17
Consistent correlation ID

SammyO:
Make correlation ID consistent throughout a flow, meaning all API request in a scenario should use the same correlation ID
- Add it to logging 
- Add it to CommandParameters
- Add it to states: any SDK result that contains a reference to a new state, will pass the correlation ID to that state instance (which is then used in the subsequent API request).

Furthermore this PR contains some improvements on the Parcelable implementation:
- Remove empty constructor body
- Used named parameters (to increase readability) 
- Create extension function to deal with `Parcel.readSerializable()` being deprecated on higher API versions.

Note:
This PR doesn't contain tests that validate the correlation ID using the mock API. When we introduce automated tests on production environment, tests for correlation ID should be added.

PR for common: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2283


msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -508,21 +523,25 @@ public static SignUpResendCodeCommandParameters createSignUpResendCodeCommandPar
...
                 .continuationToken(continuationToken)
+                .correlationId(correlationId)
                 .build();
     }
 
     /**
      * Creates command parameter for [{@link com.microsoft.identity.common.nativeauth.internal.commands.SignUpSubmitUserAttributesCommand}] of Native Auth.
      * @param configuration PCA configuration
      * @param tokenCache token cache for storing results
+     * @param correlationId correlation ID to use in the API request, taken from the previous API response in the flow
```
SaurabhMSFT: duplicate comment

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInStates.kt
```diff
@@ -443,7 +511,11 @@ class SignInContinuationState(
      * @return The results of the sign-in-continuation action.
      */
     fun signIn(scopes: List<String>? = null, callback: SignInContinuationCallback) {
-        LogSession.logMethodCall(TAG, "${TAG}.signIn")
+        LogSession.logMethodCall(
+            tag = TAG,
+            correlationId = correlationId,
+            methodName = "${TAG}.signInAfterSignUp"
```
Yuki-YuXin: Maybe ""${TAG}.signIn"?

msal/src/main/java/com/microsoft/identity/nativeauth/utils/ParcelExtensions.kt
```diff
@@ -0,0 +1,13 @@
+package com.microsoft.identity.nativeauth.utils
+
+import android.os.Build
+import android.os.Parcel
+import java.io.Serializable
+
+/**
+ * Helper method to deal with Parcel.readSerializable being deprecated in newer Android versions.
```
Yuki-YuXin, line 30: I'm personally curious about this file. Is it used to suppress warnings? Feel free to resolve the comment.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
```diff
@@ -406,11 +409,14 @@ public void onError(@NonNull BaseException exception) {
         application.signIn(username, null, signInCallback);
         // 1a. Server returns invalid user error
         assertTrue(signInResult.get(30, TimeUnit.SECONDS) instanceof SignInResult.CodeRequired);
-        SignInCodeRequiredState nextState = (((SignInResult.CodeRequired) signInResult.get(30, TimeUnit.SECONDS)).getNextState());
+        SignInCodeRequiredState nextState = spy((((SignInResult.CodeRequired) signInResult.get(30, TimeUnit.SECONDS)).getNextState()));
```
Yuki-YuXin, line 409: Why we would use spy() here? Any particular advantage? It's a confusion not a suggestion, feel free to revolt the comment.

msal/src/main/java/com/microsoft/identity/nativeauth/utils/ParcelExtensions.kt
SammyO: No, it's to accurately deal with what the warning warns about. Warnings are there to indicate that something shouldn't be used anymore, so this method tackles that. Ignoring warnings is another way to get rid of them, but not the right way (in this case).

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
SammyO: Good question. I'd recommend reading a bit about Mockito's `mock` vs. `spy`. In short: with spy you can wrap a real object into a partially mocked one. Any method/field that's not overwritten by the spy, will behave as normal. But the methods that are overwritten by the spy, will work as a mocked variant would. In this case it's helpful because, when using mock API, `correlationID` is going to be null. Using a spy we can change this field into some mocked value, so the flow continues to work as expected.

Yuki-YuXin (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInStates.kt
```diff
@@ -214,11 +232,16 @@ class SignInCodeRequiredState internal constructor(
      * @return The results of the resend code action.
      */
     suspend fun resendCode(): SignInResendCodeResult {
-        LogSession.logMethodCall(TAG, "${TAG}.resendCode()")
+        LogSession.logMethodCall(
+            tag = TAG,
+            correlationId = correlationId,
+            methodName = "${TAG}.resendCode()"
```
GBakolas, line 239: You have added () in methodName in this log but not in all the other logs

GBakolas: Oh now I see you are including the args in the log for the suspend methods. Ignore this comment

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
```diff
@@ -329,13 +360,18 @@ class ResetPasswordPasswordRequiredState internal constructor(
      * @return The results of the submit password action.
      */
     suspend fun submitPassword(password: CharArray): ResetPasswordSubmitPasswordResult {
-        LogSession.logMethodCall(TAG, "${TAG}.submitPassword(password: String)")
+        LogSession.logMethodCall(
+            tag = TAG,
+            correlationId = correlationId,
+            methodName = "${TAG}.submitPassword(password: String)"
```
GBakolas: CharArray instead of String?

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
SammyO: Addressed

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInStates.kt
SammyO: No I think it's a good point. let me clean this up a bit.

msal/src/main/java/com/microsoft/identity/nativeauth/utils/ParcelExtensions.kt
```diff
@@ -0,0 +1,13 @@
...
+ */
+inline fun <reified T : Serializable> Parcel.serializable(): T? = when {
+    Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU -> readSerializable(T::class.java.classLoader, T::class.java)
+    else -> @Suppress("DEPRECATION") readSerializable() as? T
+}
```
SaurabhMSFT: New line is missing at the end of the file.

msal/src/main/java/com/microsoft/identity/nativeauth/utils/ParcelExtensions.kt
```diff
@@ -0,0 +1,13 @@
+package com.microsoft.identity.nativeauth.utils
```
SaurabhMSFT, line 23: Missing license

msal/src/test/java/com/microsoft/identity/nativeauth/utils/MockCorrelationIdHelper.kt
```diff
@@ -0,0 +1,16 @@
+package com.microsoft.identity.nativeauth.utils
```
SaurabhMSFT, line 23: Missing license

SaurabhMSFT (changes requested):

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -586,6 +605,9 @@ public static SignInStartCommandParameters createSignInStartCommandParameters(
                 .clientId(configuration.getClientId())
                 .challengeType(configuration.getChallengeTypes())
                 .scopes(scopes)
+                // Start of the flow, so there is no correlation ID to use from a previous API response.
+                // Set it to a default value.
+                .correlationId(DiagnosticContext.INSTANCE.getThreadCorrelationId())
```
SaurabhMSFT, line 612: The value returned from the method call `DiagnosticContext.INSTANCE.getThreadCorrelationId()` is always going to be the constant value `DiagnosticContext.UNSET`. We should use that value instead of making the function call.

SammyO: > The value returned from the method call DiagnosticContext.INSTANCE.getThreadCorrelationId() is always going to be the constant value DiagnosticContext.UNSET

What makes you say that?

Please have a look at usages of `setRequestContext()`. Furthermore, manual E2E tests have shown that this field is correctly set, rather than it being value `UNSET`.

SaurabhMSFT: Thanks, I can see it getting updated via the `setRequestContext` method. The usage of `getThreadCorrelationId()` here is singleton pattern and it will cause issues if the developer creates multiple `NativeAuthPublicClientApplication` objects.

SammyO: That's a good point. Could you create a ticket for this on our engineering backlog? I'm not sure what the resolution should be, as this is how MSAL does it as well. But we can discuss that separately. If you agree, I'd like to leave this out of scope of this PR however.

rmccahill (approved):

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -444,12 +455,14 @@ public static SignUpSubmitCodeCommandParameters createSignUpSubmitCodeCommandPar
...
      * @param tokenCache token cache for storing results
      * @param continuationToken Continuation token received from the start command
+     * @param correlationId correlation ID to use in the API request, taken from the previous API response in the flow
      * @return Command parameter object
      * @throws ClientException
      */
     public static SignUpResendCodeCommandParameters createSignUpResendCodeCommandParameters(
             @NonNull final NativeAuthPublicClientApplicationConfiguration configuration,
             @NonNull final OAuth2TokenCache tokenCache,
+            @NonNull final String correlationId,
```
SaurabhMSFT: `correlationid` should come after `continuatioToken` as in other methods in this file

fadidurah (approved):

SaurabhMSFT (approved):

PR #2008, opened 2024-01-17
Move all parts of PCA creation to the same thread

fadidurah:
Relevant Bugs: 
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2502611
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2502610

**What**
We received some bugs from Teams regarding users experiencing Android Not Responding crashes while using teams (Bugs linked above). During PCA creation, the foreground thread would get blocked while waiting on a lock to open in `AzureActiveDirectory`. This shouldn't be happening, as PCA creation should be happening in the background thread, or specifically marked with the @WorkerThread annotation.

The culprit is the final stage of the PCA creation process where a new PCA instance is actually created, which calls `initializeApplication()`, a method that involves some synchronized method calls. This instance creation happens in a command callback for getDeviceMode, which is always ran on the main foreground thread today (see `AndroidPlatformUtil.postCommandResult()`). This is not ideal for PCA creation, because it means some part of it is happening in the main thread and could cause ANRs despite the user calling it from a background context. To fix this, This PR moves the PCA instance creation to the same thread that initially calls getDeviceMode.

**How**
I've extracted the instance creation outside of the command callback for getDeviceMode. Instead, we now use a future to get the command result, waiting for the command to finish so that we can instantiate the PCA on the background thread, rather than the foreground.

**Why**
This fix will remove this possibility of causing an ANR while creating the PCA and make sure that the PCA creation happens on the same thread.

**Testing**
Before the fix, debugging through PCA creation in msaltestapp would show `setEnvironment()` being ran on the main thread.
After the fix, this part of PCA creation runs on the same thread as the rest of the method (thread 2, in my case).

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
+++ b/msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
@@ -1102,21 +1102,26 @@ public class PublicClientApplication implements IPublicClientApplication, IToken
 
                     @Override
                     public void onTaskCompleted(Boolean isSharedDevice) {
-                        config.setIsSharedDevice(isSharedDevice);
+                        runOnBackground(new Runnable() {
+                            @Override
+                            public void run() {
+                                config.setIsSharedDevice(isSharedDevice);
 
-                        try {
-                            if (config instanceof NativeAuthPublicClientApplicationConfiguration
-                                    && config.getAccountMode() == AccountMode.SINGLE) {
-                                config.validateConfiguration();
-                                listener.onCreated(new NativeAuthPublicClientApplication((NativeAuthPublicClientApplicationConfiguration) config));
-                            } else if (config.getAccountMode() == AccountMode.SINGLE || isSharedDevice) {
-                                listener.onCreated(new SingleAccountPublicClientApplication(config));
-                            } else {
-                                listener.onCreated(new MultipleAccountPublicClientApplication(config));
+                                try {
+                                    if (config instanceof NativeAuthPublicClientApplicationConfiguration
+                                            && config.getAccountMode() == AccountMode.SINGLE) {
+                                        config.validateConfiguration();
+                                        listener.onCreated(new NativeAuthPublicClientApplication((NativeAuthPublicClientApplicationConfiguration) config));
+                                    } else if (config.getAccountMode() == AccountMode.SINGLE || isSharedDevice) {
+                                        listener.onCreated(new SingleAccountPublicClientApplication(config));
+                                    } else {
+                                        listener.onCreated(new MultipleAccountPublicClientApplication(config));
+                                    }
+                                } catch (final MsalClientException e) {
+                                    listener.onError(e);
+                                }
                             }
-                        } catch (final MsalClientException e) {
-                            listener.onError(e);
-                        }
+                        });
                     }
 
                     @Override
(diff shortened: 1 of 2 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -1276,8 +1276,16 @@ private void initializeApplication() throws MsalClientException {
         final Context context = mPublicClientConfiguration.getAppContext();
         setupTelemetry(context, mPublicClientConfiguration);
 
-        AzureActiveDirectory.setEnvironment(mPublicClientConfiguration.getEnvironment());
-        Authority.addKnownAuthorities(mPublicClientConfiguration.getAuthorities());
+        // Today, this method runs on the main thread. AzureActiveDirectory.setEnvironment() and Authority.addKnownAuthorities()
```
rpdome: These 2 values should actually be made as part of parameter list, not global variables.

Putting them in the background like this could also result in a race condition. (the "read" part happens before this "write" part is done)

melissaahn (approved):

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
shahzaibj: +1 on the race condition.

That said moving them to parameter list is likely a big change.

shahzaibj: Could we perhaps use a future here to block on the completion of the thread. Do we know if that would also cause ANRs?

fadidurah: > +1 on the race condition.
> 
> That said moving them to parameter list is likely a big change.

Yeah, I spent some time yesterday trying to move it away from being a global variable, it's a very big change. Doing that will require a concentrated refactoring effort, especially if we are also going to do it for knownAuthorities while we are at it.

fadidurah: > Could we perhaps use a future here to block on the completion of the thread. Do we know if that would also cause ANRs?

Anything that block the main thread is capable of casuing an ANR. The idea of this PR is to remove all locks from the main thread during the PCA init to remove the possibility of an ANR, but it introduces a possibility of a race condition

fadidurah: I'm currently working on another approach with the help Dome, where we separate the locks used in AzureActiveDirectory, as the main cause of the ANR in this case is the lock used to fetch/set environment, is the exact same lock that is used for the performCloudDiscovery network call. If we can remove that pairing, the ANR should become impossible

fadidurah: For the sake of the ANR, we can also consider just moving the callback handling to the background, so that all of the PCA intialization happens on the same thread, but nothing is locked on main. We aren't directly returning anything here just communicating to the listener. I don't think this will create any race conditions. We aren't directly returning anything here just communicating to the listener. thoughts @rpdome @shahzaibj

shahzaibj (approved):

fadidurah:
I've updated the fix to using the command future and continuing the PCA creation on the same thread.

fadidurah:
Closed as we are no longer pursuing this solution, as it may affect other customers and validate a possible unsupported use case of MSAL PCA Objects

PR #2009, opened 2024-01-18
Remove unnecessary isBrowserRequired in errors

GBakolas:
The existing native auth logic returns BrowserRequired error in more cases, compared to where the API can actually return it.
This PR removes the unnecessary isBrowserRequired from errors for certain scenarios (sign in after sign up, resend code)

Partner PR in common repo: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2298


msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
```diff
@@ -105,7 +124,7 @@ class SubmitCodeError(
     override val errorCodes: List<Int>? = null,
     val subError: String? = null,
     override var exception: Exception? = null
-): SignInSubmitCodeResult, SignUpSubmitCodeResult, ResetPasswordSubmitCodeResult, Error(errorType = errorType, error = error, errorMessage= errorMessage, correlationId = correlationId, errorCodes = errorCodes, exception = exception)
+): SignInSubmitCodeResult, SignUpSubmitCodeResult, ResetPasswordSubmitCodeResult, BrowserRequiredError(errorType = errorType, error = error, errorMessage= errorMessage, correlationId = correlationId, errorCodes = errorCodes, exception = exception)
```
Yuki-YuXin: Why SubmitCodeError use BrowserRequiredError here instead of Error but ResendCodeError use Error?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInStates.kt
```diff
@@ -244,26 +245,17 @@ class SignInCodeRequiredState internal constructor(
...
-                    ResendCodeError(
-                        errorType = ErrorTypes.BROWSER_REQUIRED,
-                        error = result.error,
-                        errorMessage = result.errorDescription,
-                        correlationId = result.correlationId
-                    )
-                }
-
-                is INativeAuthCommandResult.UnknownError -> {
+                is INativeAuthCommandResult.Redirect, is INativeAuthCommandResult.UnknownError -> {
```
Yuki-YuXin, line 248: For the sake of clarity, I would prefer to have line breaks between these results.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
Yuki-YuXin: For those valid INativeAuthCommandResult.Redirect Error, it should extend BrowserRequiredError

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInStates.kt
```diff
@@ -53,6 +53,7 @@ import com.microsoft.identity.common.java.util.StringUtil
 import com.microsoft.identity.common.java.nativeauth.util.checkAndWrapCommandResultType
 import com.microsoft.identity.nativeauth.statemachine.errors.ErrorTypes
 import com.microsoft.identity.nativeauth.statemachine.errors.ResendCodeError
+import com.microsoft.identity.nativeauth.statemachine.errors.SignInContinuationError
```
Yuki-YuXin, line 56: SignInPasswordRequiredState loses Parcelable implementation.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
```diff
@@ -233,25 +233,17 @@ class ResetPasswordCodeRequiredState internal constructor(
                     )
```
Yuki-YuXin, line 233: I am not sure about the reason but should  INativeAuthCommandResult.Redirect be part of the possible results of ResetPasswordPasswordRequiredState.submitPassword

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInStates.kt
SammyO: I don't agree with that. It's not a Kotlin code style I've seen before.

GBakolas: In almost all other places in the code base we also put multiple is statements on the same line

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInStates.kt
GBakolas: @SaurabhMSFT, what is the reason this class does not implement Parcelable?

SammyO: I did this in my PR: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2007/files

Yuki-YuXin (commented):
I personally think SignInContinuationState.signIn()'s SignInCommandResult.CodeRequired & SignInCommandResult.PasswordRequired results should return 
```
SignInContinuationError(
                    errorMessage = "Sign In is not available through this state, please use the standalone sign in methods (signInWithCode or signInWithPassword).",
                    error = "invalid_state",
                    correlationId = "UNSET",
                )
```
instead of those specific states

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
SammyO: I don't think this is the right way to achieve this. You're effectively saying that BrowserRequired is a base class, which all the sub-errors extend from. Which coincidentally achieves what you want, but is technically not correct. This parent-child hierarchy is not the right way to express what kind of results an SDK method can return, as any class can only extend 1 other class. But it can implement N interfaces. 

Why have you gone for this approach, and not the approach we use already which is to:
- for any class A that should be returned from SDK method B
- make A implement the CommandResult C that belongs to method B

so for example: 
``` 
open class BrowserRequired(...): Error(...), CommandResultA, CommandResultB, ...
```

this is the approach we're taking right now for result classes that produced from multiple SDK method. See e.g. `SignInResult.CodeRequired`

SammyO (changes requested):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
```diff
@@ -233,25 +233,17 @@ class ResetPasswordCodeRequiredState internal constructor(
...
                     Logger.warn(
                         TAG,
                         "Resend code received unexpected result: $result"
                     )
                     ResendCodeError(
-                        errorMessage = result.errorDescription,
-                        error = result.error,
-                        correlationId = result.correlationId,
-                        exception = result.exception
+                        errorMessage = (result as INativeAuthCommandResult.Error).errorDescription,
```
SammyO, line 243: My previous comment might also fix this need for casting.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
SammyO: +1 to Yuki. `ResetPasswordResendCodeCommandResult` shouldn't implement `Redirect` anymore, if the API can't return this state.
This will require changes in common.

SammyO (changes requested):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
GBakolas: With this approach we won't be able to call the isBrowserRequired() method from the existing error classes e.g SignInError.isBrowserRequired() and the application would need to handle two error classes SignInError and BrowserRequired errors.
Do we want to make BrowserRequired an individual error class or just call the isBrowserRequired method only from the error classes that can actually return it?

SammyO: Why not? You can place `isBrowserRequired` in Error, and have it check the same errorType field

SammyO: And in the future, when we've gotten rid of errorType altogether, we can replace it with an `is` check. E.g.
`fun isBrowserRequired(): Boolean = this is BrowserRequiredError`

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
GBakolas: As discussed since this is not a breaking change and this also overlaps with the general internal enum class cleanup is not needed to be done now. I will create a PBI for our backlog

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
GBakolas: Re-declared the BrowserRequiredError as an interface instead of a class what do you think about this approach?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
GBakolas: This is still neeed

Yuki-YuXin (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
```diff
@@ -79,7 +85,14 @@ open class Error(
     open var exception: Exception? = null,
     open val errorCodes: List<Int>? = null
 ) {
-    fun isBrowserRequired(): Boolean = this.errorType == ErrorTypes.BROWSER_REQUIRED
+}
+
+/**
+ * BrowserRequiredError error is an interface for all errors that could required a browser redirection in Native Auth.
```
SaurabhMSFT: `required` should be changed to `require`

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/SignInErrors.kt
```diff
@@ -62,3 +62,22 @@ class SignInSubmitPasswordError(
...
+ * @param error the error returned by the authentication server.
+ * @param errorMessage the error message returned by the authentication server.
+ * @param correlationId a unique identifier for the request that can help in diagnostics.
+ * @param errorCodes a list of specific error codes returned by the authentication server.
+ * @param exception an internal unexpected exception that happened.
+ */
+open class SignInContinuationError(
+    override val error: String? = null,
+    override val errorMessage: String?,
+    override val correlationId: String,
```
SaurabhMSFT, line 81: can you put a TODO that correlationId should be `String?` and needs to be changed after Sammy's PR is merged.

SaurabhMSFT (changes requested):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
GBakolas: Updated the comment

SaurabhMSFT (approved):

SammyO (approved):

PR #2010, opened 2024-01-18
Change the signature of signup/signin methods

SaurabhMSFT:
The signIn and signInUsingPassword methods have been refactored into one signIn method with optional password. 
The signUp and signUpUsingPassword methods have been refactored into one signUp method with optional password. 

PR for the common repo is https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2284


msal/src/main/java/com/microsoft/identity/nativeauth/INativeAuthPublicClientApplication.kt
```diff
@@ -71,89 +69,47 @@ interface INativeAuthPublicClientApplication : IPublicClientApplication {
...
      * @throws MsalClientException if an account is already signed in.
      */
-    suspend fun signUp(username: String, attributes: UserAttributes? = null): SignUpResult
+    suspend fun signUp(username: String, password: CharArray? = null, attributes: UserAttributes? = null): SignUpResult
 
     /**
      * Sign up the account starting from a username; callback variant.
      *
      * @param username username of the account to sign up.
+     * @param password password of the account to sign up.
```
SammyO: let's make it explicit that this is optional, like we do for attributes.

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -414,22 +280,24 @@ class NativeAuthPublicClientApplication(
...
         scopes: List<String>?
-    ): SignInUsingPasswordResult {
+    ): SignInResult {
         LogSession.logMethodCall(TAG, "${TAG}.signInUsingPassword")
         return withContext(Dispatchers.IO) {
             LogSession.logMethodCall(TAG, "${TAG}.signInUsingPassword.withContext")
 
             verifyNoUserIsSignedIn()
 
+            val hasPassword = password != null && !password.isEmpty()
```
SammyO, line 299: `password.isNullOrBlank()`

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -414,22 +280,24 @@ class NativeAuthPublicClientApplication(
...
+    override suspend fun signIn(
         username: String,
-        password: CharArray,
+        password: CharArray?,
         scopes: List<String>?
-    ): SignInUsingPasswordResult {
+    ): SignInResult {
         LogSession.logMethodCall(TAG, "${TAG}.signInUsingPassword")
         return withContext(Dispatchers.IO) {
             LogSession.logMethodCall(TAG, "${TAG}.signInUsingPassword.withContext")
```
SammyO: Update log (and perhaps also in other places)

msal/src/main/java/com/microsoft/identity/nativeauth/INativeAuthPublicClientApplication.kt
SaurabhMSFT: done

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
SaurabhMSFT: done

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -413,16 +374,16 @@ public static SignUpStartCommandParameters createSignUpStartCommandParameters(
      * @return Command parameter object
      * @throws ClientException
      */
-    public static SignUpStartUsingPasswordCommandParameters createSignUpStartUsingPasswordCommandParameters(
+    public static SignUpStartCommandParameters createSignUpStartCommandParameters(
             @NonNull final NativeAuthPublicClientApplicationConfiguration configuration,
             @NonNull final OAuth2TokenCache tokenCache,
             @NonNull final String username,
-            @NonNull final char[] password,
+            @Nullable final char[] password,
```
SammyO, line 381: update comment (line 373) to reflect the fact that password is optional

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -634,11 +554,11 @@ public static SignInStartCommandParameters createSignInStartCommandParameters(
      * @return Command parameter object
      * @throws ClientException
      */
-    public static SignInStartUsingPasswordCommandParameters createSignInStartUsingPasswordCommandParameters(
+    public static SignInStartCommandParameters createSignInStartCommandParameters(
             @NonNull final NativeAuthPublicClientApplicationConfiguration configuration,
             @NonNull final OAuth2TokenCache tokenCache,
             @NonNull final String username,
-            @NonNull final char[] password,
+            @Nullable final char[] password,
```
SammyO, line 561: Same - update comment

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -588,18 +492,18 @@ class NativeAuthPublicClientApplication(
...
 
             val parameters =
-                CommandParametersAdapter.createSignUpStartUsingPasswordCommandParameters(
+                CommandParametersAdapter.createSignUpStartCommandParameters(
                     nativeAuthConfig,
                     nativeAuthConfig.oAuth2TokenCache,
                     username,
                     password,
-                    attributes?.toMap()
+                    attributes?.userAttributes
```
SammyO: why this change?

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
```diff
@@ -2587,17 +2582,17 @@ public void testSignUpInvalidPasswordReturnsError() throws ExecutionException, I
                 MockApiResponseType.PASSWORD_TOO_WEAK
         );
 
-        SignUpUsingPasswordTestCallback signUpUsingPasswordTestCallback = new SignUpUsingPasswordTestCallback();
-        application.signUpUsingPassword(username, password, UserAttributes.Builder.customAttribute("password", password.toString()).build(), signUpUsingPasswordTestCallback);
+        SignUpTestCallback signUpTestCallback = new SignUpTestCallback();
+        application.signUp(username, password, UserAttributes.Builder.customAttribute("password", password.toString()).build(), signUpTestCallback);
```
SammyO: I never saw this before, but I find this strange. Why are we passing password as a custom attribute? It shouldn't matter technically, but we should remove it to avoid confusion.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
```diff
@@ -2606,13 +2601,13 @@ public void testSignUpUsingPasswordInvalidEmailReturnsError() throws ExecutionEx
                 MockApiResponseType.INVALID_USERNAME
         );
 
-        SignUpUsingPasswordTestCallback signUpUsingPasswordTestCallback = new SignUpUsingPasswordTestCallback();
-        application.signUpUsingPassword(invalidUsername, password, UserAttributes.Builder.customAttribute("password", password.toString()).build(), signUpUsingPasswordTestCallback);
+        SignUpTestCallback signUpTestCallback = new SignUpTestCallback();
+        application.signUp(invalidUsername, password, UserAttributes.Builder.customAttribute("password", password.toString()).build(), signUpTestCallback);
```
SammyO: Same here.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
```diff
@@ -2626,7 +2621,7 @@ public void testSignUpInvalidEmailReturnsError() throws ExecutionException, Inte
         );
 
         SignUpTestCallback signUpTestCallback = new SignUpTestCallback();
-        application.signUp(invalidUsername, UserAttributes.Builder.customAttribute("password", password.toString()).build(), signUpTestCallback);
+        application.signUp(invalidUsername, null, UserAttributes.Builder.customAttribute("password", password.toString()).build(), signUpTestCallback);
```
SammyO: And here

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
SaurabhMSFT: The password is nullable but is still required as Java does not support default values. Developers have to specify all parameters.

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
SaurabhMSFT: The password is nullable but is still required as Java does not support default values. Developers have to specify all parameters.

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
SammyO: This class is only used internally, but Kotlin classes (our state machine). In those cases, this field is optional. We've done the same for e.g. `scopes` parameter in other methods.

GBakolas (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -414,24 +280,26 @@ class NativeAuthPublicClientApplication(
...
-        LogSession.logMethodCall(TAG, "${TAG}.signInUsingPassword")
+    ): SignInResult {
+        LogSession.logMethodCall(TAG, "${TAG}.signIn")
         return withContext(Dispatchers.IO) {
-            LogSession.logMethodCall(TAG, "${TAG}.signInUsingPassword.withContext")
+            LogSession.logMethodCall(TAG, "${TAG}.signIn.withContext")
 
             verifyNoUserIsSignedIn()
 
+            val hasPassword = password != null && !password.isEmpty()
```
Yuki-YuXin: Maybe password.isNotEmpty()

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -568,15 +470,17 @@ class NativeAuthPublicClientApplication(
...
         username: String,
-        password: CharArray,
+        password: CharArray?,
         attributes: UserAttributes?
-    ): SignUpUsingPasswordResult {
-        LogSession.logMethodCall(TAG, "${TAG}.signUpUsingPassword(username: String, password: String, attributes: UserAttributes?)")
+    ): SignUpResult {
+        LogSession.logMethodCall(TAG, "${TAG}.signUp")
+
+        var hasPassword = password != null && !password.isEmpty()
```
Yuki-YuXin: The same here.

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
SaurabhMSFT: This is what it is on line 781 in the previous version of the file.

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
SaurabhMSFT: Done

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
SaurabhMSFT: Done

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
SaurabhMSFT: Done

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
SaurabhMSFT: Done

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
SaurabhMSFT: Done

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
SaurabhMSFT: Java does not have optional parameters. Optional parameters are simulated in Java by overloading functions. As this parameter is in the middle, it wouldn't be good style to overload function with reduced set of parameters.

Yuki-YuXin (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
SammyO: It seems we had an inconsistency between `signUp` and `signUpUsingPassword` when it comes to this field.
Could you validate that `attributes?.userAttributes` is the right way of doing this, and not `attributes?.toMap()`, by doing some E2E testing.

SammyO: `SignUpStates.submitAttribute` still uses `toMap()`, so this inconsistency needs to be fixed. Let's make sure we're not introducing a last minute bug in this flow.

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -414,24 +280,26 @@ class NativeAuthPublicClientApplication(
      * Sign in the account using username and password; Kotlin coroutines variant.
      *
      * @param username username of the account to sign in.
-     * @param password password of the account to sign in.
+     * @param password (Optional) password of the account to sign in.
```
SammyO, line 283: This needs to be removed, as the optionality doesn't apply to Java call sites. Applies in different places in the code.

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
SammyO: My point is that this method isn't called from any Java class, and so we don't have to consider the non-optionality of this parameter. You can leave it like this however, that's also fine.

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
SaurabhMSFT: `attributes?.userAttributes` and `attributes?.toMap()` return the same value. I have changed the usage here to `attributes?.toMap()`

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
SaurabhMSFT: The place where the optionality doesnot apply is in an internal package and not used from the Sample app. Optional in the doc should stay here as it provides meaningful information to the developer.

SammyO (approved):

PR #2012, opened 2024-01-18
Rename "is QR + PIN available" API to "get preferred auth method".

p3dr0rv:
After a discussion with the IOs team and OneAuth team we decided to change:
1. The API to return a more general PreferredAuthMethod enum rather than QRPINAvailability. This is to attempt to be extensible in case some other similar method is created as part of current discussions, such as NFC.
2. The admin config will reflect this - we'd ask for "preferred_auth_config= QRPIN" rather than "qrPinAvailability = true".


fadidurah (approved):

msal/src/main/java/com/microsoft/identity/client/IPublicClientApplication.java
```diff
@@ -146,12 +147,14 @@ void acquireToken(@NonNull final Activity activity,
     boolean isSharedDevice();
 
     /**
-     * Returns whether the application can support the QR code scanning + PIN protocol.
+     * Reads the preferred authentication method from Authenticator app Restriction Manager.
+     * If the Authenticator app is not installed, or not authentication method is preferred,
```
iamgusain:
````suggestion
     * If the Authenticator app is not installed, or no preferred authentication method is available,
````

**Comment by iamgusain** [comment]
> After a discussion with the IOs team and OneAuth team we decided to change:
> 
> 1. The API to return a more general PreferredAuthMethod enum rather than QRPINAvailability. This is to attempt to be extensible in case some other similar method is created as part of current discussions, such as NFC.
> 2. The admin config will reflect this - we'd ask for "preferred_auth_config= QRPIN" rather than "qrPinAvailability = true".

should it return a list of allowed AuthMethod in some order of preference determined by Admin?

File: msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java [hunk 2]
```diff
@@ -1512,22 +1513,22 @@ public void onCancel() {
             final CommandResult commandResult = future.get();
```
iamgusain: We should think about having a timeout for the  future.get() call, so it doesn't wait indefinitely. We can start with a higher timeout.

iamgusain (approved):
Left some suggestions.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
p3dr0rv: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2012/commits/0098cd938230945af5a26a063731e4b21164d194

PR #2021, opened 2024-01-25
Move otelexporter code from its own module to AADAuthenticator

shahzaibj:
Related: https://github.com/AzureAD/ad-accounts-for-android/pull/2675

### DIFF
--- a/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
+++ b/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
@@ -49,14 +49,13 @@ import com.microsoft.identity.client.exception.MsalClientException;
 import com.microsoft.identity.client.exception.MsalException;
 import com.microsoft.identity.client.exception.MsalServiceException;
 import com.microsoft.identity.client.exception.MsalUiRequiredException;
-import com.microsoft.identity.client.opentelemetry.exporter.AriaInitializer;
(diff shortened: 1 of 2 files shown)


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MainActivity.java
```diff
@@ -168,15 +172,28 @@ protected void onCreate(Bundle savedInstanceState) {
      * @param applicationContext the application context
      */
     private static synchronized void initOpenTelemetry(@lombok.NonNull final Context applicationContext) {
-        AriaInitializer.initializeAria(applicationContext, BuildConfig.otelAriaToken);
         final Resource resource = Resource.getDefault();
 
-        final AriaSpanExporter ariaSpanExporter = new AriaSpanExporter(
```
p3dr0rv, line 174: @shahzaibj could you please explain why this change?

shahzaibj: AriaSpanExporter was being used to upload spans created in msal test app to Aria (These are spans in MSAL layer - not in the Broker). This was done to test span context propagation work I did for Yammer. Now that otelexporter is removed as local dependency I removed this piece of code as well. The same test can be performed by just printing msal side spans to console.

Alternatively we can keep dist dependency (on otelexporter already published) but I didn't see a reason to do that.

rpdome (approved):

p3dr0rv (approved):

PR #2029, opened 2024-01-31
Support CIAM custom URL domains

SammyO:
This PR adds support for CIAM custom URL domains (CUD), through addressing the following:
1. `Account.getAuthority()` looks at the `iss` claim to determine whether an authority is of type CIAM.
 Since CIAM token issuer split, the issuer claim will contain `ciamlogin.com`. 
**Note**: this solution will work as long as the issuer of a CUD is not updated to reflect the custom domain. This will happen, in +- 1 year.

2.  `Account.getAuthority()` returns a CIAM authority with URL format `https://<authorityUrl>/<tenantId>`(e.g. `https://custom.domain.com/<tenantId>` or `https://<tenantName>.ciamlogin.com/<tenantId>`), rather than the previous `https://<authorityUrl>/<tenantName>.onmicrosoft.com`. The latter can't be used to compose OpenID connect URL for CUD variants.

**Note**: no E2E tests can be built for this feature yet, as the Labs environment is not set up. This is an ongoing effort (by Sammy and Ryan), see: https://identitydivision.visualstudio.com/Engineering/_dashboards/dashboard/4398d9a6-a49b-4a4a-a9b8-58ca2d24c6ab

MSAL common PR:  https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2314

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/Account.java
+++ b/msal/src/main/java/com/microsoft/identity/client/Account.java
@@ -180,7 +180,8 @@ public class Account implements IAccount {
             if (!StringUtil.isEmpty(iss)) {
                 // We're using the issuer to determine whether this is a CIAM authority, as the environment
                 // be a custom-URL-domain variant, which won't contain ciamlogin.com.
-                // Note: this will work as long as the issuer for CUD doesn't change.
+                // Note: this will work as long as the issuer for CUD doesn't change, which it will
+                // in the future. See https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2832277
                 if (iss.contains(CIAMAuthority.CIAM_LOGIN_URL_SEGMENT) && !StringUtil.isEmpty(getEnvironment())
                         && !StringUtil.isEmpty(getTenantId())) {
                     // Call static method in CIAMAuthority to create the full authority uri
--- a/msal/src/main/java/com/microsoft/identity/client/Account.java
+++ b/msal/src/main/java/com/microsoft/identity/client/Account.java
@@ -173,11 +173,6 @@ public class Account implements IAccount {
     @Override
     @NonNull
     public String getAuthority() {
-        // If the environment shows CIAM, we should return an authority of format https://tenant.ciamlogin.com/tenant.onmicrosoft.com
-        if (getEnvironment() != null && getEnvironment().contains(CIAMAuthority.CIAM_LOGIN_URL_SEGMENT)) {
-            // Call static method in CIAMAuthority to create the full authority uri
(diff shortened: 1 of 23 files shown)


msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -173,16 +173,20 @@ public String getUsername() {
...
         // TODO: The below logic only works for the case of AAD. We need to refactor this once we
         //  make a proper fix for B2C
         if (null != getClaims()) {
             final String iss = (String) getClaims().get("iss");
             if (!StringUtil.isEmpty(iss)) {
+                // We're using the issuer to determine whether this is a CIAM authority, as the environment
+                // be a custom-URL-domain variant, which won't contain ciamlogin.com.
+                // Note: this will work as long as the issuer for CUD doesn't change, which it will
+                // in the future. See https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2832277
+                if (iss.contains(CIAMAuthority.CIAM_LOGIN_URL_SEGMENT) && !StringUtil.isEmpty(getEnvironment())
```
fadidurah, line 185: Comment is a little confusing, so we know the issuer will change eventually? Linked Item has no description

fadidurah (approved):

msal/src/main/java/com/microsoft/identity/client/Account.java
SammyO: It's highly likely that the issuer will change, yes. But the full scope and business requirements of that effort are unknown (as well as timeline), so it doesn't make sense to optimise for that now.
Indeed the ADO ticket has no description; it's been difficult to get clarity on this from the owning PM and EM. They're working on other more high priority stuff. 

Let me know if there is a better way to deal with this.

msal/src/main/java/com/microsoft/identity/client/Account.java
```diff
@@ -173,16 +173,20 @@ public String getUsername() {
...
-            return CIAMAuthority.getFullAuthorityUrlFromAuthorityWithoutPath(getEnvironment());
-        }
         // TODO: The below logic only works for the case of AAD. We need to refactor this once we
         //  make a proper fix for B2C
         if (null != getClaims()) {
             final String iss = (String) getClaims().get("iss");
             if (!StringUtil.isEmpty(iss)) {
+                // We're using the issuer to determine whether this is a CIAM authority, as the environment
+                // be a custom-URL-domain variant, which won't contain ciamlogin.com.
+                // Note: this will work as long as the issuer for CUD doesn't change, which it will
```
shahzaibj, line 183: What's CUD?

shahzaibj (approved):

PR #2031, opened 2024-02-01
Add "Sign in from other device" switch

mohitc1:
Adding "Sign in from other device" to MsalTestApp.

When toggled ON, it sends is_remote_login_allowed=true extra query parameter to /authorize call. When interactive brokered auth is started with login hint, this should show "Sign in from other device" option in the UI.

This would help us setup manual test case as well as debugging.


shahzaibj (approved):

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/AcquireTokenFragment.java
```diff
@@ -599,6 +602,7 @@ private RequestOptions getCurrentRequestOptions() {
         final String claims = mClaims.getText().toString();
         final boolean enablePII = mEnablePII.isChecked();
         final boolean forceRefresh = mForceRefresh.isChecked();
+        final boolean allowSignInFromOtherDevice = mAllowSignInFromOtherDevice.isChecked();
```
somalaya, line 605: just curious, do we really need this option? I mean, what is the harm in always having to pass the is_remote_login_allowed flag as true? It won't be exercised unless user clicks on the "sign in from another device" button in ests page right?

fadidurah: I don't think there's a way currently in msal test app UI to pass anything through withAuthorizationQueryStringParameters() in the AcquireTokenParameters object. I think having it as an option makes sense, because we may want to test without it, instead of always passing it in. We could also introduce a new  ui field to pass in custom extra query string parameters in general.

mohitc1: We should also test with not passing it, that's why. I should a UI test for that as well, I will update the other automation pr with it. "Sign in from other device" should not be there, if this flag is not passed.

somalaya (approved):

PR #2040, opened 2024-02-14
Give BrokerMsalController higher priority than LocalMsalController if available

iamgusain:
### What 
Giving BrokerMsalController higher priority than LocalMsalController if available. This means that if Broker is available, msal will first try to acquire token via broker before trying its localController

### Why
In order to push brokered auth as default when available. And reduce usage of regular RT, when PRT can be used which is bound to the device.

### How
Changing the order of controllers to put BrokerMsalController (if eligible) before LocalMsalController 

### Testing
Verified AcquireTokenSilent calls reach Broker first.

### DIFF
--- a/changelog
+++ b/changelog
@@ -2,6 +2,7 @@ MSAL Wiki : https://github.com/AzureAD/microsoft-authentication-library-for-andr
 
 vNext
 ----------
+[MINOR] Give BrokerMsalController higher priority than LocalMsalController if available (#2040)
 
 Version 5.1.0
 ----------
--- a/msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.kt
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.kt
@@ -89,16 +89,15 @@ class MSALControllerFactory(
     /**
      * Returns one or more controllers to address a given request.
      *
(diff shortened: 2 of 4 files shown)


rpdome:
What would happen in the following scenario
1. Sign in with MSAL only
2. Install Broker
3. try to get a token again

would the user be prompted?

mohitc1:
>      * 5) The broker redirect URI for the client is registered

This comment should be revised based on final logic in the method.

---
Refers to: msal/src/main/java/com/microsoft/identity/client/internal/controllers/MSALControllerFactory.kt:101 in 986bd62. [](commit_id = 986bd626f9a32e33217fb6915a76fc3c9700aabf, deletion_comment = False)

iamgusain:
> What would happen in the following scenario
> 
> 1. Sign in with MSAL only
> 2. Install Broker
> 3. try to get a token again
> 
> would the user be prompted?

Yes

iamgusain:
> What would happen in the following scenario
> 
> 1. Sign in with MSAL only
> 2. Install Broker
> 3. try to get a token again
> 
> would the user be prompted?

Silent requests will fallback to local controller so no prompt.
Interactive requests will anyway go to broker so users will be prompted.

fadidurah (approved):

somalaya (approved):

PR #2044, opened 2024-02-23
Removing Passkey feature flag

melissaahn:
### Summary
See https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2329.
Since MSAL is going to be flighted by the consumers, we'll remove the feature flag in general from MSAL.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
@@ -1190,9 +1190,7 @@ public class CommandParametersAdapter {
     public static List<Map.Entry<String, String>> appendToExtraQueryParametersIfWebAuthnCapable(
             @Nullable final List<Map.Entry<String, String>> queryStringParameters,
             @NonNull final PublicClientApplicationConfiguration configuration) {
-        //Putting behind passkey feature flag until feature is ready.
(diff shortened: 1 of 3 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -1189,7 +1191,7 @@ public static List<Map.Entry<String, String>> appendToExtraQueryParametersIfWebA
             @Nullable final List<Map.Entry<String, String>> queryStringParameters,
             @NonNull final PublicClientApplicationConfiguration configuration) {
         //Putting behind passkey feature flag until feature is ready.
-        if (FidoConstants.IS_PASSKEY_SUPPORT_READY
+        if (CommonFlightManager.isFlightEnabled(CommonFlight.ENABLE_PASSKEY_FEATURE)
```
mohitc1: I don't think MSAL side can be rolled back remotely.

mohitc1 (approved):

fadidurah (approved):

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
melissaahn: True, and the feature is now slated for next release, so I'm going to remove this flag from MSAL in general.

mohitc1 (approved):

PR #2048, opened 2024-02-28
Edit test app to test QR + PIN feature

p3dr0rv:
- Add camera permission to MSAL test app, this is needed to test QR + PIN flow w/o broker. MSAL consumer apps who want to opt in in QR + PIN flow they will need to add this permission on their manifest.
- Add a configuration for PPE and PPE WebView. This is possible now that we have the msal redirect url on PPE see: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2852767
Thanks to Melissa and Ryan


melissaahn (approved):
Thank you!

rpdome (approved):

PR #2061, opened 2024-03-19
Modifying NAA tests to run with broker host

somalaya:
Issue : NAA Tests in this pipeline run are failing https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1270279&view=ms.vss-test-web.build-test-results-tab&runId=3448652&resultId=100001&paneView=debug because they can only run with brokerHost app. We have a logic https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/12d96b11ac124211e0cb0c7e0c9e0dadd07199e1/common/src/main/java/com/microsoft/identity/common/internal/platform/AndroidPlatformUtil.java#L140 that checks if the broker app is a debug app that the tests depend on. 

Fix : CP and Auth app debug versions are not produced in the pipeline currently. So we have to rely on brokerHost debug apk produced.

### DIFF
--- a/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/nestedAppAuth/TestCase2688459.java
+++ b/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/nestedAppAuth/TestCase2688459.java
@@ -30,6 +30,8 @@ import com.microsoft.identity.common.java.dto.AccountRecord;
 import com.microsoft.identity.labapi.utilities.client.LabQuery;
 import com.microsoft.identity.labapi.utilities.constants.TempUserType;
 import com.microsoft.identity.labapi.utilities.constants.UserType;
+import com.microsoft.identity.client.ui.automation.annotations.LocalBrokerHostDebugUiTest;
+import com.microsoft.identity.client.ui.automation.annotations.SupportedBrokers;
 
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -40,6 +42,8 @@ import java.util.List;
 
 // Nested App auth silent request
 // https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2688459
+@SupportedBrokers(brokers = BrokerHost.class)
+@LocalBrokerHostDebugUiTest
 @RunWith(Parameterized.class)
 public class TestCase2688459 extends AbstractMsalBrokerTest {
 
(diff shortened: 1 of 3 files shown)


p3dr0rv (approved):

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/nestedAppAuth/TestCase2688459.java
```diff
@@ -40,6 +42,8 @@
 
 // Nested App auth silent request
 // https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2688459
+@SupportedBrokers(brokers = BrokerHost.class)
+@LocalBrokerHostDebugUiTest
```
shahzaibj, line 46: Why are two similar annotations required?

somalaya: I think one of them (LocalBrokerHost) is for the pipeline to identify that it needs to run with brokerHost.

shahzaibj: I'm not sure why the pipeline needs to identify it. The test infra today has capability to skip (ignore) tests has broker variant used is not in the list of SupportBrokers for the test.

somalaya: I am looking at the daily pipeline code and I see that there is a separate annotation being used to run brokerHost tests in a separate stage. That is probably the reason why we have 2 annotation for all tests that need to run on brokerHost.
https://identitydivision.visualstudio.com/Engineering/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=1490&branch=master&nonce=vCbn0InOkEbjZ%2BgVrNEk0Q%3D%3D

fadidurah: There's a separate stage to run with local brokerhost, distAutoBroker is not always usable for tests that require running on brokerhost, I made the change a while ago but i think some tests would only automate in a separate stage, the separate stage is required because we need to run test on an msal automation app with a different build variant

fadidurah (approved):

shahzaibj (approved):

PR #2071, opened 2024-04-08
Clean up native auth msal logging

Yuki-YuXin:
### Goal:
- Review our log statements and reduce to only those that are relevant and useful. Especially the logMethodCall should be much reduced. Not every method call is relevant to know about when debugging.
- Double check that the existing logs use string interpolation correctly to avoid leaking sensitive information in the logs.

### Changes summary:
- Apply the rule "only logger unknown error" to NativeAuthPublicClientApplication and states. Exception: Sign in has password CodeRequired since it's unusual for the case.
- Add arguments under method LogSession.logMethodCall tp distinguish callback and coroutine.

### Company PR:
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2396


msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -278,11 +283,6 @@ class NativeAuthPublicClientApplication(
         scopes: List<String>?,
         callback: SignInCallback
     ) {
-        LogSession.logMethodCall(
```
SammyO, line 291: The reason for having it here as well as in `suspend fun signIn()` is so we know whether the callback or coroutine method was called. You don't think that's useful information?

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -372,11 +373,6 @@ class NativeAuthPublicClientApplication(
                         }
                     }
                     is SignInCommandResult.CodeRequired -> {
-                        Logger.warn(
```
SammyO: We have this log here on purpose, please leave it in (notice how it's a warning and not an info). This is to deal with the scenario where a developer calls signIn(username, password) with an account that's an email + OTP signup. The password is silently ignored; this log tries to warn developers of that. Perhaps not the best solution, but best to leave it in for now.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -82,11 +82,6 @@ class AccountState private constructor(
      * @param callback [com.microsoft.identity.nativeauth.statemachine.states.AccountState.SignOutCallback] to receive the result on.
      */
     fun signOut(callback: SignOutCallback) {
-        LogSession.logMethodCall(
```
SammyO, line 90: isn't this useful to know? I.e. what SDK methods the developer called.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
```diff
@@ -410,11 +395,6 @@ class ResetPasswordPasswordRequiredState internal constructor(
                     }
 
                     is ResetPasswordCommandResult.UserNotFound -> {
-                        Logger.warn(
```
SammyO: re. https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2374/files#r1555927999 - if we add in some form of CommandResult logging, then I agree we can remove this log here. But I would like to have a way of seeing what the CommandResult was that the internal SDK business logic produced, so that we for example can determine that it was an "unexpected result" in the SDK interface and why we're returning an Error to the developer. Wdyt?

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
Yuki-YuXin: Agreed. Callback or Coroutine is also something we want to gather for telemetry. The reason I removed them is because for callback method, there would be two repeating same logs.
Re-add the logs back to  callback methods and use parameters to differentiate between callback and coroutine.

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
Yuki-YuXin: Sure. The reason is strong here.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordStates.kt
Yuki-YuXin: CommandResult has been put back in that PR.

Yuki-YuXin:
Better to merge after multi access token work.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -90,7 +90,7 @@ class AccountState private constructor(
         LogSession.logMethodCall(
             tag = TAG,
             correlationId = null,
-            methodName = "$TAG.signOut"
+            methodName = "$TAG.signOut(callback: SignOutCallback)"
```
GBakolas: Missing brackets -> {TAG}

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -220,7 +220,7 @@ class AccountState private constructor(
         LogSession.logMethodCall(
             tag = TAG,
             correlationId = null,
-            methodName = "$TAG.getAccessToken"
+            methodName = "$TAG.getAccessToken(forceRefresh: Boolean = false, callback: GetAccessTokenCallback)"
```
GBakolas: Missing brackets -> {TAG}

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -295,7 +295,7 @@ class AccountState private constructor(
         LogSession.logMethodCall(
             tag = TAG,
             correlationId = null,
-            methodName = "$TAG.getAccessToken(forceRefresh: Boolean)"
+            methodName = "$TAG.getAccessToken(forceRefresh: Boolean = false)"
```
GBakolas: Missing brackets -> {TAG}

GBakolas: The correct methodName here should be getAccessTokenInternal(forceRefresh: Boolean, scopes: List<String>) right?

Also, maybe is better to print the actual value of the forceRefresh parameter to quickly identify if this was a forced token refresh by the application

GBakolas (commented):
In the AccountState.kt file there are more places where brackets are missing around the TAG variable when logging the method name.

In the same file there are getAccessToken() functions with no logMethodCall() so we can't identify if the coroutine or the callback variant was called

rmccahill (approved):

GBakolas (approved):

PR #2080, opened 2024-04-16
Add ClientException wrapper for native auth

Yuki-YuXin:
### Goal:
The SDK currently throws Exceptions in some flows (e.g. ClientException when an empty username is passed, refresh token exceptions, etc.). We should refactor this and replace it with exceptions being wrapped into an Error class and returned as a normal SDK response.  Calling applications shouldn't crash, and shouldn't require try-catch blocks.
This will also involve updating the sample app (removing try-catch surrounding SDK methods being called)

### Investigation summary:
- The incorrect configuration throws MsalClientException which it's inevitable and unsolvable by error class wrapper.
- Under CommandResultUtil.kt, the Exception has been wrapped as UnknownError.
```kotlin
 if (this.status != ICommandResult.ResultStatus.COMPLETED) {
        var exception: Exception? = null
        var exceptionMessage: String? = ""

        if (this.result is Exception) {
            exception = this.result as Exception
            exceptionMessage = exception.message
        }

        return com.microsoft.identity.common.java.nativeauth.controllers.results.INativeAuthCommandResult.UnknownError(
            error = UNSUCCESSFUL_COMMAND_ERROR,
            errorDescription = exceptionMessage,
            exception = exception,
            correlationId = this.correlationId
        ) as ExpectedType
```
- BaseNativeAuthController throws ClientException. They could only be called in NativeAuthMsalController and we didn't.
- NativeAuthMsalController throws IOException::class, ClientException::class, ServiceException::class ArgumentException::class, ServiceException::class. mainly from acquireTokenSilent & renewAT used by AcquireTokenNoFixedScopesCommand (accountState.getAccessToken). It will be handled by `is Exception` in getAccessToken`
- AcquireTokenNoFixedScopesCommandParameters throws ArgumentException -> this parameter will be delted
- Request throws ClientException -> NativeAuthRequestHandlerTest include

-  verifyNoUserIsSignedIn() throws MsalClientException

### Changes summary:
1. Add GetAccountError and SignOutError error in Error.kt
2. Use try catch block with corresponding errors in interface methods.

### Company PRs:
native sample app: https://github.com/Azure-Samples/ms-identity-ciam-native-auth-android-sample/pull/24


msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -102,63 +105,75 @@ class AccountState private constructor(
...
+                        "Unexpected error during signOut."
+                    )
+                }
+            } catch (e: Exception) {
+                Logger.error(TAG, "MSAL client exception occurred in getCurrentAccount.", e)
+                ClientExceptionError(
+                    errorType = ErrorTypes.CLIENT_EXCEPTION,
+                    errorMessage = "MSAL client exception occurred in getCurrentAccount.",
+                    exception = e,
+                    correlationId = DiagnosticContext.INSTANCE.threadCorrelationId
```
SammyO: Setting a correlation ID here is wrong. It should already have been set earlier in the flow, and the existing value should be used.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -102,63 +105,75 @@ class AccountState private constructor(
...
+                        null
+                    )
+                    throw MsalClientException(
+                        MsalClientException.UNKNOWN_ERROR,
+                        "Unexpected error during signOut."
+                    )
+                }
+            } catch (e: Exception) {
+                Logger.error(TAG, "MSAL client exception occurred in getCurrentAccount.", e)
+                ClientExceptionError(
```
SammyO: I'm not sure I agree with introducing a new error type. Why not use the existing `UnknownError` for this? We should weigh the pros and cons of different options. IMO:
- to a developer `ClientExceptionError` isn't that informative. It informs that a client exception happened, but not what type - they need to look at the error details (exception field) to understand more. Which is similar to the experience of `UnknownError`. 
- we should keep the interface as crisp and clean as possible. Introducing too many types and classes can be confusing and overwhelming, rather than informative.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
Yuki-YuXin: Use "UNSET" here right now.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
Yuki-YuXin: For sign in, sign up, reset password cases, use their own errors instead (SignInError, SignUpError, ResetPasswordError). for those do not have corresponding errors, keep ClientExceptionError as a placeholder.

SammyO: Sorry, `UnknownError` is not the right class. I mean to say: why not use the existing generic error classes for this case, i.e. `SignInError`, `SignupError`, `ResetPasswordError`. Similar to 
image

Yuki-YuXin: For those without the existing generic error classes.  Do we need to create those generic error classes for them?  I personally think it would be better than ClientExceptionError

Yuki-YuXin:
Hang this PR until the rest of the PRs get merged and then bring the changes back in.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignUpStates.kt
```diff
@@ -609,89 +639,103 @@ class SignUpAttributesRequiredState internal constructor(
...
-                            continuationToken = result.continuationToken,
+                    is INativeAuthCommandResult.Redirect -> {
+                        SignUpSubmitAttributesError(
+                            errorType = ErrorTypes.BROWSER_REQUIRED,
+                            error = result.error,
+                            errorMessage = result.errorDescription,
+                            correlationId = result.correlationId
+                        )
+                    }
+                    // This should be caught earlier in the flow, so throwing UnexpectedError
```
SammyO: nit: let's update this comment: `// This should be caught earlier in the flow, so returning generic error`

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
```diff
@@ -606,24 +605,18 @@ public void onError(@NonNull BaseException exception) {
...
-                signInExceptionResult.setResult(exception);
-            }
-        };
+        application.signIn(
+                username,
+                password,
+                null,
+                signInWithPasswordResultTestCallback
+        );
+        SignInResult signInError = signInWithPasswordResultTestCallback.get();
```
SammyO, line 636: I actually don't love this. I hadn't paid close attention to this before, but an error state should be returned in the `onError()` callback. And the callback signature should change from `exception` to `SignInError`. We can park this for now, but I consider that a future improvement.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -1179,14 +1164,9 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
-            application.resetPassword(username)
-        } catch (exception: MsalException) {
-            assertEquals(MsalClientException.INVALID_PARAMETER, exception.errorCode)
-            assertEquals("An account is already signed in.", exception.message)
-            return@runBlocking
-        }
-        fail() // An exception should happen
+        val newResult = application.resetPassword(username)
+        assertTrue(newResult is ResetPasswordError)
+        assertTrue((newResult as ResetPasswordError).exception?.message.equals("An account is already signed in."))
```
SammyO, line 1187: Do you know of any scenarios that throw an exception deeper in the SDK? If so, could you write a test for that as well? To ensure `CommandDispatcher.submitSilentReturningFuture` and `rawCommandResult.checkAndWrapCommandResultType` logic work as expected as well. 

If there are no known scenarios where this happens, then feel free to resolve this.

SammyO (changes requested):

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
Yuki-YuXin: This improvement might worth a separate work item because it might be lots of changes.
```kotlin
NativeAuthPublicClientApplication.pcaScope.launch {
            try {
                val result = submitPassword(password)
                callback.onResult(result)
            } catch (e: MsalException) {
                Logger.error(TAG, "Exception thrown in submitPassword", e)
                callback.onError(e)
            }
        }
```

rmccahill (approved):

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Scenarios that throw an exception deeper in the SDK:
- BaseNativeAuthController throws ClientException. They could only be called in NativeAuthMsalController and we didn't.
- NativeAuthMsalController throws IOException::class, ClientException::class, ServiceException::class ArgumentException::class, ServiceException::class. mainly from acquireTokenSilent & renewAT used by AcquireTokenNoFixedScopesCommand (accountState.getAccessToken). -> handled by is Exception in getAccessToken`
- AcquireTokenNoFixedScopesCommandParameters throws ArgumentException if `null == authenticationScheme` -> this parameter will be delted
- Request throws ClientException if empty arguments -> handled by CommandResultUtil.kt
- verifyNoUserIsSignedIn() throws MsalClientException -> handled by this PR

exception during command execution will be handled by CommandResultUtil.kt. The Exception has been wrapped as INativeAuthCommandResult.UnknownError and then return as generic error in the interface layer.
Except than getAccessToken, which has its own exception wrapper as below. 
```kotlin
return@withContext when (commandResult) {
                    is ServiceException -> {
                        GetAccessTokenError(
                            exception = ExceptionAdapter.convertToNativeAuthException(commandResult),
                            correlationId = "UNSET"
                        )
                    }

                    is Exception -> {
                        GetAccessTokenError(
                            exception = commandResult,
                            correlationId = "UNSET"
                        )
                    }
```

SammyO (approved):

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: - Request throws ClientException if empty arguments handled by CommandResultUtil.kt -> Add **testEmptyRequestParametersToGenericErrorNotThrownException()** 

- acquireTokenSilent & renewAT handled by is Exception in getAccessToken` -> Not so easy (maybe also not suitable) to trigger SilentTokenCommandParameters.validate() to throw ArgumentException in integration test.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -2292,4 +2271,35 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+        configureMockApi(
+            endpointType = MockApiEndpoint.SignUpContinue,
+            correlationId = correlationId,
+            responseType = MockApiResponseType.INVALID_OOB_VALUE
+        )
+
+        val submitCodeState = spy((result as SignUpResult.CodeRequired).nextState)
+        val submitCodeResult = submitCodeState.submitCode(emptyString)  // Empty code will trigger ArgUtils.validateNonNullArg(oob, "oob") of the SignUpContinueRequest to thrown ClientException
+        assertTrue(submitCodeResult is SubmitCodeError)
+        assertTrue((submitCodeResult as SubmitCodeError).error.equals(ErrorTypes.UNSUCCESSFUL_COMMAND_ERROR))  // ClientException will be caught in CommandResultUtil.kt and converted to generic error in interface layer
```
SammyO: Where is `ErrorTypes.UNSUCCESSFUL_COMMAND_ERROR` set?

Yuki-YuXin: It set under  CommandResultUtil.kt `const val UNSUCCESSFUL_COMMAND_ERROR = "unsuccessful_command"`

SammyO: I don't think those changes are pushed, I don't see them in this PR.

Yuki-YuXin: They originally existed in the common repository.
image

SammyO: But this PR also introduces `const val UNSUCCESSFUL_COMMAND_ERROR = "unsuccessful_command"` in Error.kt. Is that one used?

Yuki-YuXin: It isn't used in the SDK code base because this PR just adds try catch block with ErrorType = "client_exception". 
ErryType = "unsuccessful_command" is defined wrapping exceptions in common repo under CommandResultUtil.
However, the existing "unsuccessful_command" has some difference from "client_exception" in my opinion. They can't replace each other.
"unsuccessful_command" is used in the appending test testEmptyRequestParametersToGenericErrorNotThrownException() and will be displayed in the sample app to user SubmitError(error= "unsuccessful_command").
We cannot use `assertTrue((submitCodeResult as SubmitCodeError).error.equals(CommandResultUtil.UNSUCCESSFUL_COMMAND_ERROR)) ` instead of `assertTrue((submitCodeResult as SubmitCodeError).error.equals(ErrorTypes.UNSUCCESSFUL_COMMAND_ERROR)) ` in the `NativeAuthPublicClientApplicationKotlinTest`
We can either remove testEmptyRequestParametersToGenericErrorNotThrownException() test or remove the introduction.

Yuki-YuXin: Offline discussion conclusion: Use direct string in the test file instead of adding new definition in the code base.

SammyO (approved):

PR #2082, opened 2024-04-19
Support for multiple tokens in native auth flows

SaurabhMSFT:
This PR and its companion PR (https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2390)  enables multiple access token support for Native Auth. A pair of new method have been added in the AccountState class to allow developer to retrieve different access tokens according to the requested scopes. The way these methods acquire tokens is exactly as same as the one in MSAL.  The old method 'getAccessToken''s functionality hasn't been changed. It still returns the access token matching the default OIDC scope. This method though has been deprecated.


msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -330,6 +332,7 @@ public static AcquireTokenNoFixedScopesCommandParameters createAcquireTokenNoFix
                 .authority(authority)
                 .authenticationScheme(authenticationScheme)
                 .forceRefresh(forceRefresh)
+                .scopes(scopes)
```
SammyO: Why do we still need `AcquireTokenNoFixedScopesCommandParameters` and `AcquireTokenNoFixedScopesCommand`? Now that we're adding scopes, aren't they the same as `SilentTokenCommandParameters`? Using the latter is how I envision the reusing of existing MSAL components working. Did you explore this avenue? If so, what are your findings?

SaurabhMSFT: yes, we can remove these two classes. I looked at it and there are significant commonalities between these classes. In this set of PR, I will focus on multiple tokens and in the next PR I will remove the `AcquireTokenNoFixedScopes*` classes. This would make the review process easier.

SammyO: > In this set of PR, I will focus on multiple tokens and in the next PR I will remove the AcquireTokenNoFixedScopes* classes. 

What I mean is that to support multiple tokens, we should be using `SilentTokenCommand`. So if your goal with this PR is to explore how we can support multiple tokens, then I'd like to see the use of `SilentTokenCommand`. Unless you've found that using that existing command doesn't work - if so please outline your findings.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -227,42 +230,79 @@ class AccountState private constructor(
...
+     * Kotlin coroutines variant.
+     *
+     * @return [com.microsoft.identity.nativeauth.statemachine.results.GetAccessTokenResult] The result of the getAccessToken action
+     */
+    suspend fun getAccessToken(forceRefresh: Boolean = false, scopes: List<String>): GetAccessTokenResult {
+        var scopeSet = HashSet<String>()
+        if (!scopes.isNullOrEmpty()) {
+            scopeSet.addAll(scopes)
+        }
+        val scopeSet2 = BaseController.addDefaultScopes(scopeSet)
```
SammyO: We need to dive into this a bit deeper, but I'm leaving this comment here as a reminder to myself:
I'm not sure this is correct. I understood from Danilo's technical document that the acceptance criteria for "no scopes passed" is that the SDK returns the latest (most recent) token that was added. Which could be the one that has default scopes, but might not.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -58,6 +60,7 @@ import com.microsoft.identity.nativeauth.utils.serializable
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import java.util.Collections
 
 /**
```
SammyO, line 67: As discussed in the technical design document review, the documentation in comments needs to be updated to explain the difference between these two methods.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SammyO: Let's give this a more meaningful name than `scopeSet2` please.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -227,42 +230,79 @@ class AccountState private constructor(
...
+    suspend fun getAccessToken(forceRefresh: Boolean = false, scopes: List<String>): GetAccessTokenResult {
+        var scopeSet = HashSet<String>()
+        if (!scopes.isNullOrEmpty()) {
+            scopeSet.addAll(scopes)
+        }
+        val scopeSet2 = BaseController.addDefaultScopes(scopeSet)
+        return getAccessTokenInternal(forceRefresh, ArrayList(scopeSet2))
+    }
+
+    suspend fun getAccessTokenInternal(forceRefresh: Boolean, scopes: List<String>?): GetAccessTokenResult {
```
SammyO: should `scopes` be nullable?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -227,42 +230,79 @@ class AccountState private constructor(
...
-                .build()
+                .forceRefresh(forceRefresh)
+
+            if (scopes != null ) {
+                acquireTokenSilentParametersBuilder.withScopes(scopes)
+            }
+
+            val acquireTokenSilentParameters = acquireTokenSilentParametersBuilder.build()
 
             val accountToBeUsed = PublicClientApplication.selectAccountRecordForTokenRequest(
```
SammyO: maybe I'm overlooking it, but where is this `accountToBeUsed` param used further in the flow?

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -550,6 +553,117 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
         assertEquals(accessToken, accessTokenTwo)
     }
 
+    /**
+     * Test sign in, get access token. Compare to token from getAccount()
+     */
+    @Test
```
SammyO, line 577: I assume more tests are following, to test the wide range of use cases that we have established for this feature?

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
SammyO: @SaurabhMSFT this comment is still relevant. Why do we still need `AcquireTokenNoFixedScopesCommandParameters` and `AcquireTokenNoFixedScopesCommand`?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -227,42 +230,79 @@ class AccountState private constructor(
      * @return [com.microsoft.identity.nativeauth.statemachine.results.GetAccessTokenResult] The result of the getAccessToken action
      */
     suspend fun getAccessToken(forceRefresh: Boolean = false): GetAccessTokenResult {
```
SammyO, line 243: Support for new function with callback (Java variant) is missing.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -227,42 +230,79 @@ class AccountState private constructor(
...
+            val account = accountResult
                     ?: return@withContext GetAccessTokenError(
                         errorType = GetAccessTokenErrorTypes.NO_ACCOUNT_FOUND,
                         error = MsalClientException.NO_CURRENT_ACCOUNT,
                         errorMessage = MsalClientException.NO_CURRENT_ACCOUNT_ERROR_MESSAGE,
                         correlationId = "UNSET"
                     )
 
-            val acquireTokenSilentParameters = AcquireTokenSilentParameters.Builder()
+            val acquireTokenSilentParametersBuilder = AcquireTokenSilentParameters.Builder()
```
SammyO: What is the difference between `AcquireTokenSilentParameters` and `SilentTokenCommandParameters`? Why are we using one and not the other?

SammyO: Ah, `AcquireTokenSilentParameters` is for the "interface commands" and `SilentTokenCommandParameters` is for the "internal commands"? Is there a need for us to use `AcquireTokenSilentParameters` - i.e. can we directly use `SilentTokenCommandParameters`?

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -550,6 +553,117 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+        assertEquals(accessTokenTwo, accessTokenThree)
+        assertEquals(accessToken, accessTokenThree)
+    }
+
+    /**
+     * Test sign in, get access token. Compare to token from getAccount()
+     */
+    @Test
+    fun testGetAccessTokenFour() = runTest {
+        val correlationId = "d1a75d96-1f96-4919-ac5f-ce2d62d07b68" //UUID.randomUUID().toString()
```
SammyO: Why a hardcoded correlation ID?

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -550,6 +553,117 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+
+        assertEquals(accessTokenTwo, accessTokenThree)
+        assertEquals(accessToken, accessTokenThree)
+    }
+
+    /**
+     * Test sign in, get access token. Compare to token from getAccount()
+     */
+    @Test
+    fun testGetAccessTokenFour() = runTest {
```
SammyO: Can you update the test description and name to make it clear what this test does? Difference between this and `testThree` is not clear to me.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: yes, the scopes are null from the old deprecated method.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
SaurabhMSFT: I have updated the title.

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
SaurabhMSFT: I have removed *NoFixedScopes* classes

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: Lets discuss this on Teams

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: Comments updated

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: done

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: updated

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
SaurabhMSFT: I cannot add more tests for now due to a limitation in the mock API response. I am looking for a way to configure the `iss` claim in the idToken returned by the mock API.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: done

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: That would require more code then present. We will have to create a new method in the `CommandParametersAdapter` class that does not require `AcquireTokenSilentParameters` to be passed.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
SaurabhMSFT: It was for debugging, its fixed now

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -220,21 +236,71 @@ class AccountState private constructor(
     }
 
     /**
-     * Retrieves the access token for the currently signed in account from the cache.
+     * Retrieves the access token for the currently signed in account from the cache. If multiple
```
nilo-ms: could you put a deprecation warning for this method?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -220,21 +236,71 @@ class AccountState private constructor(
...
+    /**
+     * Retrieves the access token for the currently signed in account from the cache.
+     * If the access token is expired, it will be attempted to be refreshed using the refresh token that's stored in the cache;
+     * If any of the cached access token matches the scope specified in scopes parameter then that
+     * is returned otherwise a new access token with requested scopes is fetched.
+     * Kotlin coroutines variant.
+     *
+     * @return [com.microsoft.identity.nativeauth.statemachine.results.GetAccessTokenResult] The result of the getAccessToken action
+     */
+    suspend fun getAccessToken(forceRefresh: Boolean = false, scopes: List<String>?): GetAccessTokenResult {
```
nilo-ms: I think the scopes list should not be optional here

SaurabhMSFT: Making it null allows the developers to move to this function from the old deprecated method.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
SaurabhMSFT: I was able to configure the iss claim parameter but that still does not work because there is an internal logic to match authorities from token response and request to login.microsoftonline are same. Writing tests with mock API seems very difficult. We will have to wait for the labs tenant to automate our tests.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: Done

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -98,6 +100,19 @@ class AccountState private constructor(
         }
     }
 
+    private fun addDefaultScopes(scopes: List<String>?): List<String> {
```
SammyO: Why is this done here, rather than inside the controller? The latter is what MSAL is doing currently, and wherever possible I would like to stick to existing designs/architectures.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -194,14 +209,16 @@ class AccountState private constructor(
     interface GetAccessTokenCallback : Callback<GetAccessTokenResult>
 
     /**
-     * Retrieves the access token for the currently signed in account from the cache.
+     * Retrieves the access token for the currently signed in account from the cache. If multiple
```
SammyO: This comment doesn't make sense, as this method doesn't accept a scopes parameter.
Also, shouldn't this method be returning a token for the default scopes? If so, that should be made clear from the comments.

@nilo-ms  can you review this based on your technical proposal please.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SammyO: @nilo-ms can you confirm what should happen here? Is scopes optional, or not?
- in current MSAL, it's not optional (i.e. if an empty scopes list is passed, the method returns an error)
- if we allow empty/null scopes param, how do we communicate to developers how this method works in those scenarios?
- how is iOS implementing this for native auth? 
- I agree with Saurabh that developers need a way to migrate from the old, now-deprecated method to the new one. If the old one doesn't take a scopes parameter, then the new one needs to have the possibility of also not passing any.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
SammyO: @SaurabhMSFT please make a proposal of how the changes in this PR should be tested. Please align with @borgesmb.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -220,21 +237,72 @@ class AccountState private constructor(
...
+     * If the access token is expired, it will be attempted to be refreshed using the refresh token that's stored in the cache;
+     * If any of the cached access token matches the scope specified in scopes parameter then that
+     * is returned otherwise a new access token with requested scopes is fetched.
+     * callback variant.
+     *
+     * @return [com.microsoft.identity.client.IAuthenticationResult] If successful.
+     * @throws [MsalClientException] If the the account doesn't exist in the cache.
+     * @throws [ServiceException] If the refresh token doesn't exist in the cache/is expired, or the refreshing fails.
+     */
+    fun getAccessToken(forceRefresh: Boolean = false, scopes: List<String>, callback: GetAccessTokenCallback) {
```
SammyO, line 277: Scopes param is not nullable, unlike it's coroutine sibling method. Please align the two.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -245,24 +313,38 @@ class AccountState private constructor(
...
-                forceRefresh,
-                correlationId
+                acquireTokenSilentParameters
             )
 
-            val command = AcquireTokenNoFixedScopesCommand(
+            val command = SilentTokenCommand(
                 params,
-                NativeAuthMsalController(),
+                NativeAuthMsalController().asControllerFactory(),
```
SammyO: This factory is a new pattern. Do we know why MSAL did this, and is this something we should include in native auth?

SaurabhMSFT: Yes, Dome added this last month. We can update our code to use this method.

SammyO: Do we know why they did it? Without that I don't think we can determine whether we should include it.

SaurabhMSFT: This is the PR https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2352 but there is no clear reason in the PR to have the controller return a controllerfactory object.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
nilo-ms: I suggest adding a comment like:
`Retrieves an access token with default OIDC scopes as permission`

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
nilo-ms: I think native auth should work in the same way as the platform specific MSAL works. For example, for MSAL Objective-C, the scope list is not optional, but the external developer can specify an empty list of scopes, no error is thrown. This is how we plan to implement it.
Given that you want to provide a way to migrate to the new method, I suggest to slightly deviate from how MSAL Android work. I propose:
- adding the scope list as mandatory
- If an empty scope list is passed by the external developer, add the default OIDC scopes internally (as we do for the signIn) 
- clarify this behaviour in the method comment

Be mindful that if you want to align with the MSAL Android behaviour, it will be considered a breaking change.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -220,21 +237,72 @@ class AccountState private constructor(
...
+    }
+
+    private suspend fun getAccessTokenInternal(forceRefresh: Boolean, scopes: List<String>?): GetAccessTokenResult {
         LogSession.logMethodCall(
             tag = TAG,
             correlationId = null,
             methodName = "$TAG.getAccessToken(forceRefresh: Boolean)"
         )
+        val mergedScopes = BaseController.addDefaultScopes(
+            scopes?.toMutableSet() ?: mutableSetOf() ).toList()
```
nilo-ms: Keep in mind that converting the list to a hash set will result in the loss of the scopes' order. This should not be an issue due to the way eSTS operates when refreshing an access token for multiple scopes across different audiences. However, if this behavior changes, not maintaining the order of scopes could become problematic.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: I have made the parameter as NonNull

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: This function wasn't used and has been removed.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: Done

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: Done

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: Neither  MSAL  nor us makes any guarantee of the order of scopes passed.

msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
```diff
@@ -136,7 +136,7 @@ public ClaimsRequest getClaimsRequest() {
         return mClaimsRequest;
     }
 
-    void setAccountRecord(AccountRecord record) {
+    public void setAccountRecord(AccountRecord record) {
```
SammyO, line 139: could you update your PRs to include an overview of the changes?

SammyO (changes requested):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SammyO: That's not the question. I can see default scopes are added in the interface layer (in `getAccessTokenInternal()`), rather than inside the controller. Why not the latter, and align with how MSAL's browser-delegated flow operates?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SammyO: @SaurabhMSFT can you clarify which approach you're taking:
1. conceptually, scopes are optional. Similar to how getAccessToken() behaves right now.
2. scopes are not optional. Similar to how MSAL browser-delegated behaves right now.

If we're going with 1, then the following should be implemented IMO:
1. scopes param may be `null`.
2. scopes list may be empty.

As these two are effectively the same, I don't see a reason to limit one but not the other.

msal/src/main/java/com/microsoft/identity/client/TokenParameters.java
SaurabhMSFT: Done

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: I have changed how the `getAccessTokenInternal()` is called from the old deprecated method and then removed the merging of scopes.

SammyO (changes requested):
As discussed on Teams, please propose how these changes should be tested by reviewers, before this is merged in.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
SammyO: Test titles still don't reflect what's being tested. Neither do the comments.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: I have changed the code in getAccessMethod so that empty scopes are not passed to `getAccessTokenInternal` obviating the need to merge scopes.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: Scopes are optional. An empty list of scopes or a null value could show the same intent. Supporting one or the other is just a matter of taste and consistency.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -587,6 +700,92 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
         assertTrue((accessTokenState as GetAccessTokenError).isNoAccountFound())
     }
 
+    /**
+     * Test sign in, sign out, get access token
+     */
+    @Test
```
Yuki-YuXin, line 712: Is this test for the sign out part. It doesn't feel like a good test for the getAccessToken part.

Yuki-YuXin (approved):

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
SaurabhMSFT: This test is a slight variation of the test `testSignOutGetAccessToken()` that calls the `getAccessToken` with two parameters.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SammyO: I disagree @SaurabhMSFT. If scopes are optional, the whole parameter should be nullable. It's a pain for developers to have to create an empty list, only because they can't pas `null`. both `null` and empty list serve the same functional purpose, so both should be allowed.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -220,18 +223,66 @@ class AccountState private constructor(
...
+     * the scope of retrieved access token is a superset of requested scopes. If the access token
+     * has expired, it will be refreshed using the refresh token that's stored in the cache. If no
+     * access token matching the requested scopes is found in cache then a new access token is fetched.
+     * Kotlin coroutines variant.
+     *
+     * @return [com.microsoft.identity.nativeauth.statemachine.results.GetAccessTokenResult] The result of the getAccessToken action
+     */
+    suspend fun getAccessToken(forceRefresh: Boolean = false, scopes: List<String>): GetAccessTokenResult {
+        return getAccessTokenInternal(forceRefresh,
+            if (!scopes.isEmpty()) scopes else AuthenticationConstants.DEFAULT_SCOPES.toList() )
```
SammyO: I've left comments about this a few times before: IMO it shouldn't be the responsibility of the SDK interface to add default scopes.
I can see that you added this because you can't pass `null` to the internal function, but that topic is covered in a different thread.

My proposal:
1. keep `addDefaultScopes()` logic in the controller 
2. allow `null` scopes to be set in the interface.

Unless there is a good reason to do it differently?

SammyO:
@SaurabhMSFT  can you add a change log?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -220,18 +221,67 @@ class AccountState private constructor(
     }
 
     /**
-     * Retrieves the access token for the currently signed in account from the cache.
+     * Retrieves the access token for the default OIDC scopes from the cache.
```
SammyO: Developers won't know what this means. Let's make it more explanatory and developer friendly, please. Applies also to other instances of this comment.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -220,18 +221,67 @@ class AccountState private constructor(
...
+     * Retrieves the access token for the currently signed in account from the cache such that
+     * the scope of retrieved access token is a superset of requested scopes. If the access token
+     * has expired, it will be refreshed using the refresh token that's stored in the cache. If no
+     * access token matching the requested scopes is found in cache then a new access token is fetched.
+     * Kotlin coroutines variant.
+     *
+     * @return [com.microsoft.identity.nativeauth.statemachine.results.GetAccessTokenResult] The result of the getAccessToken action
+     */
+    suspend fun getAccessToken(forceRefresh: Boolean = false, scopes: List<String>): GetAccessTokenResult {
+        if (scopes.isEmpty()) {
```
SammyO, line 257: This check is already done in (Silent)TokenCommandParameters.validate(). Let's not add additional checks in other places; it makes the logic difficult to understand.

SammyO (changes requested):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: Done

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: I have removed this check

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -233,18 +235,64 @@ class AccountState private constructor(
...
+     * the scope of retrieved access token is a superset of requested scopes. If the access token
+     * has expired, it will be refreshed using the refresh token that's stored in the cache. If no
+     * access token matching the requested scopes is found in cache then a new access token is fetched.
+     * callback variant.
+     *
+     * @return [com.microsoft.identity.client.IAuthenticationResult] If successful.
+     * @throws [MsalClientException] If the the account doesn't exist in the cache.
+     * @throws [ServiceException] If the refresh token doesn't exist in the cache/is expired, or the refreshing fails.
+     */
+    fun getAccessToken(forceRefresh: Boolean = false, scopes: List<String>, callback: GetAccessTokenCallback) {
```
nilo-ms, line 277: Are you going to return an error if the scope list is empty?

SaurabhMSFT: yes, an error is returned. It is asserted in one of the tests. I have removed throws from the javadoc

nilo-ms: Cool. It seems that you didn't add a dedicated new error, as iOS did. 
I'm happy adding a new error or using the general one, as long as Android and iOS are aligned. Can you please sync with @borgesmb ?

SaurabhMSFT: I have added a new error type.

SammyO: If we're adding a new error type, we need a new utility function for that as well (similar to e.g. `isInvalidUsername()`).

SammyO: And please include this utility function in your tests.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: This check has been brought back to return `INVALID_SCOPES` error.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: Done.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/GetAccessTokenError.kt
```diff
@@ -27,6 +27,10 @@ import com.microsoft.identity.nativeauth.statemachine.results.GetAccessTokenResu
 
 internal class GetAccessTokenErrorTypes {
     companion object {
+        /*
+         * The NO_ACCOUNT_FOUND value indicates the user is not signed in.
+         * If this occurs, the API should be called after successful sign in
+         */
```
SaurabhMSFT, line 39: `NO_ACCOUNT_RECORD` should have been in the Error.kt file but cannot be removed now without breaking backward compatibility.

SammyO: Why should it have been in Error.kt? AFAIK that class is only for common/shared errors, which this is not right? E.g. we've also got `SignUpErrorTypes`. 
Moreover, all error type classes such as this one should be `internal`, so developers can't access them. They should be using the utility methods, or do raw String inspection if they want to. So AFAIK moving this wouldn't be a breaking change.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
```diff
@@ -77,6 +77,12 @@ internal class ErrorTypes {
          */
         const val INVALID_USERNAME = "invalid_username"
 
+        /*
+        * The INVALID_SCOPES value indicates the scopes provided by the user are not valid
+        * If this occurs, valid scopes should be resubmitted
+        */
+        const val INVALID_SCOPES = "invalid_scopes"
```
SammyO: This is not a shared/common error, right? If so and if you agree with my other comment, please move this out of the common error class and into the one that's specific for this scenario.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -600,6 +695,83 @@ class NativeAuthPublicClientApplicationKotlinTest(private val allowPII: Boolean)
...
+        val accountState = (signInResult as SignInResult.Complete).resultValue
+
+        val signOutResult = accountState.signOut()
+        assertTrue(signOutResult is SignOutResult.Complete)
+
+        var accessTokenState = accountState.getAccessToken(false, emptyList())
+
+        assertTrue(accessTokenState is GetAccessTokenError)
+        assertTrue((accessTokenState as GetAccessTokenError).isInvalidScopes())
+        assertEquals((accessTokenState as GetAccessTokenError).errorType, ErrorTypes.INVALID_SCOPES)
```
SammyO: please update this to use the utility method

SaurabhMSFT: There is a call to utility method on line 770. Are you looking for some other utility method?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
SaurabhMSFT: This is not shared now but should be as we should return similar error from signin method.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
SammyO: Ah sorry, missed that. We don't need line 271, IMO, but nit.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
SammyO: > we should return similar error from signin method

what makes you say that?

SaurabhMSFT: I have moved the variable to `GetAccessTokenErrorTypes` class

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/GetAccessTokenError.kt
SaurabhMSFT: I have moved the variable back to `GetAccessTokenErrorTypes` class

rmccahill (approved):

changelog
```diff
@@ -2,6 +2,7 @@ MSAL Wiki : https://github.com/AzureAD/microsoft-authentication-library-for-andr
 
 vNext
 ----------
+- [PATCH] Support for multiple access tokens in NativeAuth (#2082)
```
SammyO: This should be a MINOR. It's a new feature, and interface changes.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -207,13 +210,11 @@ class AccountState private constructor(
     interface GetAccessTokenCallback : Callback<GetAccessTokenResult>
 
     /**
-     * Retrieves the access token for the currently signed in account from the cache.
+     * Retrieves the access token for the default OIDC(openid, offline_access, profile)  scopes from the cache
```
SammyO: nit, whitespace missing between C and (

SammyO: and nit: double whitespace before "scopes"

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -233,18 +234,70 @@ class AccountState private constructor(
     }
 
     /**
-     * Retrieves the access token for the currently signed in account from the cache.
+     * Retrieves the access token for the default OIDC(openid, offline_access, profile) scopes from the cache.
```
SammyO: whitespace missing

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -207,13 +210,11 @@ class AccountState private constructor(
...
 
     /**
-     * Retrieves the access token for the currently signed in account from the cache.
+     * Retrieves the access token for the default OIDC(openid, offline_access, profile)  scopes from the cache
      * If the access token is expired, it will be attempted to be refreshed using the refresh token that's stored in the cache;
      * callback variant.
      *
      * @return [com.microsoft.identity.client.IAuthenticationResult] If successful.
-     * @throws [MsalClientException] If the the account doesn't exist in the cache.
-     * @throws [ServiceException] If the refresh token doesn't exist in the cache/is expired, or the refreshing fails.
```
SammyO, line 216: Why are these `throws` removed?

SaurabhMSFT: This method doesn't throw exception after PR #2080 . I updated the doc to be inline with the current functionality.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SaurabhMSFT: Done

changelog
SaurabhMSFT: Done

SammyO (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -288,14 +355,14 @@ class AccountState private constructor(
...
-                            correlationId = "UNSET"
+                            correlationId = correlationId
                         )
                     }
 
                     is Exception -> {
                         GetAccessTokenError(
                             exception = commandResult,
-                            correlationId = "UNSET"
+                            correlationId = correlationId
```
SammyO, line 365: We should actually be using `commandResult.correlationid` here, for the unlikely case that the API has returned a different correlation ID. That field might be null, so we'd need to update `GetAccessTokenError`. And that improvement would have to be applied to more places.

SaurabhMSFT: This change can be done when `commandResult is ServiceException` if `commandResult is Exception` then we will use the old correlationId.

SammyO: Feel free to move this into a separate PR. It's not blocking.

PR #2085, opened 2024-04-23
Update Account object in AccountState after access token refresh.

GBakolas:
When the access token is refreshed the SDK gets also a new id token. The updated id token needs to be returned to the application. To avoid changes to the public interface thus having impact to the user of the SDK this solution updates the Account object in the existing AccountSate.


Yuki-YuXin (commented):
According to the discussion in the team channel, are we going to add tests in this PR?

GBakolas:
> According to the discussion in the team channel, are we going to add tests in this PR?
Yes but we will do it in a future PR because we also need to update the Mock API. I have already created a new ADO item for this.

Yuki-YuXin (approved):

SammyO (approved):

PR #2092, opened 2024-04-29
Native auth safe logging test

Yuki-YuXin:
### Comments and Changes Summary:

- Setup external logger and remove it for every test under `@Before ` and `@After`.
- Use Regex rules to filter the log message under `verify().log()`. There are two pools for the regex rules: `sensitivePIIMessages` - items should not exist in the log regardless of allowPII value; `permittedPIIMessages `- items should not exist in the log when `allowPII=false` but can exist when `allowPII=true`
- Use `@ParameterizedRobolectricTestRunner.Parameters` in order to test both `allow=false & allowPII=true` for `RunWith(ParameterizedRobolectricTestRunner::class) NativeAuthPublicClientApplicationKotlinTest(private val allowPII: Boolean)`
- `verifyLogDoesNotContain` checks `message ` and 'containsPII'.  `verifyLogDoesNotContain ` checks `message`.


msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -2291,5 +2394,33 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+
+        if (!allowPII) {
+            elementsToCheck.addAll(piiFalseToCheck)
+        }
+
+        elementsToCheck.forEach { element ->
+            verifyLogDoesNotContain("Command", element)
+        }
+
+        Logger.getInstance().setExternalLogger(null)
```
SammyO: Isn't this the same as what's been done in `cleanup()`?

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -2291,5 +2394,33 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
         val result = application.signUp(invalidUsername)
         assertTrue(result is SignUpError)
         assertTrue((result as SignUpError).isInvalidUsername())
+
+        checkSafeLogging()
+    }
+
+    private fun checkSafeLogging() {
+        val piiTrueToCheck = listOf(
+            "password:","password=",
```
SammyO: let's make this list longer.
- `"password"` (so without : and =)
- `'"password"'` (so looking for the password value that includes " in the string, for example when JSON is outputted it will contain ")
- the above 2 points also for code and username
- attributes is another PII field
- channel target label is another
- add check for access token, ID token and refresh token
- add snake_case as well as camelCase for multi-word values like continuation token.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -2291,5 +2394,33 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+        }
+
+        elementsToCheck.forEach { element ->
+            verifyLogDoesNotContain("Command", element)
+        }
+
+        Logger.getInstance().setExternalLogger(null)
+    }
+    private fun verifyLogDoesNotContain(tag: String, element: String) {
+        verify(externalLogger, never()).log(contains(tag), any(), contains(element), eq(true))
```
SammyO: I think containsPII should be `any()`. Without it, I believe Mockito is looking for something like:
-> assert that the callback is never called with values tag and <any string> and <value of element> and containsPII=true.
whereas what might be happening is that <value of element> is called, but PII=false. If that happens, it won't trigger this I think.

Furthermore, you're actually doing many tests in 1 statement here. While it's efficient from a coding perspective, it may not be logistically. if this test fails, it's hard to determine why, because so many things happen in the same statement.

Consider something like:
- attach a real logger callback interface (make it a Spy, so the method is actually called, but you can also play around with its behaviour the way you can with a Mock)
- within the callback, perform a number of assertions all on different fields and callback configurations, each in a separate statement.

SammyO: I believe we want to do 2 types of checks:
- some values should *never* be in the callback
- some values may be, but only if also containsPII=true (and if the log configuration has allowPII enabled)

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -2291,5 +2394,33 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+        )
+
+        val elementsToCheck = piiTrueToCheck.toMutableList()
+
+        if (!allowPII) {
+            elementsToCheck.addAll(piiFalseToCheck)
+        }
+
+        elementsToCheck.forEach { element ->
+            verifyLogDoesNotContain("Command", element)
```
SammyO: Why do we only care about TAG=Command? I think we should make the tag matcher `any()`, as we don't care what the tag value is in this context.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -2401,12 +2402,18 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
-            "password:","password=",
-            "code:","code=",
-            "continuationToken:","continuationToken="
+            "(?<![,\\[],\")\bpassword\\s*[:=]|\bpassword(?![,\\],\"])", // 'password:'  'password='  exclude ',password' 'password,' '[password' 'password]' '"password' 'password"'
+            "\bcode(?![:=])\b", // 'code' 'code:' 'code=' exclude 'codeLength' 'error_code',
+            "\battributes(?![:=])\b",
+            "(?i)access_token|\baccessToken(?![:=])\b", // access_token, accessToken
+            "(?i)refresh_token|\brefreshToken(?![:=])\b",
+            "(?i)id_token|\bidToken(?![:=])\b",
+            "(?i)client_secret|clientSecret",
```
SammyO: do we need client secret?

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Yes. This is the legacy code. However, setExternalLogger(null) cannot be used as an alternative to new method of removeExternalLogger

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Have updated the rules and scopes.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: No, we currently don't. Removed.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
SammyO: Why not?

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -2291,5 +2396,49 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+        )
+
+        val elementsToCheck = piiTrueToCheck.toMutableList()
+
+        if (!allowPII) {
+            elementsToCheck.addAll(piiFalseToCheck)
+        }
+
+        elementsToCheck.forEach { regex ->
+            verifyLogDoesNotContain("""\S*(Command|Interactor)\S*""", regex)
```
SammyO: why not `any()`?

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Modify from eq(true) to any(). However, the value of allowPII is defined variable of true status.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Narrowing down the tag contributes to easier regex definition. Modify from "Command|Indicator" to any()

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Modify from regex to any()

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin:
```
    public synchronized void setExternalLogger(final ILoggerCallback externalLogger) {
        if (externalLogger == null) {
            return;
        }
```
Because it doesn't look like that's the recommended method.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Updated according to the suggestions.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -101,6 +107,46 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+        ) {
+            if (containsPII) {
+                val elementsToCheck = allowPIITrueToCheck.toMutableList()
+                if (!allowPII) {
+                    elementsToCheck.addAll(allowPIIFalseToCheck)
+                }
+                elementsToCheck.forEach { regex ->
+                    if (RegexMatcher(regex).matches(message)) {
+                        failCalled = true
+                        fail("PII $regex found in log message: $message")
```
SammyO: This is a smart setup, but I'm 100% sure it's going to work. Some thoughts: 
- I see we can't rely on only `fail`, as this callback is executed on a thread that's different from the thread that the test is executed on. I see that's why you've introduced the `failCalled` check, that's smart.
- however, given the fact that `failCalled` is set to true on a different thread, I expect there might be race conditions where the test thread is looking for `spyLoggerCallback.failCalled` (to perform `assertFalse`), before the callback has finished and set the value to `true`. given that the default is `false`, that means the test "thinks" failCalled wasn't set. Whereas in reality it hasn't been set _yet_
- you can try this out by adding a delay to the logger callback (e.g. `Thread.sleep(5000)`). 
- this setup will cause flaky tests: sometimes they'll pass, sometimes they'll fail. They're not deterministic, due to the race condition.
- typically a good way to deal with these async processes in tests is through for example `CountDownLatch`. Take a look at `LoggerTest.java`. Another good way in general when dealing with callbacks vs. coroutines is `suspendCancellableCoroutine` (this might not work here, but is interesting for you to look into).

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -101,6 +107,46 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
     private val password = "verySafePassword".toCharArray()
     private val code = "1234"
     private val emptyString = ""
+    private val allowPII = true
```
SammyO: We will need tests for both `true` and `false`, right? I.e. in theory, every use case (test) should be performed with both PII enabled and disabled. Brute force way would be to duplicate every test (perhaps using a parameterised JUnit setup, e.g. see `CommandResultUtilTest`). Not ideal. Can we think of a smarter way to do this?

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
SammyO: Should (user) attributes be added?

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -101,6 +107,46 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+                elementsToCheck.forEach { regex ->
+                    if (RegexMatcher(regex).matches(message)) {
+                        failCalled = true
+                        fail("PII $regex found in log message: $message")
+                    }
+                }
+            }
+        }
+    }
+    private val spyLoggerCallback = spy(loggerCallback)
```
SammyO: it's actually not necessary to make this a `spy` I think.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Thanks a million for this valuable guidance. Have added CountDownLatch into the loggercallback. But I just make changes for scenario 1 for reviewing first.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -2293,3 +2395,33 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+
+class LoggerTestCallback(private val allowPIITrueToCheck: List<String>, private val allowPIIFalseToCheck: List<String>, private val allowPII: Boolean, private val latch: CountDownLatch?) : ILoggerCallback {
+    var failCalled = false
+    override fun log(
+        tag: String?,
+        logLevel: Logger.LogLevel?,
+        message: String?,
+        containsPII: Boolean
+    ) {
+        if (containsPII) {
```
rmccahill: Not too familiar with how the latch works - but if `containsPII` is false here, will the test hang since `latch?.countDown()` is never called?

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -181,8 +214,19 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
 
+        // check the safe logging
+        val latch = CountDownLatch(1)
+        val loggerCallback = LoggerTestCallback(allowPIITrueToCheck, allowPIIFalseToCheck, allowPII, latch)
+        Logger.getInstance().setExternalLogger(loggerCallback)
+
         val result = application.signIn(username, password)
+
+        withContext(Dispatchers.IO) {
+            latch.await()
```
SammyO: Why is this happening on a coroutine? Unless there's good reason to, I wouldn't be mixing coroutine with callbacks. But moreover, I don't see the reason for this `await()` to happen on a different thread. The whole point is that the main thread waits.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
SammyO: @Yuki-YuXin  this code is not complete yet. As discussed offline and in a different comment I left, we need multiple test cases that validate different values of `containsPII`. E.g. some fields should never be called, regardless of the value of containsPII.

I also think @rmccahill has a point with the above. the countdown should be called outside of the if, i.e. always when the callback is called. 

@rmccahill  the test won't hang, but it will timeout. CountDownLatch has a timeout configured, meaning that if countdown isn't called within that timeframe, the test times out.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -2293,3 +2395,33 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+                elementsToCheck.addAll(allowPIIFalseToCheck)
+            }
+            elementsToCheck.forEach { regex ->
+                if (RegexMatcher(regex).matches(message)) {
+                    failCalled = true
+                    fail("PII $regex found in log message: $message")
+                }
+            }
+            latch?.countDown()
+        }
```
rmccahill: There's a lot of flags here which make the logic here difficult to follow - I'll add a few assumptions on the code here, correct me if I'm wrong:

* If `containsPII` is false, we should return without checking anything (this could also be debated - should we check the logs even if the function indicates there's no PII? Might help to catch some cases where it's been set incorrectly)
* `allowPIITrueToCheck` is a list of regular expressions that will detect any PII messages that might be present in the logs. If this is the case, what is `allowPIIFalseToCheck` for?

With those assumptions, the following case seems incorrect to me: 
* When `allowPII` is true, the `elementsToCheck` will contain the contents of `allowPIITrueToCheck`. This will mean that if a message contains PII, the regex matcher will return true on a message with PII, and `failCalled` will be set to true. This seems incorrect to me from how the test is laid out. If `allowPII` is true, we shouldn't fail the test, even if the message contain PII.

Could we try and simplify it to only check the messages if `allowPII` is set to false? Is there value in checking log contents when `allowPII` is true?

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Use Parameterized for this.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: I first wrongly place the latch?.countDown() which has been fixed now with more time for withTimeOut {}

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: You are right. There is a logical misunderstanding here.

I started out using containPII as a criterion.  If the object doesn't have PII info then the entry is not checked. We only check if we have put invalid attributes for entries that have PII info. However, object.containPII is determined by our log, toString().equals(toUnsanitizedString()). This means that the value of containPII should be used as part of our logging check and not a basis for decision-making.

Yuki-YuXin: I rename allowPIITrueToCheck as infoPIIToCheck. infoPIIToCheck items should never be logged even when allowPII is true. These are the log contenct we need to check when `allowPII=true`.  
allowPIIFalseToCheck items are logged or not depending on allowPII. They are allowed to be logged when  `allowPII=true`. However, under  `allowPII=false` they shouldn't.

If we think allowPIITrueToCheck as infoPIIToCheck. If a message contains PII, the regex matcher will return true, fail() and failCalled=true will be used. Because infoPIIToCheck items should not exists in both allowPII=true&false, thus the test should fail.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Remove the  withContext(Dispatchers.IO) here instead of 
```kotlin
val shouldLogBeDiscarded = false
        val timedOut: Boolean = !latch.await(DISCARDED_TIME_IN_MILLISECONDS, TimeUnit.MILLISECONDS)
        assertEquals(shouldLogBeDiscarded, timedOut)
```
Similar like LoggerTest.java

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -2293,3 +2408,48 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+    var failCalled = false
+    var count = 0  // Used to know how many logs need to be checked
+    override fun log(
+        tag: String?,
+        logLevel: Logger.LogLevel?,
+        message: String?,
+        containsPII: Boolean
+    ) {
+        if (allowPII) {
+            allowPIIFalseToCheck.forEach { regex ->
```
Yuki-YuXin: This part might be not necessary because LoggerTest takes the responsibility of validating containsPII.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: And the logic here has been updated according to the offline discussion.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -101,10 +112,31 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
     private val password = "verySafePassword".toCharArray()
     private val code = "1234"
     private val emptyString = ""
+    private val infoPIIToCheck = listOf(
```
SammyO: nit: why is this variable called `info`?

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -101,10 +112,31 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+        """(?<![\(])attributes[:=]""",
+        """(?i)\b(accessToken|access_token)[:=]""", // access_token, accessToken
+        """(?i)\b(refreshToken|refresh_token)[:=]""",
+        """(?i)\b(idToken|id_token)[:=]""",
+        """(?i)\b(continuation_token)[:=]"""
+    )
+    private val allowPIIFalseToCheck = listOf(
+        """(?<![\(])username[:=]""",
+        """(?i)\b(challengeTargetLabel|challenge_target_label)[:=]""",
+        """(?i)\b(grantType|grant_type)[:=]"""
```
SammyO: IMO grant type is not PII. It's a developer setting, just like authority URL.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -181,8 +221,21 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
             MockApiResponseType.TOKEN_SUCCESS
         )
 
+        val loggingNumber = if (allowPII) 137 else 127
```
SammyO: What is this used for? Did you manually count the amount of logs that are produced in a certain flow, and you're counting them down? If so, let's think of an alternative strategy as this is not scalable. This test would need to be updated very frequently, and the means to get to this count (i.e. manually counting logs) is time consuming.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
SammyO: Not exactly. LoggerTest validates whether, if a native auth object states that it contains PII (through `ILoggable`), the callback's `containsPII` is set to `true` (and vice versa). But there are more ways through which `containsPII` can be set. Moreover: these tests validate whether `containsPII` is correctly set according to actual fields (like username, attributes, etc.), and they're run on a wide range of scenarios. This is more of an E2E test, whereas `LoggerTest` is a unit test that tests the `ILoggable` interface specifically.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -181,8 +221,21 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
...
+
         val result = application.signIn(username, password)
         assertTrue(result is SignInResult.Complete)
+
+        // check the safe logging
+        println(loggerCallback.count)
+
+        val shouldLogBeDiscarded = false
+        val timedOut: Boolean = !latch.await(DISCARDED_TIME_IN_MILLISECONDS, TimeUnit.MILLISECONDS)
+        assertEquals(shouldLogBeDiscarded, timedOut)
```
rmccahill: you could probably simplify this to `assertFalse(timedOut)`

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Yes, counting first then counting down then. Have updated according to the offline discussion.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
rmccahill: Understood 👍 the updated logic looks good to me, very straightforward to read.

Small suggestion on the naming which might make the intention clearer:
`infoPIIToCheck` -> `sensitivePIIMessages`
`allowPIIFalseToCheck` -> `permittedPIIMessages`

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
```diff
@@ -221,8 +274,14 @@ class NativeAuthPublicClientApplicationKotlinTest : PublicClientApplicationAbstr
             MockApiResponseType.INVALID_PASSWORD
         )
 
+        val loggingNumber = if (allowPII) 114 else 114
```
rmccahill: The number is the same regardless of the condition here

I'd also follow up on Sammy's feedback - the check is quite fragile. We should only be checking if the PII logging rules are valid.

I'm assuming what we want to check is: when `allowPII` is true, that messages containing PII on the allowed list are logged, and when `allowPII` is false, those messages are not logged.

In which case, you could add a boolean to the logger called something like `permittedPIIFound`. Since you're checking the `allowPIIFalseToCheck` list for permitted PII messages, you could set `permittedPIIFound` when a permitted message is logged. Then you could do `assertEqual(allowPII, permittedPIIFound)`. There's the possibility of this approach incorrectly failing if a flow doesn't log any messages in the permitted PII list, but we can assess if it works for the provided tests and update accordingly.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Short for information. Renamed as sensitivePIIMessages as Robert suggested.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Add verifyLogCouldContain function to validate whether containsPII is correctly set according to actual fields.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Nice suggestions. Updated.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationKotlinTest.kt
Yuki-YuXin: Remove the logger callback + countdown mechanism because they are fragile and not scalable.

The assumption is right.  What we are using right now is verify().log() rather than assert(variable) or fail(). Based on your suggestions, I added allowList & disableList 
```
verify(externalLogger, never()).log(
            any(),
            any(),
            argThat(RegexMatcher(regex)),  // allowList items are logged but the containsPII should be true.
            eq(false)
        )
```

Yuki-YuXin:
Shelve and place because pending proposal decision.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
```diff
@@ -120,6 +138,30 @@ public class NativeAuthPublicClientApplicationJavaTest extends PublicClientAppli
     private final char[] password = "verySafePassword".toCharArray();
     private final String code = "1234";
     private final String emptyString = "";
+    private final boolean allowPII;
+
+    public NativeAuthPublicClientApplicationJavaTest(boolean allowPII) {
+        this.allowPII = allowPII;
+    }
+
+    private final List<String> sensitivePIIMessages = Arrays.asList(
```
SammyO: can we add one more item here? A regex for a JWT, which would cover access tokens and ID tokens. For that regex you can omit the `=` and `:` parts.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
```diff
@@ -2832,6 +2974,38 @@ public void testSignUpInvalidEmailReturnsError() throws ExecutionException, Inte
 
         assertTrue(signUpResult instanceof SignUpError);
         assertTrue(((SignUpError) signUpResult).isInvalidUsername());
+
+        checkSafeLogging();
+    }
+
+    private void checkSafeLogging() {
```
SammyO: Since this is used in both tests, and potentially more tests moving forward: can this be extracted into a dedicated component, so that reusability becomes more straight forward? Along with the `permittedPIIMessages` and `sensitivePIIMessages` lists. So that we have 1 source of truth.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
```diff
@@ -2832,6 +2974,38 @@ public void testSignUpInvalidEmailReturnsError() throws ExecutionException, Inte
...
+        if (allowPII) {
+            allowList = new ArrayList<>(permittedPIIMessages);
+            disableList = new ArrayList<>(sensitivePIIMessages);
+        } else {
+            allowList = new ArrayList<>();
+            disableList = new ArrayList<>(sensitivePIIMessages);
+            disableList.addAll(permittedPIIMessages);
+        }
+
+        for (String regex : allowList) {
```
SammyO: nice and clean 👍

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
Yuki-YuXin: Create a component called as LoggerCheckHelper which has all logger related stuff inside.

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
Yuki-YuXin: Done. a JWT token or similar structure.`"""^[A-Za-z0-9_-]+\\.[A-Za-z0-9_-]+\\.[A-Za-z0-9_-]+$""" // JWT token`

rmccahill (approved):

azure-pipelines/pull-request-validation/pr-msal.yml
```diff
@@ -66,7 +67,7 @@ jobs:
   - task: Gradle@2
     displayName: Run Unit tests
     inputs:
-      tasks: msal:testLocalDebugUnitTest -Plabtest -PlabSecret=$(LabVaultAppSecret) -ProbolectricSdkVersion=${{variables.robolectricSdkVersion}}
+      tasks: msal:testLocalDebugUnitTest -Plabtest -PlabSecret=$(LabVaultAppCert) -ProbolectricSdkVersion=${{variables.robolectricSdkVersion}}
```
SammyO: Are you sure about this change?

Yuki-YuXin: There is opening PR (and has not merged) from MSAL to update the expired secret. I include their changes to see if that's the reason for the failed pipeline verification. I personally think we need to wait that PR to be merged in order to merge our work.

SammyO: Agree. Let's not introduce changes with this PR that don't belong here. Let's mark this PR as "draft" in the meantime.

SammyO (approved):

PR #2096, opened 2024-05-08
Added blockstore dependency

somalaya:
Related broker PR : https://github.com/AzureAD/ad-accounts-for-android/pull/2790

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -77,6 +77,7 @@ ext {
     jcipAnnotationVersion = "1.0-1"
     openTelemetryVersion = "1.18.0"
     jetpackDataStoreVersion = "1.0.0"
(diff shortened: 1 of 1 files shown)


gradle/versions.gradle
```diff
@@ -77,6 +77,7 @@ ext {
     jcipAnnotationVersion = "1.0-1"
     openTelemetryVersion = "1.18.0"
     jetpackDataStoreVersion = "1.0.0"
+    blockstoreVersion="16.2.0"
```
mohitc1, line 80: Do we need this in MSAL too?

somalaya: We usually add all the dependencies versions in all the versions.gradle files afaik.

mohitc1 (approved):

rpdome (approved):

PR #2107, opened 2024-05-23
Update publishing tasks

melissaahn:
### Summary
(still in progress)
This PR:
- Adds the suggested way to automatically create sources and javadocs jar files (https://developer.android.com/build/publish-library/configure-pub-variants#single-pub-var and https://developer.android.com/reference/tools/gradle-api/8.3/null/com/android/build/api/dsl/LibraryPublishing).
- Removes tasks related to generating javadocs and pom files.

Pipeline run: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1304918&view=results&j=76b33e1f-4828-5f2f-d07d-022319483d4a

Artifacts created on a test pipeline run: https://identitydivision.visualstudio.com/Engineering/_artifacts/feed/AndroidADAL/maven/com.microsoft.identity.client%2Fmsal/overview/1.0.20240523-melissaahn-PublishingChanges

Note the names of the artifacts and that most of the dependencies are listed under Runtime now instead of Compile. This means that our external dependencies which we declare with `implementation` configuration are no longer leaking to our consumers. The configurations are now working as described [here](https://developer.android.com/build/dependencies#dependency_configurations).
Consumers who have relied on the libraries for external dependencies will now need to add the dependencies directly.


rpdome (approved):

fadidurah (approved):

melissaahn:
This PR got replaced with this one: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2111

PR #2111, opened 2024-05-31
Updated Pom task

melissaahn:
### Summary
Updating the Pom syntax, which will also denote the scope of each dependency appropriately.
Note the names of the artifacts and that most of the dependencies are listed under Runtime now instead of Compile. This means that our external dependencies which we declare with implementation configuration are no longer leaking to our consumers. The configurations are now working as described [here](https://developer.android.com/build/dependencies#dependency_configurations).


shahzaibj (approved):

msal/build.gradle
```diff
@@ -327,61 +327,33 @@ publishing {
...
+                        name = 'MIT License'
                     }
                 }
-
+                inceptionYear = '2016'
+                scm {
+                    url = 'https://github.com/AzureAD/microsoft-authentication-library-for-android/tree/master'
+                }
+                properties = [
+                        branch : 'master',
```
iamgusain: I think we can change the branch name to dev, since we don't use master anymore

iamgusain (approved):

msal/build.gradle
melissaahn: Changed to dev

PR #2121, opened 2024-06-13
Multiple token E2E tests

SammyO:
- Automate manual E2E tests for token management. See [link](https://loop.cloud.microsoft/p/eyJ1IjoiaHR0cHM6Ly9taWNyb3NvZnQuc2hhcmVwb2ludC5jb20vc2l0ZXMvNjc4ZjJkOTEtM2RhMy00M2JlLThjYjUtYjFlOWEyODcyYzA2P25hdj1jejBsTWtaemFYUmxjeVV5UmpZM09HWXlaRGt4TFROa1lUTXRORE5pWlMwNFkySTFMV0l4WlRsaE1qZzNNbU13Tmlaa1BXSWxNakZoUkdFM1RGcEtaa1JWUjI1bE5uUjFOV2hRVDBvMmVrVTNiemhLTFV0R1RXOHdSa3RUV0dKWWRVOWhhSGRhTm1ad2FIQm5VV0l0WXpocVkzUjZjWFZMSm1ZOU1ERlJRVXhaUjFCVlUwcFpRMHhSUjFSWk0wNUJNemRNTkVwU1dWazJOVXBTV0NaalBTVXlSaVpoUFV4dmIzQkJjSEFtY0QwbE5EQm1iSFZwWkhnbE1rWnNiMjl3TFhCaFoyVXRZMjl1ZEdGcGJtVnlKbmc5SlRkQ0pUSXlkeVV5TWlVelFTVXlNbFF3VWxSVlNIaDBZVmRPZVdJelRuWmFibEYxWXpKb2FHTnRWbmRpTW14MVpFTTFhbUl5TVRoWmFVWm9Va2RGTTFSR2NFdGFhMUpXVWpJMWJFNXVVakZPVjJoUlZEQnZNbVZyVlROaWVtaExURlYwUjFSWE9IZFNhM1JVVjBkS1dXUlZPV2hoU0dSaFRtMWFkMkZJUW01VlYwbDBXWHBvY1ZrelVqWmpXRlpNWmtSQmVGVlZSazFYVldSUlZucGtRbFpGZHpCVFJsSk5VbnBrVDFGc2NFTlNWbHBFVVRBeFRGUnNXa3hWUldNbE0wUWxNaklsTWtNbE1qSnBKVEl5SlROQkpUSXlaVGN5TW1OaFkyTXRaakJpTmkwME1qZzFMVGt4TVdJdFlqQTJNbU13TWpOa016RmxKVEl5SlRkRSJ9)
- Add custom Assert for asserting states (with useful message if assert fails)

MSAL common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2432

### DIFF
--- a/msal/build.gradle
+++ b/msal/build.gradle
@@ -240,7 +240,6 @@ dependencies {
     }
 
     testImplementation "io.mockk:mockk:$rootProject.ext.mockkVersion"
-    testImplementation "org.projectlombok:lombok:$rootProject.ext.lombokVersion"
 
     testImplementation "org.powermock:powermock-module-junit4:$rootProject.ext.powerMockVersion"
     testImplementation "org.powermock:powermock-module-junit4-rule:$rootProject.ext.powerMockVersion"
--- a/msal/build.gradle
+++ b/msal/build.gradle
@@ -240,7 +240,8 @@ dependencies {
     }
 
     testImplementation "io.mockk:mockk:$rootProject.ext.mockkVersion"
-    
+    testImplementation "org.projectlombok:lombok:$rootProject.ext.lombokVersion"
+
     testImplementation "org.powermock:powermock-module-junit4:$rootProject.ext.powerMockVersion"
     testImplementation "org.powermock:powermock-module-junit4-rule:$rootProject.ext.powerMockVersion"
     testImplementation "org.powermock:powermock-api-mockito2:$rootProject.ext.powerMockVersion"
--- a/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
+++ b/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
@@ -68,6 +68,7 @@ import java.util.UUID
  *  AccountState returned as part of a successful completion of sign in flow [com.microsoft.identity.nativeauth.statemachine.results.SignInResult.Complete].
  */
 class AccountState private constructor(
+    internal var authenticationResult: IAuthenticationResult?,
     private var account: IAccount,
     private val config: NativeAuthPublicClientApplicationConfiguration,
     val correlationId: String
@@ -75,7 +76,8 @@ class AccountState private constructor(
 
     interface SignOutCallback : Callback<SignOutResult>
 
-    constructor(parcel: Parcel) : this(
+    constructor (parcel: Parcel) : this (
+        authenticationResult = null, // We don't need authenticationResult transferred from one Fragment to another.
         account = parcel.serializable<IAccount>() as IAccount,
         correlationId = parcel.readString() ?: "UNSET",
         config = parcel.serializable<NativeAuthPublicClientApplicationConfiguration>() as NativeAuthPublicClientApplicationConfiguration
@@ -309,10 +311,12 @@ class AccountState private constructor(
                             correlationId = correlationId
                         )
 
+                val privateCorrelationId = if (correlationId == "UNSET") { UUID.randomUUID().toString() } else { correlationId }
+
                 val acquireTokenSilentParameters = AcquireTokenSilentParameters.Builder()
                     .forAccount(currentAccount)
                     .fromAuthority(currentAccount.authority)
-                    .withCorrelationId(UUID.fromString(correlationId))
+                    .withCorrelationId(UUID.fromString(privateCorrelationId))
                     .forceRefresh(forceRefresh)
                     .withScopes(scopes)
                     .build()
@@ -413,6 +417,7 @@ class AccountState private constructor(
             config: NativeAuthPublicClientApplicationConfiguration
         ): AccountState {
             return AccountState(
+                authenticationResult = authenticationResult,
                 account = authenticationResult.account,
                 correlationId = correlationId,
                 config = config
@@ -425,6 +430,7 @@ class AccountState private constructor(
             config: NativeAuthPublicClientApplicationConfiguration
         ): AccountState {
             return AccountState(
+                authenticationResult = null,
                 account = account,
                 correlationId = correlationId,
                 config = config
--- a/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
+++ b/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
@@ -68,7 +68,6 @@ import java.util.UUID
  *  AccountState returned as part of a successful completion of sign in flow [com.microsoft.identity.nativeauth.statemachine.results.SignInResult.Complete].
  */
 class AccountState private constructor(
-    internal var authenticationResult: IAuthenticationResult?,
     private var account: IAccount,
     private val config: NativeAuthPublicClientApplicationConfiguration,
     val correlationId: String
@@ -77,7 +76,6 @@ class AccountState private constructor(
     interface SignOutCallback : Callback<SignOutResult>
 
     constructor (parcel: Parcel) : this (
-        authenticationResult = null, // We don't need authenticationResult transferred from one Fragment to another.
         account = parcel.serializable<IAccount>() as IAccount,
         correlationId = parcel.readString() ?: "UNSET",
         config = parcel.serializable<NativeAuthPublicClientApplicationConfiguration>() as NativeAuthPublicClientApplicationConfiguration
@@ -417,7 +415,6 @@ class AccountState private constructor(
             config: NativeAuthPublicClientApplicationConfiguration
         ): AccountState {
             return AccountState(
-                authenticationResult = authenticationResult,
                 account = authenticationResult.account,
                 correlationId = correlationId,
                 config = config
@@ -430,7 +427,6 @@ class AccountState private constructor(
             config: NativeAuthPublicClientApplicationConfiguration
         ): AccountState {
             return AccountState(
-                authenticationResult = null,
                 account = account,
                 correlationId = correlationId,
                 config = config
--- a/msal/src/test/java/com/microsoft/identity/client/e2e/shadows/ShadowBaseController.kt
+++ b/msal/src/test/java/com/microsoft/identity/client/e2e/shadows/ShadowBaseController.kt
@@ -1,0 +1,74 @@
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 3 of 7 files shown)


msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/GetAccessTokenTests.kt
```diff
@@ -0,0 +1,371 @@
...
+
+        // Token should be retrieved from API, as the scope belongs to a different API resource
+        val getAccessTokenResult2 = accountState.getAccessToken(forceRefresh = false, scopes = listOf(CUSTOMERS_READ_ALL_SCOPE))
+        assertState(getAccessTokenResult2, GetAccessTokenResult.Complete::class.java)
+        Assert.assertTrue(wasRenewAccessTokenInvoked)
+        val authResult2 = (getAccessTokenResult2 as GetAccessTokenResult.Complete).resultValue
+        Assert.assertTrue(authResult2.scope.contains(CUSTOMERS_READ_ALL_SCOPE))
+        val tokenWithCustomerScope = authResult2.accessToken
+        Assert.assertNotEquals(tokenWithEmployeeScope, tokenWithCustomerScope)
+    }
```
rmccahill, line 214: I can't see a difference between this test and the previous one?

msal/src/test/java/com/microsoft/identity/client/e2e/utils/AssertUtils.kt
```diff
@@ -0,0 +1,32 @@
...
+//  THE SOFTWARE.
+package com.microsoft.identity.client.e2e.utils
+
+import org.junit.Assert
+
+fun assertState(actual: Any, expected: Class<*>) {
+    val condition = expected.isInstance(actual)
+    if (!condition) {
+        Assert.fail("Type comparison failed. Expected: $expected, actual: ${actual.javaClass}")
+    }
```
rmccahill, line 37: Is there value using this over something like `Assert.assertTrue(state is ExampleState)`? I find the above usages more difficult to follow

msal/src/test/java/com/microsoft/identity/client/e2e/shadows/ShadowBaseController.kt
```diff
@@ -0,0 +1,74 @@
...
+                OAuth2Strategy::class.java, strategy
+            ),
+            ClassParameter(ICacheRecord::class.java, cacheRecord)
+        )
+    }
+
+
+}
+
+interface OnRenewAccessTokenInvoked {
```
rmccahill: Is this interface used anywhere?

msal/src/test/java/com/microsoft/identity/client/e2e/shadows/ShadowBaseController.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    @Implementation
+    fun renewAccessToken(
+        parameters: SilentTokenCommandParameters,
+        acquireTokenSilentResult: AcquireTokenResult,
+        tokenCache: OAuth2TokenCache<*, *, *>,
+        strategy: OAuth2Strategy<*, *, *, *, *, *, *, *, *, *, *, *, *>,
+        cacheRecord: ICacheRecord
+    ) {
+        onRenewAccessTokenInvokedCallback.invoke()
```
rmccahill, line 55: Suggestion: Could it be cleaner to instead have a boolean variable `renewAccessTokenCalled` on this shadow controller instead of having to explicitly declare a callback, since all the callback is used for is resetting a local variable in the tests?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/GetAccessTokenTests.kt
```diff
@@ -0,0 +1,371 @@
...
+import org.junit.Test
+import org.robolectric.annotation.Config
+
+@Config(shadows = [ShadowBaseController::class])
+class GetAccessTokenTests : NativeAuthPublicClientApplicationAbstractTest() {
+    companion object {
+        const val EMPLOYEE_WRITE_ALL_SCOPE = "api://1e9e882d-3f7d-4b02-af06-b5db4d8466c0/Employees.Write.All"
+        const val EMPLOYEE_READ_ALL_SCOPE = "api://1e9e882d-3f7d-4b02-af06-b5db4d8466c0/Employees.Read.All"
+        const val CUSTOMERS_WRITE_ALL_SCOPE = "api://a6568f2f-47a5-4b18-b2c7-25eff03d87d6/Customers.Write.All"
+        const val CUSTOMERS_READ_ALL_SCOPE = "api://a6568f2f-47a5-4b18-b2c7-25eff03d87d6/Customers.Read.All"
```
rmccahill: Considering these scopes are tied to application IDs, how difficult would it be to extract them out from the code here in a similar way to what we are doing with usernames + passwords for the test accounts? Think it would be easier to maintain if we can configure these elsewhere, otherwise we'd have to do a pull request if they need to be changed in future.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -309,10 +309,12 @@ class AccountState private constructor(
                             correlationId = correlationId
                         )
 
+                val privateCorrelationId = if (correlationId == "UNSET") { UUID.randomUUID().toString() } else { correlationId }
+
                 val acquireTokenSilentParameters = AcquireTokenSilentParameters.Builder()
                     .forAccount(currentAccount)
                     .fromAuthority(currentAccount.authority)
-                    .withCorrelationId(UUID.fromString(correlationId))
+                    .withCorrelationId(UUID.fromString(privateCorrelationId))
```
Yuki-YuXin, line 317: Is that because we don't want to display "UNSET" to the developers?

msal/src/test/java/com/microsoft/identity/client/e2e/utils/AssertUtils.kt
SammyO: assertTrue only validates if `state` is of type `ExampleState`. It doesn't, when the assert fails, tell us what the state was (if not `ExampleState`).

msal/src/test/java/com/microsoft/identity/client/e2e/shadows/ShadowBaseController.kt
SammyO: Failing the test from within `ShadowBaseController` isn't right: it's the responsibility of the test to assert and fail. 
I could replace the callback with a static boolean that's kept inside `ShadowBaseController`, where the test verifies the value and fails the test if the value is not as expected. `ShadowBaseController` becomes a bit more than just a shadow then; it's used to manage/track state as well. Not sure if that belongs there.
Wdyt? do you see another solution?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/GetAccessTokenTests.kt
SammyO: I didn't think env vars were necessary as I don't see this as sensitive information, but I agree re. the maintainability part. Let me refactor this.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/GetAccessTokenTests.kt
SammyO: The naming could be better, but the previous test signs in with custom scope 1, then retrieves token for that same scope from cache, and then retrieves a token for custom scope 2 from API. This test signs in with no scopes, then retrieves token for scope 1 from API, and then token for scope 2 from API. Let me add some comments to the tests to explain the steps, for readability.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/GetAccessTokenTests.kt
SammyO: I'll be linking up with @Yuki-YuXin  re. similar thread in her PR https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2123/files#r1639805129

msal/src/test/java/com/microsoft/identity/client/e2e/utils/AssertUtils.kt
rmccahill: Ah, I can see how that would be useful 👍

msal/src/test/java/com/microsoft/identity/client/e2e/shadows/ShadowBaseController.kt
rmccahill: To clarify - I don't think the test should be failed from the controller. Just that it should hold the boolean reference, sort of like a spy would. 

Although I think I misunderstood how this shadow works - is this essentially being called when any instance of any base controller calls this method, instead of referring to a single base controller instance? If so, then I think your approach makes more sense.

SammyO: Correct. We don't know which instance if BaseController is used, as this is an E2E test that can't see the internal controller etc. initialisation. So can't use a real spy - for that you need the instance. 
Robolectric's Shadow works through reflection. Which I consider a bit of a code smell, but the alternative was to expose the original `authResult` as part of the sign in result (see first commit). Which is even more of a code smell. 
I'm open to further suggestions though!

rmccahill: Since we're hitting a real tenant now, can we compare the tokens between each call to see if they've been refreshed?

SammyO: yeah, that's what we're doing in the tests right now. But since signIn() doesn't return a token, I can't compare the result of signIn() with subsequent getAccessToken() calls.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SammyO: No, because of https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2958930.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/GetAccessTokenTests.kt
```diff
@@ -0,0 +1,420 @@
...
+import com.microsoft.identity.nativeauth.statemachine.results.SignInResult
+import kotlinx.coroutines.test.runTest
+import org.junit.Assert
+import org.junit.Test
+import org.robolectric.annotation.Config
+
+@Config(shadows = [ShadowBaseController::class])
+class GetAccessTokenTests : NativeAuthPublicClientApplicationAbstractTest() {
+    companion object {
+        const val EMPLOYEE_WRITE_ALL_SCOPE = "api://1e9e882d-3f7d-4b02-af06-b5db4d8466c0/Employees.Write.All"
```
nilo-ms, line 40: Shouldn't these scopes be retrieved from an env variable as you're currently doing for "NATIVE_AUTH_LABS_EMAIL_PW_APP_ID". I'm saying this because these scopes contain the app id, so we're exposing publicly the clientID. Feel free to resolve this comment if client ID isn't a sensitive information

msal/build.gradle
```diff
@@ -240,7 +240,8 @@ dependencies {
     }
 
     testImplementation "io.mockk:mockk:$rootProject.ext.mockkVersion"
-    
+    testImplementation "org.projectlombok:lombok:$rootProject.ext.lombokVersion"
```
nilo-ms: nit: I couldn't locate the class where you used a lombok class in this PR.

Yuki-YuXin (approved):

rmccahill (approved):

fadidurah (approved):

PR #2123, opened 2024-06-14
First batch of Native Auth E2E test on sign in

Yuki-YuXin:
**SignIn username + code:**
- User is not registered with given email (use case 2.2.2) - Test case 31 ✔
- Invalid OTP (use case 2.2.7) - Test case 35 ✔
- Use email and OTP to get token and sign in (hero scenario 6, use case 2.2.1) - Test case 30 ✔

**SignIn username + password:**
- Password is incorrect (1.2.3) - Test case 39 ✔
- Use email and password to get token (hero scenario 15, use case 1.2.1) - test case 37 ✔
- User is not registered with given email (use case 1.2.2) - test case 38 ✔


msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInTest.kt
```diff
@@ -43,11 +114,59 @@ class SignInTest : NativeAuthPublicClientApplicationAbstractTest() {
...
+
+        retryOperation {
+            runBlocking {
+                val user = tempEmailApi.generateRandomEmailAddress()
+                signInResult = application.signIn(user)
+                Assert.assertTrue(signInResult is SignInResult.CodeRequired)
+                otp = tempEmailApi.retrieveCodeFromInbox(user)
+                val submitCodeResult = (signInResult as SignInResult.CodeRequired).nextState.submitCode(otp)
+                Assert.assertTrue(submitCodeResult is SignInResult.Complete)
+                "Success"
```
rmccahill: Is there a reason for having this string here?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInTest.kt
```diff
@@ -24,16 +24,87 @@
...
+    // Remove default Coroutine test timeout of 10 seconds.
+    private val testDispatcher = StandardTestDispatcher()
+
+    @Before
+    override fun setup() {
+        super.setup()
+        Dispatchers.setMain(testDispatcher)
+    }
+
+    private fun <T> retryOperation(
```
rmccahill: Looks like this is used in a few places - can we move it to `NativeAuthPublicClientApplicationAbstractTest`?

Yuki-YuXin: That's a good point. I will move it to NativeAuthPublicClientApplicationAbstractTest after validating it works in the pipeline.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInTest.kt
Yuki-YuXin: There doesn't seem to be a necessary reason, thus removed.

rmccahill (approved):

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInTest.kt
```diff
@@ -43,11 +91,57 @@ class SignInTest : NativeAuthPublicClientApplicationAbstractTest() {
...
-        val password = getSafePassword()
-        val result = application.signIn(username, password.toCharArray())
-        Assert.assertTrue(result is SignInResult.Complete)
+    fun testSuccessEmailOTP() {
+        var signInResult: SignInResult
+        var otp: String
+
+        retryOperation {
+            runBlocking {
+                val user = NativeAuthCredentialHelper.nativeAuthSignInUsername
```
SammyO: wait, how is this passing? The existing user is a password user, so they shouldn't be getting "CodeRequired". AFAIK in order to support email + OTP we'll need:
- a new user
- new env vars and BuildConfig support

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInTest.kt
```diff
@@ -43,11 +91,57 @@ class SignInTest : NativeAuthPublicClientApplicationAbstractTest() {
...
+                Assert.assertTrue(submitCodeResult is SignInResult.Complete)
+            }
+        }
+    }
+
+    /**
+     * Use email and OTP to get token while user is not registered with given email (use case 2.2.2) - Test case 31
+     */
+    @Test
+    fun testErrorEmailOTPIsUserNotFound() = runTest {
```
SammyO: Isn't this actually the same as `testErrorEmailPasswordIsUserNotFound`? Or are we trying to test different auth flows (password vs. OTP), through different challenge types? If so, this test should be using a different tenant configuration.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInTest.kt
Yuki-YuXin: Agreed. Should I include those in this PR? I personally think a new one is required specific for the config. Another idea is that we are adding a lots of parameters into the command script, maybe be we can aggregate them into the <config>

SammyO: Up to you how you want to submit them PR-wise. We won't need a different tenant config to use a different auth-type user (sign in auth is at a user-level, not application level. only sign up is on application level), but we will indeed need a different config for email-OTP sign up flows. 

I agree re. the parameters; we may need a better solution for that. let's discuss with the iOS team.

PR #2124, opened 2024-06-14
First batch of Native Auth E2E tests

Yuki-YuXin:
Has a dependency on the configuration parameters work

**Closed** company PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2438
**Valid** company PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2462
The previous E2E test PR for sign in cases https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2123

**SignIn username + code:**
- User is not registered with given email (use case 2.2.2) - Test case 31 ✔
- Invalid OTP (use case 2.2.7) - Test case 35 ✔
- Use email and OTP to get token and sign in (hero scenario 6, use case 2.2.1) - Test case 30 ✔

**SignIn username + password:**
- Password is incorrect (1.2.3) - Test case 39 ✔
- Use email and password to get token (hero scenario 15, use case 1.2.1) - test case 37 ✔
- User is not registered with given email (use case 1.2.2) - test case 38 ✔

**SignUp username + code:**
- Verify email address using email OTP and sign up (hero scenario 1, use case 2.1.1) - Test case 1 ✔
- Signup user with custom attributes with verify OTP as last step (hero scenario 2, use case 2.1.2) - Test case 2 ✔
- Verify email OTP first and then collect custom attributes (hero scenario 3, use case 2.1.3) - Test case 3 ✔
- Verify email OTP first and then collect custom attributes on multiple screens (hero scenario 4, use case 2.1.4) - Test case 4 ✔

**SignUp username + password:**
- Set email and password and then verify email OOB as last step (hero scenario 9, use case 1.1.1) - Test case 13 ✔
- Sign up user with custom attributes and verify email as last step (hero scenario 10, use case 1.1.3) - Test case 15 ✔
- Verify email address using email OTP and then set password (use case 1.1.4) - Test case 16 ✔
- Sign up with verify email OOB as first step, then set password & custom attributes at end (hero scenario 12, use case 1.1.6) - Test case 28 ✔
- Sign up with verify email OOB as first step, then set password & custom attributes at end over multiple screens/API calls (hero scenario 13) - Test case 29 ✔

**SSPR:**
- Verify email with email OTP first and then reset password (hero scenario 8 & 17, use case 3.1.1) - Test case 46 ✔

**Id Token**
https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/2920473 This work item cannot be completed using mock API but current we have the condition environment for it. Finish the test/work item in this branch.


msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpTest.kt
```diff
@@ -48,6 +48,29 @@ class SignUpTest : NativeAuthPublicClientApplicationAbstractTest() {
         Dispatchers.setMain(testDispatcher)
     }
 
+    private fun <T> retryOperation(
+        maxRetries: Int = 3,
+        onFailure: () -> Unit = { Assert.fail() },
+        block: () -> T
```
SammyO: great idea!
let's rename this var to something else? This to me sounds like "blocking", as in the test is being blocked or something.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpTest.kt
```diff
@@ -48,6 +48,29 @@ class SignUpTest : NativeAuthPublicClientApplicationAbstractTest() {
         Dispatchers.setMain(testDispatcher)
     }
 
+    private fun <T> retryOperation(
```
SammyO: Let's move this into a separate file, as we're going to need to re-use this in many tests.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpTest.kt
```diff
@@ -48,6 +48,29 @@ class SignUpTest : NativeAuthPublicClientApplicationAbstractTest() {
...
+        maxRetries: Int = 3,
+        onFailure: () -> Unit = { Assert.fail() },
+        block: () -> T
+    ): T? {
+        var retryCount = 0
+        var shouldRetry = true
+
+        while (shouldRetry) {
+            try {
+                return block()
```
SammyO: Is this return needed? I.e. the entire use of generics `T`. We only need the `block` code to execute, right?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpTest.kt
```diff
@@ -60,31 +83,20 @@ class SignUpTest : NativeAuthPublicClientApplicationAbstractTest() {
...
                 val password = getSafePassword()
                 signUpResult = application.signUp(user, password.toCharArray())
                 Assert.assertTrue(signUpResult is SignUpResult.CodeRequired)
                 otp = tempEmailApi.retrieveCodeFromInbox(user)
                 val submitCodeResult = (signUpResult as SignUpResult.CodeRequired).nextState.submitCode(otp)
                 Assert.assertTrue(submitCodeResult is SignUpResult.Complete)
-                shouldRetry = false
-                break
-            } catch (e: IllegalStateException) {
-                // Re-run this test if the OTP retrieval fails. 1SecMail is known for emails to sometimes never arrive.
```
SammyO, line 71: Can we leave the comment, somewhere in the test? I feel like it's important context for (future) devs to know about.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpTest.kt
Yuki-YuXin: Removed it to NativeAuthPublicClientApplicationAbstractTest as Robert suggested.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpTest.kt
Yuki-YuXin: How about renaming it as authFlow()?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpTest.kt
Yuki-YuXin: Remove the T? and return null.

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "microsoft-authentication-library-common-for-android"]
 	path = common
 	url = https://github.com/AzureAD/microsoft-authentication-library-common-for-android.git
+	branch = yuki/key-vault-test
```
Yuki-YuXin: Revert it back when key vault branch get merged

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
```diff
@@ -91,21 +103,56 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
         return credential.password
     }
 
-    private fun setupPCA() {
-        val clientId = NativeAuthCredentialHelper.nativeAuthLabsEmailPasswordAppId
-        val authorityUrl = NativeAuthCredentialHelper.nativeAuthLabsAuthorityUrl
+    private fun getConfigsThroughSecretValue(): Map<String, NativeAuthTestConfig.Config>? {
+        val secretValue = KeyVaultHelper.getSecretForBuildAutomation("msalandroidnativeauthautomationconfjsonfile")
```
Yuki-YuXin: Maybe this string can be extracted as constant rather than hard code it.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SSPRTest.kt
```diff
@@ -62,34 +52,24 @@ class SSPRTest : NativeAuthPublicClientApplicationAbstractTest() {
...
+        var otp: String
 
-        while (shouldRetry) {
-            try {
-                val user = NativeAuthCredentialHelper.nativeAuthSSPRUsername
-                val result = application.resetPassword(user)
-                Assert.assertTrue(result is ResetPasswordStartResult.CodeRequired)
-                val otp = tempEmailApi.retrieveCodeFromInbox(user)
+        retryOperation {
+            runBlocking {
```
Yuki-YuXin: "Running with runBlocking to avoid default 10 second execution timeout." Maybe this comment can stay here.

Yuki-YuXin:
Approved but "Pull request authors can’t approve their own pull request"

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/GetAccessTokenTests.kt
```diff
@@ -394,47 +436,67 @@ class GetAccessTokenTests : NativeAuthPublicClientApplicationAbstractTest() {
...
+        val result = application.signIn(
+            username = username,
+            password = password.toCharArray()
+        )
+        assertState<SignInResult.Complete>(result)
+        val accountState = (result as SignInResult.Complete).resultValue
+        val idToken1 = accountState.getIdToken()
+        accountState.getAccessToken(forceRefresh = true)
+        val idToken2 = accountState.getIdToken()
+        Assert.assertFalse(idToken1 == idToken2)
```
SammyO: @Yuki-YuXin please use `assertNotEquals`. Besides slightly different validation, it gives more helpful information in case the assert fails.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
```diff
@@ -131,4 +131,28 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
...
+        authFlow: () -> T
+    ) {
+        var retryCount = 0
+        var shouldRetry = true
+
+        while (shouldRetry) {
+            try {
+                authFlow()
+                shouldRetry = false // authFlow() has succeeded, so we don't need to retry.
+            } catch (e: IllegalStateException) {
```
GBakolas, line 147: Is this the only possible exception type that could be thrown? Should we catch a broader range for unexpected exceptions?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailOTPTest.kt
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    /**
+     * Use invalid email address to receive a "user not found" error.
+     * (use case 2.2.2, Test case 31)
+     */
+    @Test
+    fun testErrorIsUserNotFound() = runTest {
+        val user = config.email
```
GBakolas, line 68: Can we avoid reading the valid email from the configuration, why not just generate a random string for the invalid username

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailOTPTest.kt
```diff
@@ -0,0 +1,96 @@
...
+     * (use case 2.2.7, Test case 35)
+     */
+    @Test
+    fun testErrorIsInvalidCode() {
+        retryOperation {
+            runBlocking {// Running with runBlocking to avoid default 10 second execution timeout.
+                val user = config.email
+                val signInResult = application.signIn(user)
+                assertState<SignInResult.CodeRequired>(signInResult)
+                val otp = tempEmailApi.retrieveCodeFromInbox(user)
```
GBakolas, line 87: Since we want to sent an invalid otp why not avoid retrieving the correct one from the email? is the intention to check that the otp has been sent successfully?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
```diff
@@ -30,26 +30,49 @@ import kotlinx.coroutines.test.runTest
...
     override val configType = ConfigType.SIGN_IN_PASSWORD
 
+    /**
+     * Use valid email and password to get token.
+     * (hero scenario 15, use case 1.2.1, Test case 37)
+     */
     @Test
-    fun testSignInErrorSimple() = runTest {
+    fun testSuccess() = runTest {
         val username = config.email
```
GBakolas, line 43: The email and the password are needed in every test, maybe it's better to make these variables members of the class and initialize them only once and not need to declare them in every test.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailOTPTest.kt
SammyO: In theory, setting the "invalid" value to a predefined, static one could coincidentally be the correct OTP. The only way to guarantee an incorrect one is by taking the correct one and editing it to something else. But the likelihood of that is very small, so I agree this is a good optimisation to make.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailOTPTest.kt
SammyO: what is the benefit of doing one over the other?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
SammyO: We should only catch expected exceptions (in this case, related to the retry). Catching all/too many exceptions will cause unexpected exceptions (due to bugs or other side effects we're not aware of) to also be caught and not make the test fail, which is not what we want.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailOTPTest.kt
GBakolas: We can set the predefined invalid OTP to be a short one (if I'm not mistaken the OTPs are usually 7 or 8 digits) or give it a really bogus value like 00000 to avoid this edge case

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailOTPTest.kt
GBakolas: Sorry I haven't phrased the comment correctly. I meant avoid reading the correct email and then making it invalid by appending the extra character. This feels a little weird. Just create a random string for the invalid username.

SammyO (approved):

rmccahill (approved):

PR #2130, opened 2024-06-26
Commenting out the extra MSAL renaming configuration

melissaahn:
### Summary
After I made some changes related to the POM logic, we see that this renaming task config (which was originally added as a workaround) is renaming the aar file earlier than expected for the vsts publish task, which is still looking for a file named msal-dist-release.aar. On the other hand, this renaming task config is necessary for Maven publishing. 
So while I'm investigating (wondering why this is an issue for only MSAL and not common), this PR comments out the logic in order to unblock the daily pipeline. But it should be uncommented while doing the maven release for MSAL.


somalaya (approved):

msal/build.gradle
```diff
@@ -308,16 +308,17 @@ task pmd(type: Pmd) {
...
-                println "Rename failed!"
-            }
-        }
-    }
+    // Please uncomment the section below for the Maven release step only. Currently troubleshooting why this extra renaming configuration is still needed for MSAL.
+//    tasks.named("assembleDistRelease").configure {
+//        def buildFile = file("$buildDir/outputs/aar/${archivesBaseName}-dist-release.aar")
+//        print "Build file $buildFile"
+//        doLast {
+//            println "Renaming build file $buildFile to '$buildDir${File.separator}outputs${File.separator}aar${File.separator}${archivesBaseName}-${version}.aar'"
```
iamgusain, line 316: >archivesBaseName [](http://example.com/codeflow?start=130&length=16)

I don't see this is done the similar way in common.

melissaahn: I've been trying to wrap my head around this as well. For some reason, the usual rename config outside of this afterEvaluate block (which common does have) doesn't work for MSAL, even though it works for common when doing the maven release. I wasn't around when this block was originally added in, so I don't have the context, but somehow it was ok to be present for both vsts and maven release processes. But with the latest changes I've made to this file, we'll have to comment and uncomment it, until we can figure out the difference between MSAL and common.

iamgusain (approved):

PR #2133, opened 2024-06-27
Manage test configuration through config file

SammyO:
Labs API doesn't have support for the various applications, users, etc. that are needed for native auth test cases. Instead, we manage these values manually - currently through environment variables (CI env. vars. and Gradle parameters). These variables don't scale, so we're moving to having all the variables that are needed in a configuration file that's stored in a Labs KeyVault.

MSAL common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2439

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "microsoft-authentication-library-common-for-android"]
 	path = common
 	url = https://github.com/AzureAD/microsoft-authentication-library-common-for-android.git
+	branch = yuki/key-vault-test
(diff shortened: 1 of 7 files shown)


.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "microsoft-authentication-library-common-for-android"]
 	path = common
 	url = https://github.com/AzureAD/microsoft-authentication-library-common-for-android.git
+	branch = yuki/key-vault-test
```
Yuki-YuXin, line 4: Will revert it back one the common one is merged.

fadidurah (approved):

Yuki-YuXin (approved):

PR #2135, opened 2024-07-01
Remove the unnecessary code line because of fix in common

Yuki-YuXin:
Company PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2435

1. Remove the conditional setting of the correlation ID in AccountState since it's a quick fix to pass the tests.
2. Add try catch block for the UUID to String

### DIFF
--- a/azure-pipelines/pull-request-validation/pr-msal.yml
+++ b/azure-pipelines/pull-request-validation/pr-msal.yml
@@ -30,7 +30,6 @@ resources:
   - repository: common
     type: github
     name: AzureAD/microsoft-authentication-library-common-for-android
-    ref: robert/empty-correlation-id-fixes
     endpoint: ANDROID_GITHUB
 
 pool:
--- a/azure-pipelines/pull-request-validation/pr-msal.yml
+++ b/azure-pipelines/pull-request-validation/pr-msal.yml
@@ -30,7 +30,7 @@ resources:
   - repository: common
     type: github
     name: AzureAD/microsoft-authentication-library-common-for-android
-    ref: dev
+    ref: robert/empty-correlation-id-fixes
     endpoint: ANDROID_GITHUB
 
 pool:
--- a/azure-pipelines/pull-request-validation/pr-msal.yml
+++ b/azure-pipelines/pull-request-validation/pr-msal.yml
@@ -68,7 +68,7 @@ jobs:
   - task: Gradle@2
     displayName: Run Unit tests
     inputs:
-      tasks: msal:testLocalDebugUnitTest -Plabtest -PlabSecret=$(LabVaultAppCert) -ProbolectricSdkVersion=${{variables.robolectricSdkVersion}} -PmockApiUrl=$(MOCK_API_URL) -PnativeAuthSSPRTestsUsernameValue=$(NativeAuthSSPRTestsUsernameValue) -PnativeAuthSignInTestsUsernameValue=$(NativeAuthSignInTestsUsernameValue) -PnativeAuthLabsEmailPasswordAppIdValue=$(NativeAuthLabsEmailPasswordAppId) -PnativeAuthLabsAuthorityUrlValue=$(NativeAuthLabsAuthorityUrlValue) -PnativeAuthEmployeeWriteAllScopeValue=$(NativeAuthEmployeeWriteAllScopeValue) -PnativeAuthEmployeeReadAllScopeValue=$(NativeAuthEmployeeReadAllScopeValue) -PnativeAuthCustomerWriteAllScopeValue=$(NativeAuthCustomerWriteAllScopeValue) -PnativeAuthCustomerReadAllScopeValue=$(NativeAuthCustomerReadAllScopeValue) -PnativeAuthInvalidScopeValue=$(NativeAuthInvalidValue)
+      tasks: msal:testLocalDebugUnitTest -Plabtest -PlabSecret=$(LabVaultAppCert) -ProbolectricSdkVersion=${{variables.robolectricSdkVersion}} -PmockApiUrl=$(MOCK_API_URL)
       javaHomeSelection: $(BuildParameters.javaHomeSelection)
       jdkVersion: 1.11
 - job: spotbugs
--- a/azure-pipelines/pull-request-validation/pr-msal.yml
+++ b/azure-pipelines/pull-request-validation/pr-msal.yml
@@ -30,6 +30,7 @@ resources:
   - repository: common
     type: github
     name: AzureAD/microsoft-authentication-library-common-for-android
+    ref: dev
     endpoint: ANDROID_GITHUB
 
 pool:
--- a/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
@@ -363,7 +363,7 @@ public class CommandParametersAdapter {
                 .userAttributes(userAttributes)
                 // Start of the flow, so there is no correlation ID to use from a previous API response.
                 // If default value "UNSET", generate a new UUID as correlationId.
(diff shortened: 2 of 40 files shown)


azure-pipelines/pull-request-validation/pr-msal.yml
```diff
@@ -30,7 +30,7 @@ resources:
   - repository: common
     type: github
     name: AzureAD/microsoft-authentication-library-common-for-android
-    ref: dev
+    ref: robert/empty-correlation-id-fixes
```
Yuki-YuXin: Revert it back before the merge.

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -360,8 +362,8 @@ public static SignUpStartCommandParameters createSignUpStartCommandParameters(
                 .challengeType(configuration.getChallengeTypes())
                 .userAttributes(userAttributes)
                 // Start of the flow, so there is no correlation ID to use from a previous API response.
-                // Set it to a default value.
-                .correlationId(DiagnosticContext.INSTANCE.getThreadCorrelationId())
+                // If default value "UNSET", generate a new UUID as correlationId.
+                .correlationId(Objects.equals(DiagnosticContext.INSTANCE.getThreadCorrelationId(), "UNSET") ? UUID.randomUUID().toString() : DiagnosticContext.INSTANCE.getThreadCorrelationId())
```
SammyO: This logic doesn't belong here. it's the responsibility of DiagnosticContext to generate an ID, and do additional null checks to ensure an actual ID is returned.

Yuki-YuXin: Update the logic inside DiagnosticContext getThreadCorrelationId()

azure-pipelines/pull-request-validation/pr-msal.yml
SammyO: Why do we have this here?

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -262,7 +262,7 @@ class NativeAuthPublicClientApplication(
                     GetAccountResult.AccountFound(
                         resultValue = AccountState.createFromAccountResult(
                             account = account,
-                            correlationId = DiagnosticContext.INSTANCE.threadCorrelationId,
+                            correlationId = "",
```
SammyO: Could you check what MSAL does here?

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -341,7 +341,7 @@ class NativeAuthPublicClientApplication(
                     return@withContext SignInError(
                         errorType = ErrorTypes.INVALID_USERNAME,
                         errorMessage = "Empty or blank username",
-                        correlationId = "UNSET"
+                        correlationId = ""
```
SammyO: Why not set it at all, meaning the value will be null?

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
SammyO: Also, what was your thinking behind changing this from the thread ID to an empty string?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -77,7 +77,7 @@ class AccountState private constructor(
 
     constructor (parcel: Parcel) : this (
         account = parcel.serializable<IAccount>() as IAccount,
-        correlationId = parcel.readString() ?: "UNSET",
+        correlationId = parcel.readString() ?: "",
```
SammyO: Doesn't `UUID.fromString("")` also produce an exception?

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
Yuki-YuXin: This path generate AccountState from the SDK cache (authClient.getCurrentAccount())

The default value of DiagnosticContext.INSTANCE.threadCorrelationId is UNSET. And currently, SDK won't transform “UNSET" into UUID.

If use  DiagnosticContext.INSTANCE.threadCorrelationId, 
1. authClient.getCurrentAccount() before signIn. correlation id would be "UNSET"
2. signIn. correlation id should be UUID
3. authClient.getCurrentAccount() after signIn. When getting correlation id again then becomes UUID as the sign in flow.

I am not sure if the difference between the two correlation ids before and after is what we want

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
Yuki-YuXin: That's a good point. How about generating random UUID here?

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
Yuki-YuXin: The correlationId parameter is mandatory. Setting it to “” would be much easier than dealing with if null.

azure-pipelines/pull-request-validation/pr-msal.yml
Yuki-YuXin: I cannot change the targe branch in the DevOps portal thus change the command here.

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -582,7 +582,7 @@ class NativeAuthPublicClientApplication(
                     return@withContext SignUpError(
                         errorType = ErrorTypes.INVALID_USERNAME,
                         errorMessage = "Empty or blank username",
-                        correlationId = "UNSET"
+                        correlationId = DiagnosticContext.INSTANCE.threadCorrelationId
```
Yuki-YuXin, line 585: CommandDispatcher.submitSilentReturningFuture will clear the context at end every time.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
Yuki-YuXin: Keep the original "UNSET" string.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
```diff
@@ -309,12 +309,17 @@ class AccountState private constructor(
                             correlationId = correlationId
                         )
 
-                val privateCorrelationId = if (correlationId == "UNSET") { UUID.randomUUID().toString() } else { correlationId }
+                try {
+                    UUID.fromString(correlationId)
```
GBakolas: nit: maybe assign the result to a variable and use it in the builder to avoid calling the fromString() method twice

GBakolas (dismissed):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/AccountState.kt
SammyO: @Yuki-YuXin  let's run the tests without this fix here to ensure it works not. But let's keep this fix here as well, to avoid any production crashes (but clean it up a bit, e.g. call `DiagnosticContext.INSTANCE.getThreadCorrelationId()` so that a new UUID is generated in there, rather than here at the call site).

Yuki-YuXin: Local GetAccessTokenTests passed with and without the changes here.

Yuki-YuXin:
A hint to add changelog entry

PR #2138, opened 2024-07-09
Moving OS version check for passkeys

melissaahn:
### Summary
If a device's OS does not support passkeys, then we shouldn't even reveal the passkey option. This means that the OS check should be done while we are deciding to add webauthn=1 or not. 
This PR moves the OS checks to earlier points and updates the corresponding unit tests.

### Related PRs
- https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2450


fadidurah (approved):

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -1143,10 +1144,26 @@ public static GenerateShrCommandParameters createGenerateShrCommandParameters(
...
+            if (queryStringParameters != null) {
+                try {
+                    queryStringParameters.remove(webauthnExtraParameter);
+                } catch (final UnsupportedOperationException e) {
+                    final List<Map.Entry<String, String>> result = new ArrayList<>(queryStringParameters);
+                    result.remove(webauthnExtraParameter);
+                    return result;
+                }
+            }
+        }
```
shahzaibj: This is basically us "correcting" the apps?

shahzaibj: Also I'm wondering if this whole OS check and removal of qp can go into a helper method in common? and then invoke that method in both MSAL and Broker?

testapps/testapp/src/main/res/raw/msal_config_msa.json
```diff
@@ -1,6 +1,7 @@
 {
   "client_id" : "9668f2bd-6103-4292-9024-84fa2d1b6fb2",
   "authorization_user_agent" : "DEFAULT",
+  "webauthn_capable": true,
```
shahzaibj: Do we also have a test for not capable scenario?

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
melissaahn: Addressed this comment by creating the FidoUtil class in common: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2450/commits/6857ba4522f53a10538c82123021102d199b72ea
And this commit for MSAL: 818ff76d2ace0eb3b3040a7d48ea07024fa60e77

testapps/testapp/src/main/res/raw/msal_config_msa.json
melissaahn: Added a couple of tests: c6a6f63b6b42e176d8c0108e8cea1875a3bacb5a
But noting that I removed this change in the MSA config files (was for testing; don't mean to merge this in)

shahzaibj (approved):

PR #2143, opened 2024-07-16
Update the msal test app for Native Auth

algo-1:
Fix the MSAL test app to support custom attribute test cases without the need to modify the source code.

[ADO Ticket (2979358)](https://dev.azure.com/IdentityDivision/Engineering/_boards/board/t/DevExDub%20-B2C/Backlog%20items/?workitem=2979358)

### DIFF
--- a/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/nativeauth/SignUpAttributesFragment.kt
+++ b/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/nativeauth/SignUpAttributesFragment.kt
@@ -57,7 +57,7 @@ class SignUpAttributesFragment : Fragment() {
         _binding = FragmentAttributeBinding.inflate(inflater, container, false)
 
         val bundle = this.arguments
-        currentState = bundle!!.getSerializable(Constants.STATE) as SignUpAttributesRequiredState
+        currentState = (bundle?.getParcelable(Constants.STATE) as? SignUpAttributesRequiredState)!!
 
         init()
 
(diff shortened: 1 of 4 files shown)


algo-1:
@microsoft-github-policy-service agree company="Microsoft"

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/nativeauth/SignUpAttributesFragment.kt
```diff
@@ -80,23 +79,21 @@ class SignUpAttributesFragment : Fragment() {
...
-//                }
-//
-//                val attr2Key = binding.attr2KeyText.text.toString()
-//                if (attr2Key.isNotBlank()) {
-//                    val attr2Value = binding.attr2ValueText.toString()
-//                    attributes
-//                        .customAttribute(attr2Key, attr2Value)
-//                }
+                val attr1Key = binding.attr1KeyText.text.toString()
+                if (attr1Key.isNotBlank()) {
```
SammyO: Let's not do this check. We may want to execute tests that validate whether the SDK's string check works correctly, or whether the API's string check works correctly. Same for the one below.

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/nativeauth/SignUpPasswordFragment.kt
```diff
@@ -55,7 +55,7 @@ class SignUpPasswordFragment : Fragment() {
         _binding = FragmentPasswordBinding.inflate(inflater, container, false)
 
         val bundle = this.arguments
-        currentState = bundle!!.getSerializable(Constants.STATE) as SignUpPasswordRequiredState
+        currentState = (bundle?.getParcelable(Constants.STATE) as? SignUpPasswordRequiredState)!!
```
SammyO, line 58: This needs to be fixed in many more places in the test app. Could you include that in this PR?

Yuki-YuXin (approved):

SammyO (approved):

PR #2145, opened 2024-07-24
Fix UserAttributesBuilder bug

Yuki-YuXin:
When using UserAttributes.Builder multiple times, e.g. when submitting attributes across multiple steps/screens, previously set data is persisted. This is due to the Builder (unintentionally) being a singleton. The impact of this is that unexpected/unintended user attributes are sent to the server, causing an incorrect user flow.

This PR extract related changes from https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2124/commits/0c8665134a8daa58526fc9a77832b8bac3067b7c

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/nativeauth/UserAttributes.kt
+++ b/msal/src/main/java/com/microsoft/identity/nativeauth/UserAttributes.kt
@@ -29,17 +29,19 @@ import com.microsoft.identity.common.java.util.ObjectMapper
  * operation in Native Auth
  */
 class UserAttributes(internal val userAttributes: Map<String, String>) {
-    companion object Builder {
-        private const val CITY = "city"
-        private const val COUNTRY = "country"
-        private const val DISPLAY_NAME = "displayName"
-        private const val EMAIL_ADDRESS = "email"
-        private const val GIVEN_NAME = "givenName"
-        private const val JOB_TITLE = "jobTitle"
-        private const val POSTAL_CODE = "postalCode"
-        private const val STATE = "state"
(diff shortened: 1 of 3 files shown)


msal/src/main/java/com/microsoft/identity/nativeauth/UserAttributes.kt
```diff
@@ -29,17 +29,19 @@ import com.microsoft.identity.common.java.util.ObjectMapper
...
-        private const val DISPLAY_NAME = "displayName"
-        private const val EMAIL_ADDRESS = "email"
-        private const val GIVEN_NAME = "givenName"
-        private const val JOB_TITLE = "jobTitle"
-        private const val POSTAL_CODE = "postalCode"
-        private const val STATE = "state"
-        private const val STREET_ADDRESS = "streetAddress"
-        private const val SURNAME = "surname"
+    class Builder {
+        companion object {
```
GBakolas, line 33: Not an issue, just something to consider. Do we need the companion object here, what if simply make the constants members of the Builder class (declare them as val without the const keyword)?

GBakolas (approved):

SammyO:
@Yuki-YuXin could you run the tests from https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2124 that are affected by this bug, and ensure they now pass?

Yuki-YuXin:
@SammyO Ran local tests (SignUpEmailOTPAttributesTest & SignUpEmailPasswordAttributesTest)to verify that they passed.

SammyO (approved):

PR #2156, opened 2024-08-09
Adding MSA Sign Up UI Test and accompanying ADO Test Case item

fadidurah:
This PR adds a new ui automation test to handle MSA sign up flow. Also took some time to implement "Extra query parameters" field in msal test app to support validation for such flows in the future.

ADO Test Case: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3007768
PBI: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2985851

Validated in this daily run: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1343275&view=ms.vss-test-web.build-test-results-tab


testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalWrapper.java
```diff
@@ -140,12 +140,25 @@ private AcquireTokenParameters.Builder getAcquireTokenParametersBuilder(@NonNull
                 .withPrompt(requestOptions.getPrompt())
                 .withCallback(getAuthenticationCallback(callback));
 
-        // create extra query parameters list and add "is_remote_login_allowed=true"
+        // Create extra query parameters list
+        final List<Map.Entry<String, String>> extraQP = new ArrayList<>();
+
+        // Split given extra query parameters and add them to extra query parameter list
+        // If any were passed in the RequestOptions
+        final String[] requestGenericExtraQueryParams = requestOptions.getExtraQueryParams().split(" ");
```
mohitc1: >" " [](http://example.com/codeflow?start=99&length=3)

Would different query params be separated by a whitespaces in UI?

mohitc1: you can instead take usual format like key1=value1&key2=value2&key3=value3

fadidurah: Yes, similar to scopes from what i understand

image

mohitc1 (approved):
Left a comment. Thanks for these changes.

testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalWrapper.java
fadidurah: I can also do &, perhaps that will make things easier if someone is copying from an existing url :)

mohitc1: space is more natural to scopes as they are passed that way in the request.

fadidurah: Changed this to use & instead of whitespace

melissaahn (approved):

PR #2158, opened 2024-08-12
MFA SDK interface

SammyO:
New states:
- AwaitingMFAState: returned as part of `SignInResult.MFARequiredResult`
- MFARequiredState: returned for all other MFA operations, such as `getAuthMethods()` and `sendChallenge()`

New results:
- `SignInResult.MFARequired(nextState: AwaitingMFAState)`: returned when /token returns `mfa_required` error
- `MFARequiredResult.VerificationRequired(nextState: MFARequiredState, val codeLength: Int, ...)`: returned when /challenge has successfully challenged the user's default or selected auth method.
- `MFARequiredResult.SelectionRequired(nextState: MFARequiredState, authMethods: List<AuthMethod>)`: returned when /introspect is called and returns a list of auth methods

See `SignInTest.testSignInMFASimple()` for an example scenario.

**Note**: all SDK responses are currently hardcoded and are not hooked up to underlying business logic. This PR should  therefore not be merged into `dev`


msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -333,184 +326,193 @@ class NativeAuthPublicClientApplication(
...
-                    NativeAuthMsalController(),
-                    PublicApiId.NATIVE_AUTH_SIGN_IN_WITH_EMAIL
+            SignInResult.MFARequired(
+                nextState = AwaitingMFAState(
+                    continuationToken = "1234",
+                    correlationId = "abcd",
+                    scopes = scopes,
+                    config = nativeAuthConfig
                 )
+            )
```
rmccahill: I'm confused by the logic here - won't this break all our sign in flows? Why are we removing all of the handling logic and hard-coding an MFA required response?

Edit: just read the PR description, I understand now. Probably worth marking this as a draft PR just to make sure we don't merge this in.

Also I know that this is just a demonstration, but maybe instead of removing all the logic here, is it worth constructing the initial state in the MFA test, instead of hard-coding the sign in logic? That way there's still visibility on how the flow works without having to remove / comment out all the logic here.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/results/SignInResult.kt
```diff
@@ -92,12 +106,12 @@ interface SignInResendCodeResult : Result {
      *
      * @param codeLength the length of the code required by the server.
      * @param sentTo the email/phone number the code was sent to.
-     * @param channel channel(email/phone) the code was sent through.
+     * @param channel channel(email/phone) the code was sent    through.
```
rmccahill: nit: extra space probably shouldn't be here

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
SammyO: I can look into setting the state in the tests.

I want to actually merge this into the MFA epic branch, so we can start building the sample app against it.

SammyO: The alternative is that the interface and business logic branches are lumped together into 1 PR, but that makes reviewing quite substantial

rmccahill: I hadn't seen the target branch - that makes more sense. 👍

SammyO: Good suggestion, I updated this.

rmccahill (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/AuthMethods.kt
```diff
@@ -0,0 +1,68 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.nativeauth
+
+import com.microsoft.identity.common.java.nativeauth.providers.responses.signin.AuthenticationMethodApiResult
+import com.microsoft.identity.common.java.nativeauth.util.ILoggable
+
+/**
+ * RequiredUserAttribute represents details about the account attributes required by the server.
```
GBakolas: comment needs updating

msal/src/main/java/com/microsoft/identity/nativeauth/AuthMethods.kt
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Converts a list of auth method API response to a list of [AuthMethod] objects
+ */
+internal fun List<AuthenticationMethodApiResult>.toListOfAuthMethods(): List<AuthMethod> {
+    return this.map { it.toAuthMethod() }
+}
+
+/**
+ * Converts the required user attribute API received as part of signup API to
```
GBakolas: comment needs updating

GBakolas (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/AuthMethods.kt
nilo-ms, line 29: nit: is this the right comment?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/MFAStates.kt
```diff
@@ -0,0 +1,220 @@
...
+        )
+    }
+
+    // Call /token
+    suspend fun submitChallenge(code: Int): MFASubmitChallengeResult {
+        // If /token returns HTTP 200
+        return if (true) {
+            SignInResult.DummyComplete()
+        } else {
+            // If /token returns another mfa_required error
```
nilo-ms: Does this means that the `MFASubmitChallengeResult` has `SignInComplete` and  `MFARequired` type? I thought we wanted to limit to only a second factor authentication.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/results/MFAResult.kt
```diff
@@ -0,0 +1,45 @@
...
+        val channel: String,
+    ) : MFARequiredResult, Result.SuccessResult(nextState = nextState)
+
+    class SelectionRequired(
+        override val nextState: MFARequiredState,
+        val authMethods: List<AuthMethod>
+    ) : MFARequiredResult, MFAGetAuthMethodsResult, Result.SuccessResult(nextState = nextState)
+}
+
+interface MFASubmitChallengeResult : Result
```
nilo-ms, line 43: Where can I see the subtype of this result?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/MFAStates.kt
SammyO: Correct. We're currently discussing this in the design doc.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/results/MFAResult.kt
SammyO: In IDE, click on "implements" arrow. In the PR, look for classes/interfaces that have `: MFASubmitChallengeResult`.

image

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/MFAStates.kt
SammyO: I've removed it for now, and will put it back after we conclude the discussion with Kaushik.

PR #2176, opened 2024-09-11
MFA E2E tests

SammyO:


msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInMFATest.kt
```diff
@@ -0,0 +1,180 @@
...
+    @Test
+    fun `test submit invalid challenge, request new challenge, submit correct challenge and complete MFA flow`() = runTest {
+        val username = config.email
+        val password = getSafePassword()
+        val result = application.signIn(username, password.toCharArray())
+        assertResult<SignInResult.MFARequired>(result)
+
+        // Initiate challenge, send code to email
+        val sendChallengeResult = (result as SignInResult.MFARequired).nextState.requestChallenge()
+        assertResult<MFARequiredResult.VerificationRequired>(sendChallengeResult)
```
nilo-ms, line 64: What this assertResult does? Does just check the type of the result or does something more?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInMFATest.kt
```diff
@@ -0,0 +1,180 @@
...
+        (sendChallengeResult as MFARequiredResult.VerificationRequired)
+        assertNotNull(sendChallengeResult.sentTo)
+        assertNotNull(sendChallengeResult.codeLength)
+        assertNotNull(sendChallengeResult.channel)
+
+        // Retrieve other auth methods
+        val getAuthMethodsResult = sendChallengeResult.nextState.getAuthMethods()
+        assertResult<MFARequiredResult.SelectionRequired>(getAuthMethodsResult)
+        (getAuthMethodsResult as MFARequiredResult.SelectionRequired)
+        assertTrue(getAuthMethodsResult.authMethods.size == 1)
```
nilo-ms, line 121: I'd consider completing the test if the authMethods size is 0. If not, an error will be encountered at line 123

nilo-ms, line 121: On iOS we check also that the challenge type is "email"

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInMFATest.kt
SammyO: it's a fancy version of `assertTrue(sendChallengeResult is MFARequiredResult.VerificationRequired)`. If that previous statement fails, it just shows that the assert wasn't true, i.e. sendChallengeResult is not MFARequiredResult.VerificationRequired. However that's not very informative to an engineer inspecting the result: we want to know what sendChallengeResult is in that case. So the wrapper prints that information if the assertion fails.

SammyO: See `AssertUtils.kt`

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInMFATest.kt
SammyO: the test will be completed (with failure) if an assert fails. So if the size != 1, the test will fail and stop, meaning line 123 won't be executed.

rmccahill (approved):

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInMFATest.kt
SammyO: updated

PR #2185, opened 2024-09-18
Introduce specific MFA errors

SammyO:
- specific error classes makes sense when you have errors that can only happen in one method. E.g. `invalidAttributesError()` should only be available after `submitAttributes()`, i.e. we wrap that utility method in `SignUpSubmitAttributesError` which is emitted from `submitAttributes()`
- this has worked well so far: every method we have has method-specific errors, so there has always been a reason to create dedicated error classes
- however, for MFA this is not the case. Only `submitChallenge()` has a specific error (invalid challenge), all the rest doesn't. 
- however, we should still introduce specific MFA error classes anyway, because without it it would either be a breaking change when we add it (e.g. `requestChallenge()` will not longer emit `MFAError`, but now emit `MFASubmitChallengeError`), or developers are going to need to check for both error classes (e.g. `if error is MFAError || error is MFASubmitChallengeError`)

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/MFAErrors.kt
+++ b/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/MFAErrors.kt
@@ -5,7 +5,7 @@ import com.microsoft.identity.nativeauth.statemachine.results.MFARequiredResult
 import com.microsoft.identity.nativeauth.statemachine.results.MFASubmitChallengeResult
 
 /**
- * MFA error. Use the utility methods of this class
+ * MFA request challenge error. Use the utility methods of this class
  * to identify and handle the error. This error is produced by
  * [com.microsoft.identity.nativeauth.statemachine.states.MFARequiredState.requestChallenge] and
  * [com.microsoft.identity.nativeauth.statemachine.states.AwaitingMFAState.requestChallenge].
@@ -27,7 +27,7 @@ class MFARequestChallengeError(
 ): MFARequiredResult, BrowserRequiredError, Error(errorType = errorType, error = error, errorMessage= errorMessage, correlationId = correlationId, errorCodes = errorCodes, exception = exception)
 
 /**
- * MFA error. Use the utility methods of this class
+ * MFA get authentication methods error. Use the utility methods of this class
  * to identify and handle the error. This error is produced by
  * [com.microsoft.identity.nativeauth.statemachine.states.MFARequiredState.getAuthMethods]
  * @param errorType the error type value of the error that occurred.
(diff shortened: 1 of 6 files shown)


msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
```diff
@@ -2564,10 +2564,12 @@ public boolean isValidBrokerPackage(@NonNull final Context context,
     /**
      * Returns active name of the broker app.
      * NOTE: Expose to Company Portal for telemetry purpose.
+     * This method will be removed. Use BrokerAppsApi.getInstance(context).getActiveBrokerPackageName() method instead.
```
nilo-ms: Is this change expected?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/MFAErrors.kt
```diff
@@ -5,24 +5,47 @@ import com.microsoft.identity.nativeauth.statemachine.results.MFARequiredResult
...
+ * [com.microsoft.identity.nativeauth.statemachine.states.MFARequiredState.getAuthMethods]
  * @param errorType the error type value of the error that occurred.
  * @param error the error returned by the authentication server.
  * @param errorMessage the error message returned by the authentication server.
  * @param correlationId a unique identifier for the request that can help in diagnostics.
  * @param errorCodes a list of specific error codes returned by the authentication server.
  * @param exception an internal unexpected exception that happened.
  */
-class MFAError(
+class GetAuthMethodsError(
```
nilo-ms: Should this be "MFAGetAuthMethodsError" ?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/MFAErrors.kt
```diff
@@ -5,24 +5,47 @@ import com.microsoft.identity.nativeauth.statemachine.results.MFARequiredResult
 import com.microsoft.identity.nativeauth.statemachine.results.MFASubmitChallengeResult
 
 /**
- * MFA error. The user should use the utility methods of this class
- * to identify and handle the error. This error is produced by the various
- * MFA operations.
+ * MFA error. Use the utility methods of this class
```
nilo-ms: Maybe this should be "MFARequestChallengeError"
```suggestion
 * MFA RequestChallengeError. Use the utility methods of this class
```

Same comment applies also for other MFA error classes

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/MFAErrors.kt
SammyO: I didn't do it as get auth methods won't be specific to MFA (e.g. /introspect might be introduced in the sign up flow). But currently `getAuthMethods()` is on the `MFARequiredState`, so yeah you're right.

msal/src/main/java/com/microsoft/identity/client/PublicClientApplication.java
SammyO: removed

nilo-ms (approved):

PR #2190, opened 2024-09-25
Add test verifications for removing account from account mgr, Fixes AB#3001904

iamgusain:
[SDM] Added verification that removing account from account settings page removes account from broker and all other apps 
[Non SDM] Added Test case to verify removing accounts (WPJ owner and non-owner).

[AB#3001904](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3001904)


mohitc1 (approved):
:shipit:

fadidurah (approved):

PR #2199, opened 2024-10-07
Update and enable tests to run in pipeline

iamgusain:
Enabling 2 FLW test cases, which have been disabled to run in pipeline.

TestCase833513 - Updated account type and error verification
TestCase833515 - Replaced Edge steps with Outlook
TestCase1592468 - Removed as this is not valid

Ran locally with latest RC builds to verify these are passing.


fadidurah (approved):
Can you also update the test plan, moving 833513 to the automated folder?

mohitc1 (approved):

PR #2212, opened 2024-10-31
Android E2E Sign Up, Fixes AB#3073857

Yuki-YuXin:
[AB#3073857](https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/3073857)

PR #2213, opened 2024-10-31
Android E2E Sign Up, Fixes AB#3073857

Yuki-YuXin:
Create a draft version instead of an official one: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2212 because the PR is still in the draft version.

[AB#3073857](https://dev.azure.com/IdentityDivision/Engineering/_workitems/edit/3073857)

Android Test plan: https://identitydivision.visualstudio.com/Engineering/_testPlans/define?planId=3054662&suiteId=3054727
2.1.5 
2.1.6
2.1.7 (out of scopes for the Lab environment, include them once the environment is ready)
2.1.8
2.1.9
2.1.10
2.1.11

1.1.2
1.1.5
1.1.10
1.1.11
1.1.12
1.1.13
1.1.14


Yuki-YuXin:
image

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAnotherAbstractTest.kt
```diff
@@ -0,0 +1,156 @@
...
+import org.junit.After
+import org.junit.Assert
+import org.junit.Before
+import org.junit.runner.RunWith
+import org.mockito.Mockito
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+import org.robolectric.annotation.LooperMode
+
+// TODO: move to "PAUSED". A work in RoboTestUtils will be needed though.
```
nilo-ms: Is this TODO needed?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAnotherAbstractTest.kt
```diff
@@ -0,0 +1,156 @@
...
+@LooperMode(LooperMode.Mode.LEGACY)
+@RunWith(RobolectricTestRunner::class)
+@Config(shadows = [ShadowAndroidSdkStorageEncryptionManager::class])
+abstract class NativeAuthPublicClientApplicationAnotherAbstractTest : IPublicClientApplicationTest {
```
nilo-ms: @Yuki-YuXin Could you please clarify why you created a new abstract class for native auth public client application? I can see that a lot of code is repeated between this class and the NativeAuthPublicClientApplicationAbstractTest.kt file

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
```diff
@@ -85,8 +85,8 @@ class SignInEmailPasswordTest : NativeAuthPublicClientApplicationAbstractTest()
         Assert.assertTrue((result as SignInError).isInvalidCredentials())
     }
 
-    @Test
     @Ignore("Ignore until MFA is available on test slice")
```
nilo-ms: Can we remove this "Ignore" line? It should be available now

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
```diff
@@ -56,4 +68,153 @@ class SignUpEmailOTPTest : NativeAuthPublicClientApplicationAbstractTest() {
...
+    fun testResendCode() {
+        config = getConfig(defaultConfigType)
+        application = setupPCA(config, defaultChallengeTypes)
+
+        runBlocking { // Running with runBlocking to avoid default 10 second execution timeout.
+            val user = tempEmailApi.generateRandomEmailAddress()
+            val signUpResult = application.signUp(user)
+            assertResult<SignUpResult.CodeRequired>(signUpResult)
+            val codeRequiredState = (signUpResult as SignUpResult.CodeRequired).nextState
+            val resendCodeResult = codeRequiredState.resendCode()
```
nilo-ms: To test that the resend code works properly we should:
- signUp
- retrieve code from mailbox (otp1)
- resend code
- retrieve code from mailbox (otp2)
- check that otp1 and otp2 are different

Same comment applies also for use case 1.1.2

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAnotherAbstractTest.kt
Yuki-YuXin: Removing this TODO is not inside the work of this scope thus I would be prudential here of removing it.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAnotherAbstractTest.kt
Yuki-YuXin: The new abstract interface is to include more config and challenge situations based on the sign up test case needs. Meanwhile, there are still sign in files need the original abstract class. They are waiting for the next phase of work.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
Yuki-YuXin: Done

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
Yuki-YuXin: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2702377 According to the steps defined in the test case, SignInResendCodeResult.Success would be better than otp1 and otp2 difference.

nilo-ms: based on the test case you linked:
Screenshot 2024-11-20 at 16 57 55
You need to check that a new code is received in the mailbox.  Reviewing your code, it seems that this step is missing. Do you agree?
You decide the best way to check if a new code has been received.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAnotherAbstractTest.kt
nilo-ms: It's not clear to me why we need to create a new class. Could you include "more config and challenge situations" in the existing abstract class?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
Yuki-YuXin: SDK.resendCode() action then SDK return SignInResendCodeResult.Success Result. Receive a new code in the email box is the UI appearance. E2E test is focused on the SDK test without any UI requirement.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAnotherAbstractTest.kt
Yuki-YuXin: The original structure of abstract class can only have one set of config type and challenge types setting, which does not satisfy the needs of this sign up work. With the new structure of new abstract class, the input of config and challenge type is more flexible.

For example.
```
        config = getConfig(defaultConfigType)
        application = setupPCA(config, defaultChallengeTypes)
```
and 

```
        config = getConfig(ConfigType.SIGN_UP_PASSWORD)
        application = setupPCA(config, listOf("oob"))
```

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
nilo-ms: I think Yuki we're not understanding each other. Let's have a call and I can explain you what I mean

Yuki-YuXin: Would you mind explaining the things to me in the words? Better for me to understand them well.

nilo-ms: Me and Yuki had a chat, but we couldn't reach an agreement. Explaining here my reasoning why this test is missing a crucial piece:
- E2E tests are meant to validate the real product in a production environment. No component should be mocked. These tests should check that the entire product works as intended. In this case we want to test that the SDK resendCode feature works properly. This means, that when the sdk.resendCode() method is called, we need to check that the SDK returns a successful result *and* that a new email containing the OTP code is received in the email box.
- Additionally, we're automating the manual test (https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2702377). scenario 1.1.2 from acceptance criteria. One step of the test requires verifying that a new code is sent to the email box. When automating this test, we should include all steps.

Yuki-YuXin: That's a good point. Done.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
Yuki-YuXin: Removing this tag causes failure in the pipeline. I may need to re-add that.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
diegojerezba: I agree with @nilo-ms. Checking if the code has arrived in the mailbox should be included in our logic.

The E2E test simulates what a real user would do. A user would first check if the code has arrived before doing the next step. 

Also, as we’ve seen before, the main reason our tests are unreliable is that emails don’t always arrive on time. If we check for the email before moving to the next step, it will make it easier to understand why the test is failing (we can log this information to the console).

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAnotherAbstractTest.kt
nilo-ms: Can you make the input of config and challenge type of "NativeAuthPublicClientApplicationAbstractTest" more flexible? In this way we don't have to create a clone of this class? I don't think it's a good idea to have two almost identical classes:
- NativeAuthPublicClientApplicationAbstractTest
- NativeAuthPublicClientApplicationAnotherAbstractTest

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
nilo-ms: MFA is available for the lab tenant. What the pipeline error is about?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAnotherAbstractTest.kt
Yuki-YuXin: NativeAuthPublicClientApplicationAbstractTest is singleton pattern, which only accept one config. The improvement of NativeAuthPublicClientApplicationAnotherAbstractTest is to make it more flexible on config setting. The reason we have two almost identical classes is because the remaining classes except than sign up like sign in are still applied the singleton one. At last, there would be only one abstract class once we finished all the automation work.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
Yuki-YuXin: "Error: invalid_grant - AADSTS9002313: Invalid request. Request is malformed or invalid. Trace ID: 567dd14e-a142-4138-83d4-056260310100 Correlation ID: 48bd9ed2-491d-46cd-ba3a-83f32d6e23f2 Timestamp: 2024-11-20 16:50:25Z"

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAnotherAbstractTest.kt
diegojerezba: @Yuki-YuXin if refactoring the original `NativeAuthPublicClientApplicationAbstractTest` is too much work for this PR, can you add a comment saying that the work will be done as part of a different PR, and create a PBI for that? Let's try to target that PBI for this quarter, just right after this PR gets merged.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
diegojerezba: Can you try to run the test locally and fix it? From the code it seems the response is being mocked, so this test will probably need to be updated to work with the real API in the same line as the tests you've added in the PR.

For reference, this is the test file on iOS: https://github.com/AzureAD/microsoft-authentication-library-for-objc/blob/dev/MSAL/test/integration/native_auth/end_to_end/mfa/MSALNativeAuthSignInWithMFAEndToEndTests.swift

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAnotherAbstractTest.kt
spetrescu84: Could all tests (including the existing ones) not use the initialiser in `NativeAuthPublicClientApplicationAbstractTest`? If so, can't we delete the old one and make all tests use the new one? Or somehow prevent the duplicated code by subclassing?
I see these 2:
```
    private val defaultConfigType = ConfigType.SIGN_UP_OTP
    private val defaultChallengeTypes = listOf("password", "oob")
```
Being used in a lot of tests, and if that's why the class has to be duplicated, can't they be moved into a config file or made global so all tests can use them?

Yuki-YuXin: Hi @spetrescu84 , the first batch of E2E tests were all happy scenarios (no errors in the process), whereas this batch of E2E tests had many unhappy scenarios due to misconfiguration. The correct configuration of the first batch of E2E tests are all initialized in the` NativeAuthPublicClientApplicationAbstractTest` first.  After add the second batch, `SignUpEmailOTPTest` has tests using the correct configuration as well as tests using the wrong configuration (And they're wrong about different things). So there's no way to globalize the configuration, only to move it out of the ` NativeAuthPublicClientApplicationAbstractTest`

Yuki-YuXin: Hi @diegojerezba, I am going to let all the existing test cases use the new AbstractTest in the PR to prevent further PBI work/ delay. Tell me if you think it's inappropriate.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
Yuki-YuXin: Hi @nilo-ms  @diegojerezba . Under `SignInMFATest`, there is a test case called fun `test get other auth methods, request challenge on specific auth method and complete MFA flow`(). It overlaps with this one under `SignInEmailPasswordTest`, so I'll remove this test case

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
```diff
@@ -38,95 +41,70 @@ import org.junit.Ignore
...
+        config = getConfig(defaultConfigType)
+        application = setupPCA(config, defaultChallengeTypes)
+
+        runBlocking {
+            val username = config.email
+            val password = getSafePassword()
+            // Turn an existing username to a non-existing username
+            val alteredUsername = username.replace("@", "1234@")
+            val result = application.signIn(alteredUsername, password.toCharArray())
+            Assert.assertTrue(result is SignInError)
```
diegojerezba: Nit: these `Assert.` are redundant and can be removed, in the same style as the test `testErrorIsInvalidCredentials()` (there are several in the PR)

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
```diff
@@ -56,4 +68,158 @@ class SignUpEmailOTPTest : NativeAuthPublicClientApplicationAbstractTest() {
...
+            Assert.assertTrue((signUpResult as SignUpError).isUserAlreadyExists())
+        }
+    }
+
+    /**
+     * Sign up with email + OTP. User already exists with given email as email-pw account.
+     * (Test case 13)
+     */
+    @Test
+    fun testErrorUserExistAsPassword() {
```
diegojerezba, line 117: This test doesn't belong here, and it already exists in `SignUpEmailPasswordTests` (test 1.1.10) so I think you can remove it.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
```diff
@@ -56,4 +68,158 @@ class SignUpEmailOTPTest : NativeAuthPublicClientApplicationAbstractTest() {
...
+    }
+
+    /**
+     * Sign up with email + OTP. User already exists with given email as social account.
+     * (use case 2.1.7, Test case 14)
+     */
+    @Ignore("TODO: Add social account in the tenant.")
+    @Test
+    fun testErrorUserExistAsSocial() {
+        config = getConfig(ConfigType.SIGN_IN_PASSWORD)
```
diegojerezba: Nit: it doesn't really matter, but this test is defined as an OTP test in the test plan.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
```diff
@@ -38,95 +41,70 @@ import org.junit.Ignore
...
-        Assert.assertTrue((result as SignInError).isUserNotFound())
+    fun testErrorIsUserNotFound() {
+        config = getConfig(defaultConfigType)
+        application = setupPCA(config, defaultChallengeTypes)
+
+        runBlocking {
+            val username = config.email
+            val password = getSafePassword()
+            // Turn an existing username to a non-existing username
+            val alteredUsername = username.replace("@", "1234@")
```
diegojerezba: Nit: Is it possible to use the `INVALID_EMAIL` from `NativeAuthPublicClientApplicationAbstractTest` to keep consistency with other tests? (For example, it is used in `SignUpEmailOTPTest.testErrorInvalidEmailFormat()`)

Same in line 184 with the `alteredPassword`.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
diegojerezba, line 1: Nit: Can you put the tests of this file in order? This helps a lot while reading them for first time.

This is what I mean:

Test 1.1.1
Test 1.1.2
Test 1.1.4
...

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
```diff
@@ -84,19 +73,13 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
...
     }
 
-    abstract val defaultConfigType: ConfigType
-
     @Before
     open fun setup() {
         context = ApplicationProvider.getApplicationContext()
         activity = Mockito.mock(Activity::class.java)
         Mockito.`when`(activity.applicationContext).thenReturn(context)
-        Logger.getInstance().setEnableLogcatLog(true)
```
spetrescu84, line 94: Why were these removed?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
```diff
@@ -93,4 +110,148 @@ class SignUpEmailPasswordTest : NativeAuthPublicClientApplicationAbstractTest()
...
+     * (use case 1.1.10, Test case 28)
+     */
+    @Test
+    fun testErrorUserExistAsPassword() {
+        config = getConfig(ConfigType.SIGN_IN_PASSWORD)
+        application = setupPCA(config, defaultChallengeTypes)
+
+        runBlocking { // Running with runBlocking to avoid default 10 second execution timeout.
+            val user = config.email
+            val signUpResult = application.signUp(user)
```
diegojerezba: Password is missing.

Same in tests 1.1.2, 1.1.11 and 1.1.12

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
```diff
@@ -38,95 +41,70 @@ import org.junit.Ignore
...
-    fun testSuccess() = runTest {
-        val username = config.email
-        val password = getSafePassword()
-        val result = application.signIn(username, password.toCharArray())
-        Assert.assertTrue(result is SignInResult.Complete)
+    fun testSuccess() {
+        config = getConfig(defaultConfigType)
+        application = setupPCA(config, defaultChallengeTypes)
+
+        runBlocking {
```
spetrescu84, line 63: Why do these tests not have retry operation? They are not supposed to be retried if they fail?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
```diff
@@ -93,4 +110,148 @@ class SignUpEmailPasswordTest : NativeAuthPublicClientApplicationAbstractTest()
...
+        config = getConfig(ConfigType.SIGN_IN_PASSWORD)
+        application = setupPCA(config, defaultChallengeTypes)
+
+        runBlocking { // Running with runBlocking to avoid default 10 second execution timeout.
+            val user = tempEmailApi.generateRandomEmailAddress()
+            val signUpResult = application.signUp(user)
+            assertResult<SignUpResult.CodeRequired>(signUpResult)
+            val codeRequiredState = (signUpResult as SignUpResult.CodeRequired).nextState
+            val resendCodeResult = codeRequiredState.resendCode()
+            assertResult<SignUpResendCodeResult.Success>(resendCodeResult)
```
diegojerezba: I know this test is disabled for now, but it should result in userAlreadyExists error.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
```diff
@@ -56,4 +68,158 @@ class SignUpEmailOTPTest : NativeAuthPublicClientApplicationAbstractTest() {
...
+     * Sign up with email + OTP. Developer makes a request with invalid format email address.
+     * (use case 2.1.8, Test case 15)
+     */
+    @Test
+    fun testErrorInvalidEmailFormat() {
+        config = getConfig(defaultConfigType)
+        application = setupPCA(config, defaultChallengeTypes)
+
+        runBlocking { // Running with runBlocking to avoid default 10 second execution timeout.
+            val user = INVALID_EMAIL
```
spetrescu84, line 157: Is this validation from inside SDK or it comes back from API as an error.
This email for iOS would be caught before it even gets sent.
An email such as test@domain would be an invalid one that passes our validation. 
Or is this E2E test supposed to catch internal validation?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
```diff
@@ -56,4 +68,158 @@ class SignUpEmailOTPTest : NativeAuthPublicClientApplicationAbstractTest() {
...
+            val user = tempEmailApi.generateRandomEmailAddress()
+            val signUpResult = application.signUp(user)
+            Assert.assertTrue(signUpResult is SignUpError)
+            Assert.assertTrue((signUpResult as SignUpError).isBrowserRequired())
+        }
+    }
+
+    /**
+     * Sign up with email + OTP. Server requires password authentication, which is supported by the developer.
+     * (hero scenario 11, use case 2.1.11, Test case 17)
```
spetrescu84: This should read "Test case 18"

spetrescu84: Actually not even sure test cases should be there, if you are referring to the ones in the test plan, E2E that were there previously have been removed so order has shifted. Use case is ok and enough to track down I would say.
An example is:
 * (hero scenario 9, use case 1.1.1,  Test case 13)
 Test case 13 is something else now, and Test Case 13 is twice in the code now

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
```diff
@@ -93,4 +110,148 @@ class SignUpEmailPasswordTest : NativeAuthPublicClientApplicationAbstractTest()
...
+     * Sign up with email + password. Developer makes a request with invalid format email address.
+     * (use case 1.1.12, Test case 30)
+     */
+    @Test
+    fun testErrorInvalidEmailFormat() {
+        config = getConfig(defaultConfigType)
+        application = setupPCA(config, defaultChallengeTypes)
+
+        runBlocking { // Running with runBlocking to avoid default 10 second execution timeout.
+            val user = INVALID_EMAIL
```
spetrescu84, line 214: Same thing about INVALID_EMAIL

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
```diff
@@ -93,4 +110,148 @@ class SignUpEmailPasswordTest : NativeAuthPublicClientApplicationAbstractTest()
...
+     * (use case 1.1.13, Test case 31)
+     */
+    @Test
+    fun testErrorInvalidPasswordFormat() {
+        config = getConfig(defaultConfigType)
+        application = setupPCA(config, defaultChallengeTypes)
+
+        runBlocking { // Running with runBlocking to avoid default 10 second execution timeout.
+            val user = tempEmailApi.generateRandomEmailAddress()
+            val password = INVALID_PASSWORD
```
spetrescu84, line 233: nit: This password does not meet reqs because it doesn't contain numbers or special characters. If the reqs of the tenants change this will start failing as a false positive. I would say a short password such as "pass" is more likely to be invalid because it's too short rather not having numbers/characters

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
diegojerezba: @Yuki-YuXin I see that you've implemented this in the test case 2.1.5, but not in the 1.1.2. Can you please implement it there as well? I will resolve this conversation afterwards.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
diegojerezba: OK, but the tests in SignInMFATest are all disabled. Can you enable them?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAnotherAbstractTest.kt
diegojerezba: @nilo-ms @spetrescu84 I'm resolving this conversation as Yuki has merged everything into the same NativeAuthPublicClientApplicationAbstractTest.

Feel free to unresolve it if you think there's something else to do.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
Yuki-YuXin: Because of reason here https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2213#discussion_r1858467717

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
Yuki-YuXin: Because they are email password sign in test case without otp requirement

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
spetrescu84: I was asking about the logging itself, I don't see it happening anywhere else, unless I am wrong? Is the one in the real public client application active now?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
Yuki-YuXin: Done.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
Yuki-YuXin: Since each test case now runs an application, the attached logging will cause conflicts.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
Yuki-YuXin: Done.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
Yuki-YuXin: Done.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
Yuki-YuXin: Done. Use `ConfigType.SIGN_IN_OTP` instead.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
Yuki-YuXin: Android side doesn't have any parameter validation except from the empty string. Thus, it comes back from API as an error.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
Yuki-YuXin: Using "pass" as the invalid password.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
Yuki-YuXin: test 1.1.10 is email+password authentication and this one is using email+otp authentication. 

There is a test case in the Android board there we need to run it manually.
image

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
Yuki-YuXin: Done on both.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
Yuki-YuXin: https://microsofteur-my.sharepoint.com/:w:/g/personal/sodenhoven_microsoft_com/ERrvxU_ZybBIvnq1UfVKWagBG-u_xFS7pcvRbVXCzFHnsQ?e=5EAUlc It was not included in the numbering of the Native Auth Acceptance criteria, so it was added to the Test plan later

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
Yuki-YuXin: Nice catch. Done.

Yuki-YuXin: I agree. Test case encoding is more unreliable and unstable than user case. The original intention was to make it easier to track the progress of the test plan in the future, I will remove the test plan number, please let me know if you think it is inappropriate

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
Yuki-YuXin: My mistake. Update the 1.1.2 as well.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
Yuki-YuXin: Done.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
Yuki-YuXin: Nice catch. Done.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
Yuki-YuXin: Done.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
Yuki-YuXin: Using "email" as the invalid email. Done.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
Yuki-YuXin: Also, according to Silviu's suggestions, I may remove all the Test case number for clarity.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
spetrescu84: Ok, just in case, on IOS all tests are retried 5 times, so if there is a temporary 500 we don't mark the test as failed. Even the ones that don't retry the OTP

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
spetrescu84: I would say remove the test case from all tests, even existing ones.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
spetrescu84: Just in case I am talking about the order here, not Number SO 1.1.3 is ok, 13 is not

spetrescu84 (approved):

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignInEmailPasswordTest.kt
Yuki-YuXin: Well confirmed. I'll stick to avoid retry logic here and try to keep it simple.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailOTPTest.kt
Yuki-YuXin: Done.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/SignUpEmailPasswordTest.kt
Yuki-YuXin: Test plan 13 is a special one that does not exist in our original user case sequence, but was added as a patch to the test plan at a later stage. At the moment I have removed all Test plan numbers and sorted them by user case.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAnotherAbstractTest.kt
diegojerezba: Now that we are back to having one single `NativeAuthPublicClientApplicationAbstractTest` class, the TODO is still there. Do we need it? (I'm just asking because I'm not aware of the context)

Yuki-YuXin: The TODO comment here is to give an idea of using new functionality of Robolectric @LooperMode(PAUSED) instead of @LooperMode(LooperMode.Mode.LEGACY).

Yuki-YuXin: We can do this boost, but I'm not sure if it should be included in this pull request. If I want to solve this TODO, I'll need some extra time and background. And perhaps I should seek guidance from Sammy about this TODO.

diegojerezba: No worries, it was just for my understanding, I'm resolving this comment and will let @nilo-ms handle that decision with you.

diegojerezba (approved):

Yuki-YuXin:
@microsoft-github-policy-service agree company="Microsoft"

Yuki-YuXin:
@microsoft-github-policy-service rerun

PR #2231, opened 2024-12-20
Add support for OneBox Environment, Fixes AB#3113751

shahzaibj:
Fixes [AB#3113751](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3113751)

Adds a new MSAL config file for the OneBox environment. This specifies exact authority that onebox is using. 

Related common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2559

### DIFF
--- a/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/Constants.java
+++ b/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/Constants.java
@@ -56,6 +56,7 @@ public class Constants {
         WEBVIEW_WITH_PPE,
 
         WEBVIEW_PPE_MSA,
+        ONEBOX
     }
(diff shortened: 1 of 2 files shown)


melissaahn (approved):
Are real devices able to make use of this support as well? Or just emulators?

shahzaibj:
> Are real devices able to make use of this support as well? Or just emulators?

Real physical devices can use this too. That's what I've been using. Just need to make sure that it is connected to same network where local eSTS is running. So you may need a proxy or VPN to send traffic from mobile device to PC where eSTS is running.

mohitc1 (approved):
:shipit:

PR #2245, opened 2025-01-15
Move native auth interface towards parameters class methods, Fixes AB#3117217

nilo-ms:
Fixes [AB#3117217](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3117217)

Add parameters class methods to native auth interface. Deprecate parameter list methods.
The signIn with continuation token method has a custom name for Java to avoid ambiguity. Without this custom name the introduction of the new method could introduce a breaking change in case the external developer used null as scopes value.

#### Additional info
The ideal test strategy for these new methods would be wrapping the `CommandDispatcher` in a class and inject it a spy class during tests to check that the new parameters class is used properly. This refactor would require quite some time, so this work will be done in a separate PBI.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
+++ b/msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
@@ -60,6 +60,9 @@ import com.microsoft.identity.common.nativeauth.internal.commands.ResetPasswordS
 import com.microsoft.identity.common.nativeauth.internal.commands.SignInStartCommand
 import com.microsoft.identity.common.nativeauth.internal.commands.SignUpStartCommand
 import com.microsoft.identity.common.nativeauth.internal.controllers.NativeAuthMsalController
+import com.microsoft.identity.nativeauth.parameters.NativeAuthResetPasswordParameters
+import com.microsoft.identity.nativeauth.parameters.NativeAuthSignInParameters
+import com.microsoft.identity.nativeauth.parameters.NativeAuthSignUpParameters
 import com.microsoft.identity.nativeauth.statemachine.errors.ErrorTypes
 import com.microsoft.identity.nativeauth.statemachine.errors.GetAccountError
 import com.microsoft.identity.nativeauth.statemachine.errors.ResetPasswordError
@@ -292,6 +295,7 @@ class NativeAuthPublicClientApplication(
      * @param callback [com.microsoft.identity.nativeauth.NativeAuthPublicClientApplication.SignInCallback] to receive the result.
      * @return [com.microsoft.identity.nativeauth.statemachine.results.SignInResult] see detailed possible return state under the object.
      */
+    @Deprecated("This method is now deprecated. Use the method 'signIn(parameters:, callback:)' instead.")
     override fun signIn(
         username: String,
         password: CharArray?,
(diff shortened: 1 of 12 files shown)


msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -333,6 +362,251 @@ class NativeAuthPublicClientApplication(
...
+            override fun onError(exception: BaseException) {
+                Logger.error(TAG, "Exception thrown in checkForPersistedAccount", exception)
+                future.setException(exception)
+            }
+        })
+
+        return future
+    }
+
+    private suspend fun internalSignIn(
```
spetrescu84, line 605: Are these ok to be in the same class or should they be in extension?

msal/src/main/java/com/microsoft/identity/nativeauth/parameters/NativeAuthSignInParameters.kt
```diff
@@ -0,0 +1,46 @@
...
+//  THE SOFTWARE.
+
+package com.microsoft.identity.nativeauth.parameters
+
+/**
+ * Encapsulates the parameters passed to the signIn methods of NativeAuthPublicClientApplication
+ */
+class NativeAuthSignInParameters(
+    /**
+     * username of the account to sign in
```
spetrescu84: Missing dot at the end, should be 

`* username of the account to sign in.`

msal/src/main/java/com/microsoft/identity/nativeauth/parameters/NativeAuthSignUpParameters.kt
```diff
@@ -0,0 +1,47 @@
...
+package com.microsoft.identity.nativeauth.parameters
+
+import com.microsoft.identity.nativeauth.UserAttributes
+
+/**
+ * Encapsulates the parameters passed to the signUp methods of NativeAuthPublicClientApplication
+ */
+class NativeAuthSignUpParameters(
+    /**
+     * username of the account to sign up
```
spetrescu84: Missing dot at the end, should be

`* username of the account to sign in.`

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
nilo-ms: On the Android project I noticed it is not common practice using extension as we did on iOS, so I didn't do it

msal/src/main/java/com/microsoft/identity/nativeauth/parameters/NativeAuthSignUpParameters.kt
nilo-ms: Thanks, message updated

msal/src/main/java/com/microsoft/identity/nativeauth/parameters/NativeAuthSignInParameters.kt
nilo-ms: Thanks, message updated

Yuki-YuXin (approved):

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
```diff
@@ -858,6 +973,73 @@ public void testGetAccessTokenWithSignInScopes() throws ExecutionException, Inte
...
+        params.setPassword(password);
+
+        application.signIn(params, signInTestCallback);
+        SignInResult signInWithPasswordResult = signInTestCallback.get();
+        assertTrue(signInWithPasswordResult instanceof SignInResult.Complete);
+
+        // Get access token from sign in result
+        GetAccessTokenTestCallback getAccessTokenCallback = new GetAccessTokenTestCallback();
+
+        NativeAuthGetAccessTokenParameters getAccessTokenParameters = new NativeAuthGetAccessTokenParameters();
```
spetrescu84, line 1012: Should we declare `NativeAuthGetAccessTokenParameters` like this:


```
class NativeAuthGetAccessTokenParameters(
    var forceRefresh: Boolean = false,
    var scopes: List<String>? = null
)
```

So it can be used like this?

`val getAccessTokenParameters = NativeAuthGetAccessTokenParameters(forceRefresh = false)`

In the tests it's not a big problem, but I'm thinking in user's code it's way more easy to read like this IMO.

The same would apply for the rest of the declarations.

nilo-ms: This type of constructor is not extensible for Java users, because this language doesn't support default parameters value. This means that if we need to add an optional parameter to the constructor, it will result in a breaking change for the external developer. This is why constructor should only contain mandatory parameters.

spetrescu84: Ah, ok, in that case we can go with the current ones.

spetrescu84 (approved):

Yuki-YuXin:
One thing to mention, please update the change log file for the changes. Otherwise, use the tag "No-ChangeLog"

For this PR, there might need to update the change log.

PR #2118, opened 2024-06-10
Fix msal test app after merging with the fadi/release/5.4.0

Yuki-YuXin:
Fix update according to the following team feedback:
**Batch2**
- [hero scenario 6, use case 2.2.1] Use email and OTP to get token and sign in
getParcelable(Constants.STATE) 
- [use case 2.2.6] Ability to provide scope to control auth strength of the token
Update the scope of profile to User.Read 
Update the test app EmailSignInSignUp page: signIn(scopes = User.Read)
- [use case 1.2.2] User is not registered with given email
Update test app EmailPasswordSignInSignUp page: displayDialog("Unexpected result", actionResult.errorMessage) in sign in function
- [use case 1.2.6] Ability to provide scope to control auth strength of the token
Update the test app EmailPasswordSignInSignUp page: signIn(scopes = User.Read)
Update test app EmailPasswordSignInSignUp page: displayDialog("Unexpected result", actionResult.errorMessage) in sign in function
- [hero scenario 2, use case 2.1.2] Signup user with custom attributes with verify OTP as last step
Update the test app SignUpCode page: currentState = (bundle?.getParcelable(Constants.STATE) as? SignUpCodeRequiredState)!!
- [hero scenario 2, use case 2.1.2] Signup user with custom attributes with verify OTP as last step
Update the test app SignUpCode page: currentState = (bundle?.getParcelable(Constants.STATE) as? SignUpCodeRequiredState)!!
- [hero scenario 2, use case 2.1.2] Signup user with custom attributes with verify OTP as last step

- [hero scenario 3, use case 2.1.3] Verify email OTP first and then collect custom attributes
Update the test app SignUpAttributes page: currentState = (bundle?.getParcelable(Constants.STATE) as? SignUpAttributesRequiredState)!!
SignUpAttributesUI disorder

- [use case 1.1.4] Verify email address using email OTP and then set password 
In SignUpCodeFragment modify to currentState = (bundle?.getParcelable(Constants.STATE) as? SignUpCodeRequiredState)!!
In SignUpPasswordFragment modify to currentState = (bundle?.getParcelable(Constants.STATE) as? SignUpPasswordRequiredState)!!

**Batch1**
- [use case 1.2.2] User is not registered with given email
No error is popped up in UI (Probably this is a bug of the test app)
- [use case 2.2.5] Resend email OTP
App crashing
- [use case 1.2.3] Password is incorrect
No error is popped up in UI (Probably this is a bug of the test app)
[...]

PR #2001, opened 2024-01-10
Add native auth to MSAL test app

SammyO:
### For MSAL team:
In order to let the native auth team execute (manual) release tests, an application is needed that supports the various use flows that are part of the release tests.

Changes:
- The existing MSAL test app is extended with a Native Auth section, which is accessed from the side menu.
- Updated CODEOWNERS to reflect native auth ownership of the newly added classes.

### For native auth team:
Changes compared to previous versions of this application:
- Sign up attributes are no longer hardcoded to `country` and `city`, but taken dynamically from the UI (through new key and value text fields).
- AttributesRequired` state will navigate to the new, generic AttributesFragment. Logs need to be checked for details about which attributes are required, and thus which keys need to be supplied.
- `CodeFragment` displays API's `channel`, `sentTo` and `codeLength` fields, which can now be used for validation for correct behaviour.

Note:
This application is not in a state where it can be used as a CI-generated artefact by the native auth team. Developers using this application will need to change the config JSON file and update it to one of several tenant configurations, recompile and run it. Support for native auth in the Labs environment is needed to make this application usable as a CI-generated artefact. Once Labs support is added, the native auth screens will need config selection UI, similar to the Fragment dedicated to testing `acquireToken()`.

PR #2248, opened 2025-01-23
Remove native auth from MSAL test app

Yuki-YuXin:
## Introduction

Native auth adopted E2E automation test without any UI. Remove/Revert previous changes to MSAL test app.

## Reference
The previous merged related PR, especially the first one.
https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2001
https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2143
https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2118


shahzaibj (approved):
cc/ @fadidurah

fadidurah (approved):

nilo-ms (changes requested):
This app was added not for E2E tests but to let the native auth team execute (manual) release tests. We didn't automate all manual release tests in E2E, so we still need to use this application. @Yuki-YuXin can you please clarify why we're removing this app?

Yuki-YuXin:
@nilo-ms This part of the code was created before the idea of E2E automation test was born and implemented, during that time all native auth tests relied on manual testing, so we copied a sample app with some changes and put it into the MSAL test app to make it a manual test environment. 

For the reason below:
1. At the moment, E2E automation test covers >80% of our test cases and the remaining <20% can be done entirely using Sample app. Test app are more than 90% replicas of Sample app and the differentiation(attributes) has been covered by automation test.

2. At that time(2024,Jan), the idea of paving the way for the subsequent automation test existed in the Test app, but after the practice of the E2E automation test, we found that the UI is not necessary for the test of SDK reliability. 

3. No one is maintaining this portion of the Test app code like we did for Sample app (Reference: The last maintenance was in 2024,Jun). Since then we added MFA feature and made MFA private preview release using E2E automation tests without adding or updating Test app for manual test cases. Test app is not compatible/behind with the latest MSAL Native Auth section. In other words, man-made long-term bugs.

4. In practice, release engineers use Sample app more than Test app because they are separate repository, easy testing by switching branches. However, the Test app is embedded in the MSAL package. Each modification to the Test App is equivalent to a modification to the MSAL package. This partial redundancy is equal to the redundancy of the MSAL package.

If you hold another view on this, I'd be happy to bring it up as a topic in our panel discussion on Thursday.

PR #2262, opened 2025-03-06
Melissaahn/release/5.10.2

melissaahn:
Note: the files had existing versions of "20.1.0" and "5.10.1" because release/5.10.2 was created from release/5.10.1, which is a scrapped version due to some issues which occurred during the release process.

See here that release/5.10.2 has no differences from dev other than the version numbers: https://github.com/AzureAD/microsoft-authentication-library-for-android/compare/dev...release/5.10.2

### DIFF
--- a/msal/build.gradle
+++ b/msal/build.gradle
@@ -188,12 +188,12 @@ task sourcesJar(type: Jar) {
 
 // In dev, we want to keep the dependencies (common4j, common) to 1.0.+ to be able to be consumed by daily dev pipeline.
 // In release/*, we change these to specific versions being consumed.
-def commonVersion = "20.1.0"
+def commonVersion = "20.1.1-RC1"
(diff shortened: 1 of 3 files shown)


msal/build.gradle
```diff
@@ -188,12 +188,12 @@ task sourcesJar(type: Jar) {
 
 // In dev, we want to keep the dependencies (common4j, common) to 1.0.+ to be able to be consumed by daily dev pipeline.
 // In release/*, we change these to specific versions being consumed.
-def commonVersion = "20.1.0"
+def commonVersion = "20.1.1-RC3"
```
shahzaibj: Is the RC tag still supposed to be here?

melissaahn: I missed this; thank you. Updated.

shahzaibj (approved):

mohitc1 (approved):

PR #2278, opened 2025-04-09
Native auth: Add support for claims request in getAccessToken and signIn after signUp/SSPR, Fixes AB#3188133

nilo-ms:
Adding support on native authentication for claims request for getAccessToken and signIn after signUp/SSPR flows.

MSAL common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2622

[AB#3188133](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3188133)

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
@@ -532,6 +532,7 @@ public class CommandParametersAdapter {
      * @param username email address of the user
      * @param password password of the user
      * @param scopes scopes requested during sign in flow
+     * @param claimsRequest claims request object. Nullable object
      * @return Command parameter object
      * @throws ClientException
      */
@@ -587,6 +588,7 @@ public class CommandParametersAdapter {
      * @param username email address of the user
      * @param correlationId correlation ID to use in the API request, taken from the previous API response in the flow
      * @param scopes scopes requested during sign in flow
+     * @param claimsRequest claims request object. Nullable object
      * @return Command parameter object
      * @throws ClientException
      */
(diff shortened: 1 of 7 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -621,6 +625,7 @@ public static SignInWithContinuationTokenCommandParameters createSignInWithConti
                 .username(username)
                 .challengeType(configuration.getChallengeTypes())
                 .authenticationScheme(authenticationScheme)
+                .claimsRequestJson(claimsRequestJson)
```
spetrescu84, line 628: From my understanding because in Android there is only one path for SignIn, there is no need to declare it anywhere else as it's implemented down the chain. Is this correct?

spetrescu84 (approved):

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
nilo-ms: Sorry Silviu, I didn't get the question

spetrescu84: On iOS we have to pass the `claimsRequestJson` on each of the 3 flows - SignIn/SingInAfterSignUp and SignInAfterResetPassword, but on Android the constructor for all is the same and we are not missing anything. Is this correct?

spetrescu84 (approved):

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
nilo-ms: Yes, correct

Yuki-YuXin (approved):
Please update the common reference to pass the pipeline check.

spetrescu84 (approved):

PR #2304, opened 2025-05-30
Handle redirect across all endpoints and return the redirect reason to the developers, Fixes AB#3288191

Yuki-YuXin:
Fixes [AB#3288191](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3288191)

Related common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2660

Pipeline failed: SignInStates.kt: (294, 51): Unresolved reference: redirectReason - common reference

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
+++ b/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
@@ -175,7 +175,7 @@ class ResendCodeError(
     override val correlationId: String,
     override val errorCodes: List<Int>? = null,
     override var exception: Exception? = null
-): SignInResendCodeResult, SignUpResendCodeResult, ResetPasswordResendCodeResult, Error(errorType = errorType, error = error, errorMessage= errorMessage, correlationId = correlationId, errorCodes = errorCodes, exception = exception)
+): BrowserRequiredError, SignInResendCodeResult, SignUpResendCodeResult, ResetPasswordResendCodeResult, Error(errorType = errorType, error = error, errorMessage= errorMessage, correlationId = correlationId, errorCodes = errorCodes, exception = exception)
 
 class GetAccountError(
     override val errorType: String? = null,
--- a/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/JITStates.kt
+++ b/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/JITStates.kt
@@ -122,6 +122,15 @@ abstract class BaseJITSubmitChallengeState(
                         )
                     )
                 }
+                is INativeAuthCommandResult.Redirect -> {
+                    RegisterStrongAuthChallengeError(
+                        errorType = ErrorTypes.BROWSER_REQUIRED,
+                        error = result.error,
+                        errorMessage = result.errorDescription,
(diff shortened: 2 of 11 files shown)


msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
```diff
@@ -175,7 +175,7 @@ class ResendCodeError(
     override val correlationId: String,
     override val errorCodes: List<Int>? = null,
     override var exception: Exception? = null
-): SignInResendCodeResult, SignUpResendCodeResult, ResetPasswordResendCodeResult, Error(errorType = errorType, error = error, errorMessage= errorMessage, correlationId = correlationId, errorCodes = errorCodes, exception = exception)
+): BrowserRequiredError, SignInResendCodeResult, SignUpResendCodeResult, ResetPasswordResendCodeResult, Error(errorType = errorType, error = error, errorMessage= errorMessage, correlationId = correlationId, errorCodes = errorCodes, exception = exception)
```
spetrescu84, line 178: This is not a breaking change for developers is it?

Yuki-YuXin: No, it isn't.  `BrowserRequiredError` is an interface which has `isBrowserRequired()` method. From the client (sample), there is a general `ResendCodeError` branch `when(result)`.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
```diff
@@ -175,7 +175,7 @@ class ResendCodeError(
...
 
 class GetAccountError(
```
nilo-ms, line 180: Does "GetAccountError" has "BrowserRequiredError" ?

nilo-ms (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/JITStates.kt
```diff
@@ -122,6 +122,15 @@ abstract class BaseJITSubmitChallengeState(
                         )
                     )
                 }
+                is INativeAuthCommandResult.Redirect -> {
+                    RegisterStrongAuthChallengeError(
+                        errorType = ErrorTypes.BROWSER_REQUIRED,
+                        error = result.error,
+                        errorMessage = result.redirectReason,
```
nilo-ms, line 129: general comment here: What happens when redirectReason is empty or null? On iOS, we use a default error message to instruct the user on the next step, which I find useful. Do we have a similar approach on Android? Interested to know what @Yuki-YuXin and @spetrescu84 think about this

Yuki-YuXin: In the network layer, the redirectReason could return null. When is null, inside `ApiResponse`, `@SerializedName("redirect_reason") val redirectReason: String?,` will set a default null, however, `redirectReason.orEmpty()` will make it an empty string.

Yuki-YuXin: Continue the discussion under the common repo PR.

spetrescu84 (approved):

PR #2331, opened 2025-07-09
[Android SDK] Change Native Auth E2E tests variables, Fixes AB#3320709

spetrescu84:
This PR introduces the following:
-Add NATIVE_AUTH_CONFIG_STRING to Pipeline variables and use it in scripts to run E2E tests instead of the KeyVault
-Add NATIVE_AUTH_CONFIG_FILE_PATH to be used locally from build settings and use it in scripts to run E2E tests
This PR is connected to the MSAL Android Common PR [[Android SDK] Change Native Auth E2E tests variables, Fixes AB#3320709](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2708)

### DIFF
--- a/msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
+++ b/msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
@@ -185,7 +185,8 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
                 }
             }
         } catch (e: IOException) {
-            e.printStackTrace()
+            android.util.Log.e("NativeAuthTest", "Failed to read config file: $filePath", e)
+            throw RuntimeException("Error reading config file: $filePath", e)
         }
         return sb.toString()
     }
--- a/msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
+++ b/msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
@@ -103,15 +103,15 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
     }
 
     private fun getConfigsThroughBuildValue(): Map<String, NativeAuthTestConfig.Config>? {
-        // We are no longer fetching the config from the Key Vault, we are using a BuildConfig value
-        // val secretValue = KeyVaultFetchHelper.getSecretForBuildAutomation("msalandroidnativeauthautomationconfjsonfile")
         var buildConfigString = BuildValues.getNativeAuthConfigString()
         // If the buildConfigString is null or empty, we will try to read the config from the file path
-        if (buildConfigString == null || buildConfigString.isEmpty()) {
+        if (buildConfigString.isNullOrBlank()) {
             val buildConfigFilePath = BuildValues.getNativeAuthConfigFilePath()
             // If the build config file path is set, read the file and set buildConfigString to its content
             if (buildConfigFilePath != null && buildConfigFilePath.isNotEmpty()) {
                 buildConfigString = readConfigFile(buildConfigFilePath)
+            } else {
+                throw IllegalStateException("Native auth config file pipeline variable or local file path not found")
             }
         }
         val type = TypeToken.getParameterized(
--- a/msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
+++ b/msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
@@ -32,8 +32,8 @@ import com.microsoft.identity.client.e2e.shadows.ShadowAndroidSdkStorageEncrypti
 import com.microsoft.identity.client.e2e.tests.IPublicClientApplicationTest
 import com.microsoft.identity.client.exception.MsalException
 import com.microsoft.identity.common.internal.controllers.CommandDispatcherHelper
+import com.microsoft.identity.common.java.nativeauth.BuildValues
 import com.microsoft.identity.internal.testutils.TestUtils
-import com.microsoft.identity.internal.testutils.labutils.KeyVaultFetchHelper
 import com.microsoft.identity.internal.testutils.labutils.LabConstants
 import com.microsoft.identity.internal.testutils.labutils.LabUserHelper
 import com.microsoft.identity.internal.testutils.labutils.LabUserQuery
@@ -52,6 +52,10 @@ import org.mockito.Mockito
 import org.robolectric.RobolectricTestRunner
 import org.robolectric.annotation.Config
 import org.robolectric.annotation.LooperMode
+import java.io.BufferedReader
+import java.io.FileInputStream
+import java.io.IOException
+import java.io.InputStreamReader
 
 // TODO: move to "PAUSED". A work in RoboTestUtils will be needed though.
 @LooperMode(LooperMode.Mode.LEGACY)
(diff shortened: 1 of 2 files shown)


msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
```diff
@@ -98,19 +102,29 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
         return credential.password
     }
 
-    private fun getConfigsThroughSecretValue(): Map<String, NativeAuthTestConfig.Config>? {
-        val secretValue = KeyVaultFetchHelper.getSecretForBuildAutomation("msalandroidnativeauthautomationconfjsonfile")
+    private fun getConfigsThroughBuildValue(): Map<String, NativeAuthTestConfig.Config>? {
+        // We are no longer fetching the config from the Key Vault, we are using a BuildConfig value
+        // val secretValue = KeyVaultFetchHelper.getSecretForBuildAutomation("msalandroidnativeauthautomationconfjsonfile")
```
nilo-ms: I suppose this is here by mistake

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
```diff
@@ -98,19 +102,29 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
         return credential.password
     }
 
-    private fun getConfigsThroughSecretValue(): Map<String, NativeAuthTestConfig.Config>? {
-        val secretValue = KeyVaultFetchHelper.getSecretForBuildAutomation("msalandroidnativeauthautomationconfjsonfile")
+    private fun getConfigsThroughBuildValue(): Map<String, NativeAuthTestConfig.Config>? {
+        // We are no longer fetching the config from the Key Vault, we are using a BuildConfig value
```
nilo-ms: I don't think we need to say what was the previous config source. I suggest removing this comment

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
```diff
@@ -98,19 +102,29 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
...
+        // val secretValue = KeyVaultFetchHelper.getSecretForBuildAutomation("msalandroidnativeauthautomationconfjsonfile")
+        var buildConfigString = BuildValues.getNativeAuthConfigString()
+        // If the buildConfigString is null or empty, we will try to read the config from the file path
+        if (buildConfigString == null || buildConfigString.isEmpty()) {
```
nilo-ms: nit: you can use "isNullOrBlank" method

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
```diff
@@ -98,19 +102,29 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
...
+            val buildConfigFilePath = BuildValues.getNativeAuthConfigFilePath()
+            // If the build config file path is set, read the file and set buildConfigString to its content
+            if (buildConfigFilePath != null && buildConfigFilePath.isNotEmpty()) {
+                buildConfigString = readConfigFile(buildConfigFilePath)
+            }
```
nilo-ms, line 115: What happen if the build config file path is not set? Can you throw a meaningful error and log what is missing? This will make debug easier.
Something like:
"Native auth config file pipeline variable or local file path not found."

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
```diff
@@ -98,19 +102,29 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
...
             NativeAuthTestConfig.Config::class.java
         ).type
 
-        return Gson().fromJson(secretValue, type)
+        return Gson().fromJson(buildConfigString, type)
     }
 
     fun getConfig(configType: ConfigType): NativeAuthTestConfig.Config {
-        val secretValue = getConfigsThroughSecretValue()
+        val secretValue = getConfigsThroughBuildValue()
```
nilo-ms, line 127: How often this "getConfig" is called? If it is called more than once would be more efficient to call the function "getConfigsThroughBuildValue" just once and parse the JSON file only the first time

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
```diff
@@ -157,4 +171,22 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
...
+                BufferedReader(InputStreamReader(inputStream)).use { reader ->
+                    var line: String? = reader.readLine()
+                    while (line != null) {
+                        sb.append(line)
+                        line = reader.readLine()
+                    }
+                }
+            }
+        } catch (e: IOException) {
+            e.printStackTrace()
```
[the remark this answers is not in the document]
nilo-ms: Actually copilot is not wrong here @spetrescu84

Yuki-YuXin (approved):

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
spetrescu84: Removed, left it to clarify when reviewing PR what happens.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
spetrescu84: By looking at tests, normally it's once/test so updating it would not help if we would read on setup for example.
There are also tests such as `testErrorOutOfPersistenceDifferentAccount` which first call `getConfig(defaultConfigType)` (`defaultConfigType` is `ConfigType.SIGN_IN_PASSWORD`) here and then call `getConfig(ConfigType.SIGN_IN_OTP)` which means we either read for both from start or there would be errors.
I would say to leave as is for this one.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
spetrescu84: Done

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
spetrescu84: If you mean the comment, I left it only for PR review, was planning to remove

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
nilo-ms: More that caching the result of `getConfig(configType: ConfigType)` I was suggesting caching the `getConfigsThroughBuildValue` result, that contains the whole JSON file, which is always the same. If this doesn't noticeably slow down test execution, I think it's fine to keep it as is.

nilo-ms (approved):

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
```diff
@@ -98,19 +102,29 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
         return credential.password
     }
 
-    private fun getConfigsThroughSecretValue(): Map<String, NativeAuthTestConfig.Config>? {
-        val secretValue = KeyVaultFetchHelper.getSecretForBuildAutomation("msalandroidnativeauthautomationconfjsonfile")
+    private fun getConfigsThroughBuildValue(): Map<String, NativeAuthTestConfig.Config>? {
```
mustafamizrak, line 105: Just wondering, could we use profiles for this purpose? Like local/dev, kv, prod, test etc to specify the purpose?

spetrescu84: I have looked into this originally but the only way to distinguish between pipelines and local dev is to have a flag such as `-Dprofile=local` or `-Dprofile=pipeline`, but we would need also the other flags. From what I see, Android Studio doesn't have inbuilt profiles for pipeline runs, but I could be wrong. We could use a string locally also, but I went for the file option as that's how it works on iOS.
The file can't be pushed to github so adding it to the pipeline doesn't work.

fadidurah (approved):

PR #2361, opened 2025-08-22
Add null checks for guest account ids, Fixes AB#3330213

melissaahn:
### Summary
[AB#3330213](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3330213)

We got a report from a partner team that they were seeing very low amounts (like 5 instances) of an NPE related to AccountAdapter. Looking at past related issues, this seems like some account entry data got malformed for particular accounts/devices, perhaps due to very old broker versions. The patch in this PR is to add null checks for both the home and local id from the guest account, and if either are null, then it won't be added to the guest account array.

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
@@ -60,7 +60,9 @@ public class AccountAdapter {
                 final String acctHomeAccountId = cacheRecord.getAccount().getHomeAccountId();
                 final String acctLocalAccountId = cacheRecord.getAccount().getLocalAccountId();
 
-                if (!acctHomeAccountId.contains(acctLocalAccountId)) {
+                if (acctHomeAccountId != null
+                        && acctLocalAccountId != null
+                        && !acctHomeAccountId.contains(acctLocalAccountId)) {
                     result.add(cacheRecord);
                 }
             }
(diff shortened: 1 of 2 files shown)


fadidurah (approved):

somalaya (approved):

msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java
```diff
@@ -60,7 +60,9 @@ public List<ICacheRecord> filter(@NonNull List<ICacheRecord> records) {
                 final String acctHomeAccountId = cacheRecord.getAccount().getHomeAccountId();
                 final String acctLocalAccountId = cacheRecord.getAccount().getLocalAccountId();
 
-                if (!acctHomeAccountId.contains(acctLocalAccountId)) {
+                if (acctHomeAccountId != null
+                        && acctLocalAccountId != null
+                        && !acctHomeAccountId.contains(acctLocalAccountId)) {
                     result.add(cacheRecord);
```
somalaya, line 74: You may also want to add a warning log to say that these values are null. Ideally, they should not be null.

melissaahn: Took suggestion.

somalaya (approved):

fadidurah (approved):

PR #2379, opened 2025-09-12
Native auth: Update Email OTP MFA to Match EC Implementation, Fixes AB#3351233

nilo-ms:
This PR updates the SDK to match the latest flow from EC.

In this new flow, the developer must always supply an auth Method to the /oauth2/v2.0/challenge endpoint which means once the .mfaRequired error is received from token endpoint, the /oauth2/v2.0/introspect endpoint needs to be called to retrieve the methods which are automatically returned to the external developer.

Furthermore, whenever calling the /token endpoint is called with an MFA Email OTP code, the grant type should be mfa_oob

Fixes [AB#3351233](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3351233)

MSAL Common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2760

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
@@ -62,8 +62,7 @@ import com.microsoft.identity.common.java.dto.AccountRecord;
 import com.microsoft.identity.common.java.exception.ClientException;
 import com.microsoft.identity.common.java.nativeauth.authorities.NativeAuthCIAMAuthority;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.GetAuthMethodsCommandParameters;
-import com.microsoft.identity.common.java.nativeauth.commands.parameters.MFADefaultChallengeCommandParameters;
-import com.microsoft.identity.common.java.nativeauth.commands.parameters.MFASelectedDefaultChallengeCommandParameters;
+import com.microsoft.identity.common.java.nativeauth.commands.parameters.MFAChallengeAuthMethodCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.MFASubmitChallengeCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.ResetPasswordResendCodeCommandParameters;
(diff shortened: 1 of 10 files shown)


msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -836,7 +786,7 @@ public static MFADefaultChallengeCommandParameters createMFADefaultChallengeComm
      * @return Command parameter object
      * @throws ClientException
      */
-    public static MFASelectedDefaultChallengeCommandParameters createMFASelectedChallengeCommandParameters(
+    public static MFAChallengeAuthMethodCommandParameters createMFASelectedChallengeCommandParameters(
```
shenj: should this method name be changed as well?

shenj (approved):

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
```diff
@@ -511,7 +510,7 @@ public void onError(@NonNull BaseException exception) {
     }
 
     @Test
-    public void testSignInMFAScenario1() throws ExecutionException, InterruptedException, TimeoutException {
+    public void testSignInMFAScenario2() throws ExecutionException, InterruptedException, TimeoutException {
```
spetrescu84, line 513: Why did this change to 2? Where is 1

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
nilo-ms: Yes, let me change it

msal/src/test/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplicationJavaTest.java
nilo-ms: This was originally MFA scenario 2 and I removed MFA scenario 1. I don't know why the diff let it look like I renamed the name of the test

spetrescu84 (approved):

PR #2384, opened 2025-09-25
Native auth: add new SDK error when auth method is blocked, Fixes AB#3313635

nilo-ms:
This PR adds a new SDK error with instructions when the auth method is blocked during MFA flow.

MSAL Common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2772

Fixes [AB#3313635](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3313635)

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
+++ b/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
@@ -87,6 +87,12 @@ internal class ErrorTypes {
         const val VERIFICATION_CONTACT_BLOCKED = "verification_contact_blocked"
 
         /*
+         * The AUTH_METHOD_BLOCKED value indicates that the server blocked the strong authentication method.
+         * Try contacting customer support to seek assistance.
+         */
+        const val AUTH_METHOD_BLOCKED = "auth_method_blocked"
+
+        /*
          * The INVALID_STATE value indicates a misconfigured or expired state, or an internal error
          * in state transitions. If this occurs, the flow should be restarted.
          */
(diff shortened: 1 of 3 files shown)


andwhysoft (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
```diff
@@ -86,6 +86,12 @@ internal class ErrorTypes {
          */
         const val VERIFICATION_CONTACT_BLOCKED = "verification_contact_blocked"
 
+        /*
+         * The AUTH_METHOD_BLOCKED value indicates that the server blocked the strong authentication method.
```
spetrescu84, line 90: This is for both MFA and Register Strong Auth right? Would it be more clear to say 
`The AUTH_METHOD_BLOCKED value indicates that the server blocked the strong authentication method during registration or MFA.` or something similar? Or it's clear during MFA that a strong authentication method is used?

spetrescu84 (approved):

mustafamizrak (approved):

spetrescu84 (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/Error.kt
nilo-ms: This error is only for MFA, when the auth method is actually strong (because has been already registered). For JIT the error id is VERIFICATION_CONTACT_BLOCKED. I used two different error id because the user needs to act differently for this error.

PR #2393, opened 2025-10-23
Add WebAuthn version support in configuration, Fixes AB#3385532

p3dr0rv:
[AB#3385532](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3385532)
https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/20357
### Add WebAuthn Version Support and Passkey Headers

This PR adds support for handling the WebAuthn protocol version in the app configuration and authentication flow for broker-less scenarios. also enables testing on WEBVIEW PPE MSA

**Changes:**
- Added a new `webauthn_version` field to `PublicClientApplicationConfiguration`, including serialization, accessors, and merge logic, allowing apps to define and retrieve the WebAuthn version from configuration files.  
- Updated `CommandParametersAdapter` to include passkey protocol headers in authentication requests when WebAuthn is enabled, supported (Android 9+), Authorization agent is Webview and the version is 1.1.  
- Updated the test app (`MsalWrapper`) to append the `msaoauth2=true` parameter to query strings when running in the pre-production environment with WebAuthn 1.1 enabled, enabling proper testing of WebAuthn flows.

Related PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2769

Test
1- create account https://signup.live-int.com/?lic=1
2 - Install msal test app, (ensure no broker is installed)
3 - change config to MSA_WEBVIEW_PPE
4- Click acquire token and complete auth flow (username, password)
5 - User is presented with the option to register a passkey, complete the flow, and you will end up with a token and a passkey.
6 - try again with no user selected and use the passkey.


gradle/versions.gradle
```diff
@@ -80,6 +80,7 @@ ext {
     AndroidCredentialsVersion="1.2.2"
     LegacyFidoApiVersion="20.1.0"
     GoogleIdVersion="1.1.0"
```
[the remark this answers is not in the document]
p3dr0rv: Used when building common submodule

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -203,6 +206,7 @@ public static InteractiveTokenCommandParameters createInteractiveTokenCommandPar
                 .powerOptCheckEnabled(configuration.isPowerOptCheckForEnabled())
                 .correlationId(parameters.getCorrelationId())
                 .preferredAuthMethod(parameters.getPreferredAuthMethod())
+                .requestHeaders(addPasskeyHeader(parameters.getExtraQueryStringParameters(), configuration))
```
melissaahn, line 210: kinda of a nit: I think parameters.getExtraQueryStringParameters() is going to be returning the original set of query string paramters and not the updated one which would be generated by appendToExtraQueryParametersIfWebAuthnCapable(parameters.getExtraQueryStringParameters(), configuration). I think based on the logic below, it potentially would be possible that the header would be added when it shouldn't be (like in the case where Android version <= 9 and the qs param should be removed.

Maybe call the appendToExtraQueryParameterIfWebAuthnCapable method earlier and save it to a variable?

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -1366,4 +1370,70 @@ public static List<Map.Entry<String, String>> appendToExtraQueryParametersIfWebA
...
+
+
+        // Skip if no webauthn query parameter and the configuration isn't webauthn-capable
+        if (!containsQueryParameter(queryStringParameters, FidoConstants.WEBAUTHN_QUERY_PARAMETER_FIELD)
+                && !configuration.isWebauthnCapable()) {
+            return headers;
+        }
+
+        // Add header only if version is 1.1,
+        // By default, version is 1.0 and no header is added.
```
melissaahn: I think I sort of understand this comment (the 1.0 header would get added in the common code) but without that context this wouldn't make as much sense to me. Can you please put some more details in this comment?

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -1366,4 +1370,70 @@ public static List<Map.Entry<String, String>> appendToExtraQueryParametersIfWebA
...
+    private static boolean containsQueryParameter(
+            @Nullable final List<Map.Entry<String, String>> queryParameters,
+            @NonNull final String queryParameter) {
+
+        if (queryParameters == null || queryParameters.isEmpty()) {
+            return false;
+        }
+
+        for (Map.Entry<String, String> entry : queryParameters) {
+            if (queryParameter.equals(entry.getKey())) {
```
melissaahn: A nit: technically the webauthn query string param key needs to be present AND have a value of 1 (webauthn=1). Unless that logic has changed since I last worked on this. 
Though idk who would be sending something like webauthn=0, maybe either this helper methods should be updated to consider both key and value, or the logic which called it above should also check for key.

melissaahn (approved):

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
p3dr0rv: reworked

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
p3dr0rv: I created the function with this in mind, which is why I added the API version check. I want to avoid calling `appendToExtraQueryParameterIfWebAuthnCapable` here, as it could be confusing IMO.

testapps/testapp/src/main/res/raw/msal_config_webview_ppe_msa.json
```diff
@@ -3,6 +3,7 @@
   "authorization_user_agent" : "WEBVIEW",
   "redirect_uri" : "msauth://com.msft.identity.client.sample.local/1wIqXSqBj7w%2Bh11ZifsnqwgyKrY%3D",
   "webauthn_capable" : true,
+  "webauthn_version" : "1.1",
```
fadidurah, line 6: Should we update the default configuration file to have some value for this?

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -1371,4 +1375,81 @@ public static List<Map.Entry<String, String>> appendToExtraQueryParametersIfWebA
...
+
+        switch (configuration.getWebauthnVersion()) {
+            case FidoConstants.PASSKEY_PROTOCOL_VERSION_1_0:
+                headers.put(FidoConstants.PASSKEY_PROTOCOL_HEADER_NAME, FidoConstants.PASSKEY_PROTOCOL_HEADER_AUTH_AND_REG);
+                Logger.verbose(methodTag, "Passkey header added for WebAuthn version 1.0");
+                break;
+            case FidoConstants.PASSKEY_PROTOCOL_VERSION_1_1:
+                headers.put(FidoConstants.PASSKEY_PROTOCOL_HEADER_NAME, FidoConstants.PASSKEY_PROTOCOL_HEADER_AUTH_AND_REG);
+                Logger.verbose(methodTag, "Passkey header added for WebAuthn version 1.1");
+                break;
```
fadidurah, line 1423: Only difference is the log statement?

fadidurah (approved):

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
p3dr0rv: that actually a bug, fix and add test! appreciated

testapps/testapp/src/main/res/raw/msal_config_webview_ppe_msa.json
p3dr0rv: I have added a new config to test this as well

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
fadidurah: nice, thanks pedro :)

fadidurah (approved):

PR #2413, opened 2025-12-06
Move our Libraries to only use artifact feed for packages, no Maven Central, Fixes AB#3395409

fadidurah:
Received an s360 item asking our libraries to always pull packages from feeds, and not declare mavenCentral as a repository to source dependencies. The feed itself has maven as an upstream. We cannot use AndroidADAL feed as that one has some versions deleted (once a version is deleted from a feed, it cannot be restored, even if an upstream has it). Created a new feed to support this https://identitydivision.visualstudio.com/Engineering/_artifacts/feed/NewAndroid. We pull packages from this new feed, but we still publish our versions to AndroidADAL. The new feed has AndroidADAL as an upstream still, so our existing collection of artifacts is still accessible.

I also took this chance to consolidate the Maven VSTS Username and Access Token fields, which are fields used to authenticate and pull artifacts from the feed. Previously, each library had their own names for these fields with an identifier based on the library in question. These PRs make it so all libraries use the same names, `ENV_VSTS_MVN_CRED_USERNAME` and `ENV_VSTS_MVN_CRED_ACCESSTOKEN`.

Validation: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1568877&view=results

[AB#3395409](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3395409)

### DIFF
--- a/azure-pipelines/pull-request-validation/pr-msal.yml
+++ b/azure-pipelines/pull-request-validation/pr-msal.yml
@@ -41,8 +41,8 @@ jobs:
     submodules: recursive
     persistCredentials: True
   - bash: |
-      echo "##vso[task.setvariable variable=ENV_VSTS_MVN_ANDROID_MSAL_USERNAME]VSTS"
-      echo "##vso[task.setvariable variable=ENV_VSTS_MVN_ANDROID_MSAL_ACCESSTOKEN]$(System.AccessToken)"
+      echo "##vso[task.setvariable variable=ENV_VSTS_MVN_CRED_USERNAME]VSTS"
+      echo "##vso[task.setvariable variable=ENV_VSTS_MVN_CRED_ACCESSTOKEN]$(System.AccessToken)"
     displayName: 'Set VSTS Fields in Environment'
   - template: azure-pipelines/templates/steps/automation-cert.yml@common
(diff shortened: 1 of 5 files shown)


azure-pipelines/pull-request-validation/pr-msal.yml
```diff
@@ -40,6 +40,10 @@ jobs:
     clean: true
     submodules: recursive
     persistCredentials: True
+  - bash: |
```
p3dr0rv: this is duplicated, right? we set credentials in automation-cert.yml

shahzaibj (approved):
Double check Pedro's comment; else LGTM

p3dr0rv (approved):
Approve with comment

azure-pipelines/pull-request-validation/pr-msal.yml
fadidurah: For whatever reason the state is not carried over, i think it's something that happens in pipeline when templates are used...

I tried it here, failed https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1570162&view=results

shahzaibj (approved):

PR #2451, opened 2026-02-06
Add precondition checks to skip tests if debug brokers are not trusted, Fixes AB#3505565

p3dr0rv:
This pull request adds conditional logic to several Android test cases to ensure they only run when the device is configured to trust or not trust debug brokers, according to their requirements. This improves test reliability and prevents false failures by skipping tests when preconditions are not met.

[AB#3505565](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3505565)


wzhipan (approved):
LGTM

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/mam/TestCase2516967.kt
```diff
@@ -49,6 +51,12 @@ class TestCase2516967 : AbstractMsalBrokerTest(){
 
     @Test
     fun test_2516967_MAM_CanUseOutlookAfterRegistration() {
+        // Skip test if preconditions are not met
+        Assume.assumeTrue(
+            "Only run this test if the device is set to trust debug brokers",
+            getShouldTrustDebugBrokers()
+        )
```
shahzaibj: Similar logic in multiple places.

Alternatively, have you considered creating a rule in combination with an annotation to handle this?

p3dr0rv: BrokerData isn’t available in uiautomationutilities. To use it, I would need to add common as a dependency. Since we already consume common4j, do you foresee any issues with adding common as well? @shahzaibj

p3dr0rv: This pattern is already used in other tests,
image
So, for now I’m inclined to leave it as is. That said, I think it would be a good idea to refactor this in the future by introducing a configuration file that defines the flights and flags used by the broker apps under test, along with rules that determine whether a test should run based on that configuration. However, this would require a medium-sized effort across the test infrastructure and pipelines.
 https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3516216

testapps/testapp/src/main/res/raw/msal_config_instance_aware_organization.json
```diff
@@ -6,7 +6,7 @@
   "authorities" : [
     {
       "type": "AAD",
-      "authority_url": "https://login.microsoftonline.com/organization"
+      "authority_url": "https://login.microsoftonline.com/organizations"
```
fadidurah, line 9: Is this change needed?

fadidurah (approved):

testapps/testapp/src/main/res/raw/msal_config_instance_aware_organization.json
p3dr0rv: This doesn’t strictly belong in this PR, but since it’s a small fix, I included it here. Creating a separate PR for such a minor change seemed like too much overhead.

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/brokerapi/TestCase1600567.java
```diff
@@ -46,6 +48,10 @@
 public class TestCase1600567 extends AbstractMsalBrokerTest {
     @Test
     public void test_1600567_nonAllowedBrokerApp() throws Throwable {
+        // Skip test if preconditions are not met
+        Assume.assumeTrue( "Only run this test if there are no local flights",
+                BuildConfig.COPY_OF_LOCAL_FLIGHTS_FOR_TEST_PURPOSES.isEmpty()
```
wzhipan, line 53: Can you explain why we allow this test only if COPY_OF_LOCAL_FLIGHTS_FOR_TEST_PURPOSES is empty? Do we need to check for the existence of a particular local flight to decide whether to skip the test?

p3dr0rv: If there are no local flights configured, we assume the test is not using flags like "-TrustDebugBrokers". 
https://identitydivision.visualstudio.com/Engineering/_git/AuthClientAndroidPipelines?path=/scripts/setup-alternating-schedule.ps1
This means the broker does not trust debug broker apps, which is exactly what we want to validate in this scenario. (test_1600567_nonAllowedBrokerApp)

This behavior is not very intuitive and relies on the assumption that using the ECS flight provider implies no flags are enabled. While this is not ideal, it is the best option we have at the moment. I’ve proposed a more robust and explicit approach, but it would require a larger refactor.
https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2451#discussion_r2795215064

wzhipan: In many scenarios the test won't run even if it doesn't trust debug broker. Can we explicitly check the local flights, so we only skip the test if we need to?

testapps/testapp/src/main/res/raw/msal_config_instance_aware_organization.json
fadidurah: sounds good, i was just checking it was intentional

fadidurah (approved):

Issue #237, opened 2017-12-01
No Activity found to handle AuthenticationActivity in msal library


NavalKishor:
Fatal Exception: java.lang.RuntimeException: Unable to resume activity {com.naval.libs/com.microsoft.identity.client.AuthenticationActivity}: android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW cat=[android.intent.category.BROWSABLE] dat=https://www.xxxxlogin.com/... pkg=com.android.chrome launchParam=MultiScreenLaunchParams { mDisplayId=0 mBaseDisplayId=0 mFlags=0 } }
at android.app.ActivityThread.performResumeActivity(ActivityThread.java:3873)
at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3914)
at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3070)
at android.app.ActivityThread.-wrap14(ActivityThread.java)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1659)
at android.os.Handler.dispatchMessage(Handler.java:102)
at android.os.Looper.loop(Looper.java:154)
at android.app.ActivityThread.main(ActivityThread.java:6823)
at java.lang.reflect.Method.invoke(Method.java)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1563)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1451)

Caused by android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW cat=[android.intent.category.BROWSABLE] dat=https://www.xxxxxlogin.com/... pkg=com.android.chrome launchParam=MultiScreenLaunchParams { mDisplayId=0 mBaseDisplayId=0 mFlags=0 } }
at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:1844)
at android.app.Instrumentation.execStartActivity(Instrumentation.java:1531)
at android.app.Activity.startActivityForResult(Activity.java:4411)
at android.app.Activity.startActivityForResult(Activity.java:4369)
at android.app.Activity.startActivity(Activity.java:4708)
at android.app.Activity.startActivity(Activity.java:4676)
at com.microsoft.identity.client.AuthenticationActivity.onResume(SourceFile:237)
at android.app.Instrumentation.callActivityOnResume(Instrumentation.java:1277)
[...]

PR #2546, opened 2026-07-25
Gate msal PR check on code-coverage drop

fadidurah:
## What
Make the msal PR check **fail on a code-coverage drop** and show where it regressed.

`pr-msal.yml` already builds JaCoCo for the PR branch (`jacocoReport`) and the dev branch (`jacocoReportDev`) and calls the shared `compare-code-coverage.yml@common` template. This passes explicit `metric: LINE` / `tolerance: 0` / `failOnDrop: true` so the comparison now gates the PR on any LINE coverage regression and posts a per-class breakdown of exactly which classes lost coverage.

## Depends on
The per-class diff + gating logic lives in `common` (AzureAD/microsoft-authentication-library-common-for-android PR #3194) and is consumed via `@common` (ref: dev) once merged. This mirrors the broker gate (ad-accounts-for-android #237).

### DIFF
--- a/azure-pipelines/pull-request-validation/pr-msal.yml
+++ b/azure-pipelines/pull-request-validation/pr-msal.yml
@@ -167,6 +167,10 @@ stages:
   - template: azure-pipelines/code-coverage/compare-code-coverage.yml@common
     parameters:
       commonRepoAlias: common
+      # Fail the PR check on any LINE coverage drop and surface a per-class breakdown of where it regressed.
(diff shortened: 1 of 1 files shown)


fadidurah:
Not needed - the shared compare-code-coverage.yml@common template now defaults to metric=LINE/tolerance=0/failOnDrop=true, so the msal PR check gates on coverage drop automatically via @common once common PR #3194 merges. No per-repo change required.

PR #2547, opened 2026-07-30
Native Auth V2 SDK interface for Android Fixes AB#3693579

spetrescu84:
This is the PR for the Native Auth V2 SDK interface for Android.

Adds the public SDK interface surface only:
- `NativeAuthResultV2` — unified result type
- `NativeAuthFlowStateV2` — unified state with continuation steps
- `NativeAuthErrorV2` and `NativeAuthFlowScenarioV2`
- `signInV2` / `signUpV2` / `resetPasswordV2` entry points (suspend + callback) on the public interface and application
- Kotlin and Java tests

All step and entry-point methods currently return a `not_implemented` result.

Fixes [AB#3693579](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3693579)


msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NativeAuthFlowStateV2.kt
```diff
@@ -0,0 +1,281 @@
...
+
+/**
+ * NativeAuthFlowStateV2 is the single unified state for the Native Auth V2 surface.
+ *
+ * @property continuationToken Continuation token passed in the next request.
+ * @property correlationId Correlation ID taken from the previous API response and passed to the next request.
+ * @property scenario Identifies which part of the Native Auth V2 surface this state belongs to.
+ * @property config Configuration used by Native Auth.
+ */
+class NativeAuthFlowStateV2 internal constructor(
```
djanardhan98: `NativeAuthFlowStateV2` currently exposes all continuation methods (`submitCode`, `submitPassword`, `submitAttributes`, etc.) from one type. 

Can you help me understand how we plan to enforce valid next-step transitions in V2, since iOS constrains this by returning a concrete per-step state type (e.g., `CodeRequiredState`, `PasswordRequiredState`) that only exposes the next method(s) allowed for that step?

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -568,6 +573,92 @@ class NativeAuthPublicClientApplication(
...
+    override suspend fun signInV2(parameters: NativeAuthSignInParameters): NativeAuthResultV2 {
+        LogSession.logMethodCall(
+            tag = TAG,
+            correlationId = null,
+            methodName = "${TAG}.signInV2(parameters: NativeAuthSignInParameters)"
+        )
+        return notImplementedV2(NativeAuthFlowScenarioV2.SIGN_IN)
+    }
+
+    override fun signInV2(parameters: NativeAuthSignInParameters, callback: NativeAuthV2Callback) {
```
djanardhan98, line 596: I see we're reusing V1 parameters for all of the entry points here, do we intend to create V2 versions for all the parameters later and wire them here in subsequent PRs?

Are we not doing it now because we're unsure of the parameters for v2?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/NativeAuthFlowScenarioV2.kt
```diff
@@ -0,0 +1,34 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+package com.microsoft.identity.nativeauth.statemachine.errors
+
+/**
+ * Native Auth V2 flow scenario identifier (sign-in, sign-up, reset password).
+ */
+enum class NativeAuthFlowScenarioV2 {
```
djanardhan98, line 29: FlowScenario file should be outside the errors path?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/NativeAuthErrorV2.kt
```diff
@@ -0,0 +1,74 @@
...
+        error = error,
+        errorMessage = errorMessage,
+        correlationId = correlationId,
+        errorCodes = errorCodes,
+        exception = exception
+    ) {
+
+    fun isNotImplemented(): Boolean = this.errorType == NativeAuthV2ErrorTypes.NOT_IMPLEMENTED
+
+    fun isInvalidCode(): Boolean = this.errorType == ErrorTypes.INVALID_CODE
```
djanardhan98: Will eventual deprecation of v1 become messy if we try to mix v1 and v2 here? (Like some errors are using NativeAuthV2ErrorTypes and others are relying on v1 ErrorTypes)

Should we define all the types in NativeAuthV2ErrorTypes to keep everything contained in V2? 

Open to hearing whether we already thought about this and we intentionally want to reuse parts of V1 as is for some reason, or if there's a better way to do it

djanardhan98: I'm generally trying to understand whether we intend to completely decouple v2 from v1

forge.json
```diff
@@ -0,0 +1,15 @@
...
+  "$schema": ".forge/forge.schema.json",
+  "forgeVersion": 2,
+  "repo": "AzureAD/microsoft-authentication-library-for-android",
+  "host": "github",
+  "baseBranch": "dev",
+  "team": "MSAL Android + DevEx CIAM",
+  "mcps": {
+    "source-code": "github-mcp-server",
+    "work-items": "ado-identitydivision"
+  },
```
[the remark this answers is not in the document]
spetrescu84: Removed

.vscode/mcp.json
```diff
@@ -0,0 +1,11 @@
+{
+  "servers": {
+    "ado-identitydivision": {
+      "type": "http",
+      "url": "https://mcp.dev.azure.com/identitydivision",
+      "headers": {
+        "X-MCP-Toolsets": "repos,wit,pipelines,wiki,work"
+      }
+    }
+  }
```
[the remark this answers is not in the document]
spetrescu84: Removed

antrix1989:
@spetrescu84 how do we provide scopes for sign in after SSPR? Will it be same as in v1?

spetrescu84:
> Silviu Petrescu (@spetrescu84) how do we provide scopes for sign in after SSPR? Will it be same as in v1?

Will implement this with the SSPR flow to make clear where it's coming from. Otherwise we would need to add SingInAfterSignUp also

antrix1989 (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NativeAuthFlowStateV2.kt
spetrescu84: Good question, this has actually changed, it's now per state.

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
spetrescu84: It's because they can be reused. Whenever stuff from v1 can be reused we reuse it. We do plan to rename the interfaces before GA if possible or after v1 deprecation

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/NativeAuthErrorV2.kt
spetrescu84: Good point, I've removed `NativeAuthV2ErrorTypes` and moved `NOT_IMPLEMENTED` to `ErrorTypes` as going forward, even after deprecation we will use `ErrorTypes`. We can add all errors there for v2.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/NativeAuthFlowScenarioV2.kt
spetrescu84: Done

djanardhan98 (dismissed):

fadidurah (dismissed):

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -568,6 +572,93 @@ class NativeAuthPublicClientApplication(
...
+        }
+    }
+
+    override suspend fun signUpV2(parameters: NativeAuthSignUpParameters): NativeAuthResultV2 {
+        LogSession.logMethodCall(
+            tag = TAG,
+            correlationId = null,
+            methodName = "${TAG}.signUpV2(parameters: NativeAuthSignUpParameters)"
+        )
+        return notImplementedV2(NativeAuthFlowScenarioV2.SIGN_UP)
```
fadidurah, line 618: Are we sure we even want to ship these? Why not just omit these methods from PCA until they are implemented?

spetrescu84: Yeah, actually I switched branch to `dev` but added the warning ` * Warning: This API is experimental. It may be changed in the future without notice. Do not use in production applications.` across all APIs

djanardhan98 (approved):

fadidurah (approved):

PR #2553, opened 2026-08-12
Re-enable Native Auth SSPR tests with Mail.tm, Fixes AB#3729032

djanardhan98:
## Summary

- Start the Native Auth v1 E2E rollout with SSPR only, using authenticated Mail.tm inboxes.
- Restrict the dedicated PR-validation invocation to `com.microsoft.identity.client.e2e.tests.network.nativeauth.SSPRTest`; retain the existing opt-out label, timeout, and secret handling.
- Enable the six implemented SSPR cases, including resend. Keep only the existing social-account placeholder ignored.
- Keep mailbox checkpoints before requesting/resending OTPs, and submit the resent OTP through the state returned by `resendCode()`.
- Defer the sign-in, sign-up, MFA, and JIT scenario migrations to follow-up PRs. Their test files are unchanged from this PR's base.
- Keep the required assertion/throttle diagnostics and bounded retry behavior. Run their regression tests in the offline unit-test package rather than the network E2E package.
- Keep the retry policy and its regression tests independent of the Robolectric E2E fixture, so they do not initialize shared account-cache handles or leak state into ordinary auth tests.
- Authenticate the skip-label lookup using the existing checkout credential, scoped to this repository. For PR builds, skip only when the exact label is confirmed; failed lookups warn and run SSPR instead of silently bypassing the gate.

## Rollout approach

Observe repeated, non-skipped SSPR CI runs before enabling another scenario. Review failures, retries, and duration rather than treating an opted-out run as evidence of stability.

Use separate follow-up PRs for the remaining scenario groups and their fixes, following the staged approach in AzureAD/microsoft-authentication-library-for-objc#3040, AzureAD/microsoft-authentication-library-for-objc#3045, and AzureAD/microsoft-authentication-library-for-objc#3066.

Resend remains enabled. This change does not copy the iOS blanket resend skip or convert exhausted throttle retries into successful/skipped coverage.

## Validation

- Ran 21 PowerShell/Pester label-gate regression cases: all passed. Coverage includes repository-scoped authentication, explicit skip/no-skip decisions, missing credentials, HTTP 401/403/429/500, timeouts, malformed responses, retry recovery, and sanitized diagnostics. Reintroducing automatic skip on lookup failure makes the 12 affected regression cases fail.
- Ran the full ordinary CI unit-test and coverage task, `:msal:jacocoTestReport`, locally with code coverage, Robolectric SDK 33, and the mock API configured: 736 tests, 725 passed, 11 skipped, zero failures/errors. JaCoCo report generated.
- Reproduced the earlier unit-test failure with the retry-helper tests plus existing V1 sign-in and V2 Java tests. The inherited E2E cleanup initialized a stale preferences singleton; later account cleanup used the wrong preferences handle. Extracting the retry policy into a pure JVM utility removes that fixture dependency without changing production code or bypassing tests.
- Re-ran that combined 24-test reproducer after the fix: all passed, zero failures, none skipped. The full suite also includes the 15 classifier/retry-helper regression tests.
- Verified actual Gradle test selection with `--test-dry-run`: only `SSPRTest` is selected, including resend. This was selection validation, not live E2E execution.
- Live SSPR/resend success has not been verified locally because `EMAIL_PROVIDER_PASSWORD` is not configured. The secured CI run must establish the live result.

## Pipeline configuration

- The dedicated task and test-run title identify SSPR explicitly.
- The secured `EMAIL_PROVIDER_PASSWORD` variable remains an environment variable on the E2E task.
- Normal unit-test validation remains separate from the SSPR-only live invocation.
- The label lookup reuses `checkout: self` with `persistCredentials: true`; it does not introduce a new PAT or use `System.AccessToken` as a GitHub token. Missing credentials produce a warning without making an unauthenticated request.
- Label lookup uses three attempts with 2/4-second backoff, connection and operation timeouts, JSON validation, and no redirects. Failed lookups leave SSPR enabled. The existing non-PR-run exemption is preserved.
- The pipeline runs the label-gate regression suite before evaluating the skip label.

## Dependency

The Mail.tm integration in AzureAD/microsoft-authentication-library-common-for-android#3219 is already merged. This PR does not modify the Common submodule gitlink.

- Fixes [AB#3729032](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3729032)


azure-pipelines/pull-request-validation/pr-msal.yml
```diff
@@ -75,6 +75,17 @@ stages:
           tasks: msal:jacocoTestReport  -PcodeCoverageEnabled=true -ProbolectricSdkVersion=${{variables.robolectricSdkVersion}} -PmockApiUrl=$(MOCK_API_URL) -PnativeAuthConfigString=$(NATIVE_AUTH_CONFIG_STRING)
           javaHomeSelection: $(BuildParameters.javaHomeSelection)
           jdkVersion: 1.17
+      - task: Gradle@2
+        displayName: Run Native Auth E2E Tests
+        inputs:
+          tasks: msal:testLocalDebugUnitTest -Plabtest -ProbolectricSdkVersion=${{variables.robolectricSdkVersion}} -PnativeAuthConfigString=$(NATIVE_AUTH_CONFIG_STRING) --tests com.microsoft.identity.client.e2e.tests.network.nativeauth.*
```
shahzaibj: Are we gating every PR on these now? They hit mail.tm + a live tenant, and these are the same tests that were @Ignore'd for OTP flakiness. I feel like unrelated PRs will start failing validation whenever mail.tm has a hiccup or the tenant throttles OTPs. Could these run on a nightly/scheduled pipeline instead of the per-PR gate?

azure-pipelines/pull-request-validation/pr-msal.yml
```diff
@@ -75,6 +75,17 @@ stages:
...
+          javaHomeSelection: $(BuildParameters.javaHomeSelection)
+          jdkVersion: 1.17
+          publishJUnitResults: true
+          testResultsFiles: '**/build/test-results/testLocalDebugUnitTest/TEST-*.xml'
+          testRunTitle: Native Auth E2E Tests (PR Branch)
+        env:
+          EMAIL_PROVIDER_PASSWORD: $(EMAIL_PROVIDER_PASSWORD)
```
shahzaibj: This drops EMAIL_PROVIDER_PASSWORD into the env of a task that runs the PR branch's own code (checkout: self), and these tests already make outbound http calls. So anyone who can push to a PR branch could read it with a getenv and ship it out to themselves. It's a throwaway mail.tm password so low stakes, but on a public repo are we sure fork/external PRs don't get secrets here?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
```diff
@@ -154,25 +156,31 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
...
         }
     }
 
+    private fun retryOrFail(error: Throwable, retryCount: Int, maxRetries: Int) {
+        if (retryCount >= maxRetries) {
+            throw AssertionError(error.message).apply { initCause(error) }
+        }
+
+        // Avoid repeatedly requesting OTPs while the Native Auth test tenant is throttling them.
+        Thread.sleep(5_000L * (1L shl retryCount))
```
shahzaibj: maxRetries=5 with this backoff means a persistently throttled test sleeps 5+10+20+40+80 = ~155s before it finally fails, and every retry re-runs the whole flow (new inbox + signup + OTP request). Across the suite on a throttling day that adds up fast. Do we really need 5 retries here?

spetrescu84 (approved):

azure-pipelines/pull-request-validation/pr-msal.yml
shahzaibj: Are we gating every PR on these now? They hit mail.tm + a live tenant, and these are the same tests that were @Ignore'd for OTP flakiness. I feel like unrelated PRs will start failing validation whenever mail.tm has a hiccup or the tenant throttles OTPs. Could these run on a nightly/scheduled pipeline instead of the per-PR gate?

azure-pipelines/pull-request-validation/pr-msal.yml
shahzaibj: This drops EMAIL_PROVIDER_PASSWORD into the env of a task that runs the PR branch's own code (checkout: self), and these tests already make outbound http calls. So anyone who can push to a PR branch could read it with a getenv and ship it out to themselves. It's a throwaway mail.tm password so low stakes, but on a public repo are we sure fork/external PRs don't get secrets here?

azure-pipelines/pull-request-validation/pr-msal.yml
shahzaibj, line 123: Are we gating every PR on these now? They hit mail.tm + a live tenant, and these are the same tests that were @Ignore'd for OTP flakiness. I feel like unrelated PRs will start failing validation whenever mail.tm has a hiccup or the tenant throttles OTPs. Could these run on a nightly/scheduled pipeline instead of the per-PR gate?

azure-pipelines/pull-request-validation/pr-msal.yml
shahzaibj, line 130: This drops EMAIL_PROVIDER_PASSWORD into the env of a task that runs the PR branch's own code (checkout: self), and these tests already make outbound http calls. So anyone who can push to a PR branch could read it with a getenv and ship it out to themselves. It's a throwaway mail.tm password so low stakes, but on a public repo are we sure fork/external PRs don't get secrets here?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
```diff
@@ -154,25 +156,31 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
...
-                //1secmail occasionally has a delay for emails to arrive / return from the API, or throws an internal server error, which causes tests to fail
-                //In this case, retry the test
-                if (retryCount >= maxRetries) {
-                    Assert.fail(e.message)
-                    shouldRetry = false
-                } else {
-                    retryCount++
+                return
+            } catch (e: AssertionError) {
+                if (!e.message.orEmpty().contains("AADSTS701014")) {
```
shahzaibj: What's AADSTS701014 here? We only retry AssertionErrors whose message contains it and rethrow everything else. Is it a transient eSTS thing that's actually worth riding out with the backoff? Matching the code as a substring of the assert message also feels a bit brittle since that text isn't really a stable contract.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
shahzaibj: maxRetries=5 with this backoff means a persistently throttled test sleeps 5+10+20+40+80 = ~155s before it finally fails, and every retry re-runs the whole flow (new inbox + signup + OTP request). Across the suite on a throttling day that adds up fast. Do we really need 5 retries here?

azure-pipelines/pull-request-validation/pr-msal.yml
```diff
@@ -75,6 +75,59 @@ stages:
...
+            echo "No PR number available (not a PR build). Skipping Native Auth E2E tests."
+            echo "##vso[task.setvariable variable=SKIP_NATIVE_AUTH_E2E]true"
+            exit 0
+          fi
+
+          LABELS_URL="https://api.github.com/repos/${REPO}/issues/${PR_NUMBER}/labels"
+          echo "Fetching PR labels from ${LABELS_URL}"
+          # Capture the body and the HTTP status separately, so that a failed API call is
+          # distinguishable from a successful call that simply returned no matching label.
+          HTTP_STATUS=$(curl -sS -o "$LABELS_FILE" -w "%{http_code}" "$LABELS_URL" || echo "000")
```
shahzaibj, line 96: This label check hits api.github.com unauthenticated. Unauth github API is 60/hr per IP, and on the shared windows-2022 pool that egress IP is shared across a ton of jobs, so on a busy day this'll come back 403 rate-limited. The fail-safe right below then treats any non-200 as skip=true, so the native auth e2e tests we're re-enabling would just silently stop running with a green pipeline and nobody notices. kinda defeats the point of making them a PR gate.

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
```diff
@@ -150,29 +170,40 @@ abstract class NativeAuthPublicClientApplicationAbstractTest : IPublicClientAppl
...
+                return
+            } catch (e: AssertionError) {
+                // Classification (typed errorCodes first, assertion message as unconditional
+                // fallback) lives in the classifier so it is unit-testable without standing up
+                // this Robolectric-backed abstract test.
+                if (!NativeAuthEmailOTPErrorClassifier.isThrottleError(e)) {
+                    throw e
                 }
+                retryOrFail(e, retryCount++, maxRetries)
+            } catch (e: Exception) {
```
shahzaibj: The AssertionError path is throttle-gated now (retry only if it's 701014, otherwise rethrow), but this Exception catch still retries anything 3x. So a real non-throttle exception rides out the full backoff before failing and comes out as a flattened AssertionError. Is that asymmetry intended, or should this path classify too?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
djanardhan98: Fixed in `a1be871d`. AADSTS701014 is transient email OTP throttling. Classification now uses the structured `errorCodes` value first, with a tested message fallback only when structured data is unavailable.

azure-pipelines/pull-request-validation/pr-msal.yml
djanardhan98: Fair point — a mail.tm hiccup shouldn't fail a PR that never touched Native Auth.

I went with an escape hatch rather than nightly: a `skip-native-auth-e2e-tests` label drops the E2E task (same thing iOS does in `pr-validation.yml`), plus a 30 min timeout. The label check fail-safes to skipping, so if the labels API is down it can't block anything either.

Keeping them on the gate is what makes a regression attributable to the PR that caused it. The label gives us the unblock without giving up that attribution.

Let me know if you think that's a reasonable solution

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
djanardhan98: Fixed in `a1be871d`. Reduced the retry budget from five to three and capped exponential backoff at 20 seconds, reducing worst-case sleep from approximately 155 seconds to 35 seconds.

azure-pipelines/pull-request-validation/pr-msal.yml
djanardhan98: I checked the pipeline definition.

On **MSAL - Build & Test** (the pipeline that runs this YAML), the PR trigger has:

```
forks.enabled      : False
forks.allowSecrets : False
```

So fork PRs don't build here at all, and secrets wouldn't be exposed even if they did. Same on the other two repo pipelines with PR triggers. 

That leaves the case you described — someone with push access to a PR branch — which is true, but that's collaborators only, and it's a throwaway mail.tm inbox password. It's also the same pattern `NATIVE_AUTH_CONFIG_STRING` already uses in this job, so it's not new exposure.

Worth flagging for later: the day anyone turns on fork builds, `allowSecrets` has to stay off or this needs rethinking. 

Open to discussing if it's still worth making the KeyVault change and pull from there instead for this particular secret though, let me know!

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTest.kt
djanardhan98: Fixed in `03d18c5fa`. I removed the broad `catch (Exception)` path. `retryOperation` now only intercepts `AssertionError`, and retries only when it is classified as AADSTS701014. All other exceptions propagate immediately with their original type, stack trace, and cause—without backoff or conversion to `AssertionError`. I also added a regression test confirming a non-throttle exception is attempted once and rethrown unchanged.

djanardhan98:
@copilot resolve the merge conflicts in this pull request, point the submodule to latest common dev branch

msal/src/test/java/com/microsoft/identity/client/e2e/utils/NativeAuthEmailOTPErrorClassifier.kt
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @param errorCodes the error codes returned by the authentication server, if available.
+     * @param errorDescription the error message/description text, if available.
+     */
+    @JvmStatic
+    fun isThrottleError(errorCodes: List<Int>?, errorDescription: String?): Boolean {
+        if (errorCodes?.contains(THROTTLE_ERROR_CODE) == true) {
+            return true
+        }
+        return errorDescription?.contains(THROTTLE_ERROR_DESCRIPTION_MARKER) ?: false
```
shahzaibj, line 69: I don't think the typed errorCodes path here fires for SSPR or sign-up at all. ResetPasswordChallengeApiResponse and SignUpChallengeApiResponse in common don't parse `error_codes`, and NativeAuthMsalController builds the APIError for those two challenge failures without errorCodes (reset password around line 1310, signup around 2025), so Error.errorCodes comes back null on both. Only signin/JIT challenge responses carry error_codes. 701014 also isn't in any of common's known-code lists.

So for the flows this PR re-enables (SSPR is right in the title) isThrottleError only ever reaches the errorDescription.contains("AADSTS701014") fallback. That's the same substring match we had before the classifier. The typed-first shape kinda implies the structured code is the primary signal, but for these two flows it's always null.

Are we ok leaning on the AADSTS701014 substring for SSPR/signup, given the typed path can't fire there?

msal/src/test/java/com/microsoft/identity/client/e2e/tests/network/nativeauth/NativeAuthPublicClientApplicationAbstractTestTest.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    @Test
+    fun retryOperationDoesNotRetryNonThrottleExceptions() {
+        val expected = IllegalStateException("Non-throttle failure")
+        var attempts = 0
+
+        val actual = assertThrows(IllegalStateException::class.java) {
+            retryOperation(maxRetries = 1) {
+                attempts++
+                throw expected
```
shahzaibj, line 41: IllegalStateException isn't an AssertionError, so it never reaches the catch (e: AssertionError) in retryOperation. it just falls straight through here, so this is only really checking that a stray RuntimeException isn't swallowed, not the throttle gating.

the branch I'd want pinned is a non-701014 AssertionError getting rethrown with no retry (the isThrottleError false -> throw e path). can we add that case? attempts should still be 1 and the original error should come back out.

azure-pipelines/pull-request-validation/pr-msal.yml
djanardhan98: As discussed offline, we’ll keep SSPR as a required per-PR gate for a two-week observation period, with  skip-native-auth-e2e-tests  available for urgent unblock.
I’ll own monitoring and review failures, retries, skips, and author impact weekly. I’ll also let the team know to flag blocking/flaky runs to me so authors aren’t routinely expected to investigate provider issues themselves. We’ll reassess after two weeks of rollout, or sooner if it causes recurring disruption.

msal/src/test/java/com/microsoft/identity/client/e2e/utils/NativeAuthTestRetryTest.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    @Test
+    fun retryOperationDoesNotRetryNonThrottleExceptions() {
+        val expected = IllegalStateException("Non-throttle failure")
+        var attempts = 0
+
+        val actual = assertThrows(IllegalStateException::class.java) {
+            retryOperation(maxRetries = 1) {
+                attempts++
+                throw expected
```
djanardhan98: Added  retryOperationDoesNotRetryNonThrottleAssertions  in  NativeAuthTestRetryTest . It throws a non-throttle  AssertionError , verifies that the exact original instance is rethrown, and asserts one attempt.
The retry tests now exercise the shared policy as plain JVM tests, without inheriting the E2E fixture.

msal/src/test/java/com/microsoft/identity/client/e2e/utils/NativeAuthEmailOTPErrorClassifier.kt
djanardhan98: You’re correct—for SSPR challenge failures, Common currently doesn’t propagate  error_codes , so the classifier relies on the  AADSTS701014  text fallback. I’m comfortable retaining that fallback for this SSPR-only rollout rather than expanding this PR.
I’ve created AB#3755379 to track structured error-code propagation and improved E2E failure diagnostics, including safe correlation IDs, failing-stage information, and retry context. We’ll take this up alongside the staged work to enable additional Native Auth test scenarios.
For now, retries remain bounded, non-throttle failures are propagated unchanged, and exhausted retries still fail the test rather than being skipped.

djanardhan98: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3755379

PR #2557, opened 2026-08-25
Add GetCookies WebApps automation coverage, Fixes AB#3715681

melissaahn:
## Summary
- Add BrokerHost UI automation coverage for the WebApps `GetCookies` API.
- Parse the response and require at least one cookie with nonblank data.
- Verify cookie names use the same base-and-suffix enumeration as `GetAllSsoTokens`.
- Update the Common submodule to the GetCookies contract commit.

## Related Common PRs
- GetCookies reenablement: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/3230
- Previous GetCookies disablement: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/3135

## Testing
- `:msalautomationapp:assembleLocalBrokerHostDebugAndroidTest`

## Tracking
- [AB#3715681](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3715681)

### DIFF
--- a/changelog
+++ b/changelog
@@ -2,7 +2,6 @@ MSAL Wiki : https://github.com/AzureAD/microsoft-authentication-library-for-andr
 
 vNext
 ----------
-- [PATCH] Reenable GetCookies WebApps API automation coverage (#2557)
 
 Version 8.4.2
 ----------
--- a/changelog
+++ b/changelog
@@ -2,6 +2,7 @@ MSAL Wiki : https://github.com/AzureAD/microsoft-authentication-library-for-andr
 
 vNext
 ----------
+- [PATCH] Reenable GetCookies WebApps API automation coverage (#2557)
 
 Version 8.4.2
 ----------
--- a/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/brokerapi/TestCase3571739.java
+++ b/msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/brokerapi/TestCase3571739.java
@@ -60,7 +60,7 @@ import org.junit.Test;
 //           must resolve the account from the PRT/broker-account store keyed by homeAccountId
 //           rather than from the (empty) per-clientId MSAL token cache. It must succeed without
 //           falling back to an interactive account picker.
-// GetAllSsoTokens and SignOut are also exercised.
+// GetCookies, GetAllSsoTokens, and SignOut are also exercised.
 // https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3571739
 @SupportedBrokers(brokers = BrokerHost.class)
 @LocalBrokerHostDebugUiTest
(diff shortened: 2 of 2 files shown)


changelog
```diff
@@ -2,6 +2,7 @@ MSAL Wiki : https://github.com/AzureAD/microsoft-authentication-library-for-andr
 
 vNext
 ----------
+- [PATCH] Reenable GetCookies WebApps API automation coverage (#2557)
```
fadidurah: SUPER DUPER NIT, we probably don't want this in changelog

fadidurah (approved):

changelog
melissaahn: Good point... changelog now has no updates and I added the tag.

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/brokerapi/TestCase3571739.java
```diff
@@ -209,7 +215,64 @@ public void test_3571739_webAppsOperations() throws Throwable {
...
+        );
+        for (int i = 0; i < cookiesResponse.response.size(); i++) {
+            final CookieItem cookie = cookiesResponse.response.get(i);
+            Assert.assertNotNull("GetCookies should not return a null cookie", cookie);
+            final String expectedName = i == 0 ? PRT_COOKIE_NAME : PRT_COOKIE_NAME + i;
+            Assert.assertEquals(
+                    "GetCookies should enumerate cookie names",
+                    expectedName,
+                    cookie.name
+            );
```
shahzaibj: Is the cookie order here actually guaranteed? This asserts index 0 is x-ms-RefreshTokenCredential, index 1 is ...1, index 2 is ...2 and so on. If GetCookies doesn't hand them back in that exact suffix order (or ever returns some non-PRT cookie for login.microsoftonline.com) the assertEquals will flake. do we know the list order is deterministic here, or would matching on the set of names be safer than going by position?

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/brokerapi/TestCase3571739.java
```diff
@@ -209,7 +215,64 @@ public void test_3571739_webAppsOperations() throws Throwable {
...
+                cookiesResponse.response.isEmpty()
+        );
+        for (int i = 0; i < cookiesResponse.response.size(); i++) {
+            final CookieItem cookie = cookiesResponse.response.get(i);
+            Assert.assertNotNull("GetCookies should not return a null cookie", cookie);
+            final String expectedName = i == 0 ? PRT_COOKIE_NAME : PRT_COOKIE_NAME + i;
+            Assert.assertEquals(
+                    "GetCookies should enumerate cookie names",
+                    expectedName,
+                    cookie.name
```
shahzaibj: I don't think the suffix naming holds on this path. GetCookies builds each item's name straight from the PRT result's cookieName, which is always `x-ms-RefreshTokenCredential` with no index. the `1`, `2`... suffix only gets tacked on in the browser-sso header path (SsoHeaderResult), not in the WebApps GetCookies suboperation. so the moment there's more than one PRT account on the device, cookie[1].name is still `x-ms-RefreshTokenCredential`, not `...1`, and this assertEquals fails. it only passes today because the lab account is single-PRT so we never get past i==0. can we just assert each name equals PRT_COOKIE_NAME instead of baking the index into the expected name?

msalautomationapp/src/androidTest/java/com/microsoft/identity/client/msal/automationapp/testpass/broker/brokerapi/TestCase3571739.java
```diff
@@ -209,7 +215,64 @@ public void test_3571739_webAppsOperations() throws Throwable {
...
-        // -------- Step 3: GetAllSsoTokens --------
+        // -------- Step 3: GetCookies --------
+        // Navigate to the Broker API tab
+        brokerHost.brokerApiFragment.launch();
+
+        // Scroll down to make the cookies URL field and button visible
+        new UiScrollable(new UiSelector().scrollable(true)).scrollIntoView(new UiSelector().resourceId(BUTTON_GET_COOKIES));
+
+        UiAutomatorUtils.handleInput(EDIT_TEXT_COOKIES_URL, MICROSOFT_ONLINE);
+        UiAutomatorUtils.handleButtonClick(BUTTON_GET_COOKIES);
```
djanardhan98, line 226: Would it be useful for this E2E test to check  getSupportedWebAppContracts  before calling  GetCookies ? Since real callers use capability discovery first, asserting that the installed Broker advertises  GetCookies  would also cover the Common discovery → Broker dispatch integration. Not sure if that discovery flow is already covered elsewhere.

djanardhan98 (approved):

PR #2558, opened 2026-08-26
Implement Native Auth V2 SSPR flow, Fixes AB#3693580

djanardhan98:
## Summary

Implements the Native Auth V2 self-service password reset (SSPR) flow in MSAL Android, from reset initiation through verification-code handling, password update, and explicit post-reset sign-in.

Fixes [AB#3693580](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3693580).

## Changes

- Implements `resetPasswordV2(...)` with signed-in-account validation and typed Common command/result mapping.
- Implements all five V2 SSPR command paths and adapter factories:
  - reset-password start
  - submit verification code
  - resend verification code
  - submit new password
  - sign in after password reset
- Adds typed state transitions for code entry, new-password entry, and post-reset sign-in.
- Preserves the opaque Common continuation state across state transitions and Parcelable process recreation without exposing or fabricating its continuation token.
- Propagates Common error codes and sub-errors through typed and generic V2 errors, with concrete retry-state types for recoverable failures.
- Preserves correlation IDs in command parameters, error results, and exception logs.
- Preserves structured coroutine cancellation, interrupts pending command waits, and signals Common to cancel active HTTP work.
- Documents `NativeAuthResetPasswordParameters.scopes` as V2-only; V1 behavior is unchanged.
- Updates the Common submodule to `665f676f8`, which contains the V2 SSPR command, result, continuation-state, cancellation, and public API ID contracts used by this implementation.

## Tests

Adds deterministic coverage for:

- field propagation through all five V2 command-parameter factories, including scopes, claims, continuation state, correlation ID, and token cache
- positive, recoverable-error, terminal-error, redirect, not-implemented, exception, and cancellation mappings across all five command paths
- suspend and callback precondition/error contracts
- concrete retry-state typing
- non-null continuation-state Parcelable round trips and restored-state continuation
- cancellation while blocked on a command future, including Common HTTP cancellation signaling

Validated with targeted `:msal:testLocalDebugUnitTest` selectors for `CommandParametersTest`, `NativeAuthV2InterfaceKotlinTest`, `NativeAuthV2ErrorsTest`, and `NativeAuthV2StatesTest`.

## Dependencies and limitations

- The pinned Common SSPR V2 changes must land before this PR can merge.
- This PR implements the `RESET_PASSWORD` scenario. Sign-in/sign-up reuse of the shared V2 states requires scenario-specific `PublicApiId` constants in Common; no unsupported IDs are introduced here.
- Post-reset sign-in uses scopes and claims retained in Common's continuation state. Caller-provided post-reset scope/claim overrides require a future Common contract change.


djanardhan98:
@copilot whole of native auth v2 flows are completely new and have not been shipped before, re-check the review comments according to that and resolve if they don't apply anymore

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/NativeAuthErrorV2.kt
```diff
@@ -51,7 +53,8 @@ open class NativeAuthErrorV2(
     override val correlationId: String,
     override val scenario: NativeAuthFlowScenarioV2,
     override val errorCodes: List<Int>? = null,
-    override var exception: Exception? = null
+    override var exception: Exception? = null,
+    open val nextState: NativeAuthBaseStateV2? = null
```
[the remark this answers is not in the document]
djanardhan98: V2 is a new surface, not shipped before

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/CodeErrorsV2.kt
```diff
@@ -48,8 +50,9 @@ class SubmitCodeErrorV2(
     scenario: NativeAuthFlowScenarioV2,
     errorCodes: List<Int>? = null,
     val subError: String? = null,
-    exception: Exception? = null
-) : NativeAuthErrorV2(errorType, error, errorMessage, correlationId, scenario, errorCodes, exception) {
+    exception: Exception? = null,
+    override val nextState: CodeRequiredStateV2? = null
```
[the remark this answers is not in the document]
djanardhan98: V2 is a new surface, not shipped before

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/ResetPasswordErrorsV2.kt
```diff
@@ -75,8 +77,9 @@ class SubmitNewPasswordErrorV2(
     scenario: NativeAuthFlowScenarioV2,
     errorCodes: List<Int>? = null,
     val subError: String? = null,
-    exception: Exception? = null
-) : NativeAuthErrorV2(errorType, error, errorMessage, correlationId, scenario, errorCodes, exception) {
+    exception: Exception? = null,
+    override val nextState: NewPasswordRequiredStateV2? = null
```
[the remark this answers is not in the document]
djanardhan98: V2 is a new surface, not shipped before

antrix1989:
## CIAM Native Auth v2 SSPR parity review

Compared this PR at `3f620f051045498c8801e3a7d2e39e387ab7f9e5` with MSAL iOS/macOS `dev` at `7ff87130ca32b7641f450b441a0230dad4a3d398`. I found four API-behavior differences that should be reviewed before merging.

### Severity: High – Invalid OTP uses the wrong recoverability contract

**Issue:** `NativeAuthV2ResponseParser.mapInteractionError` maps `invalidGrant + invalidContinuationToken` and `invalidGrant + invalid_oob_value` to `InvalidCode`, but does not recognize the iOS/service v2 contract `invalidGrant + invalidOneTimeCode`.

Locations:
- `common/common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/v2/NativeAuthV2ResponseParser.kt:327-375`
- `msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/CodeRequiredStateV2.kt:150-188`
- iOS reference: `MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift:255-275`

**Impact:** A mistyped OTP can become a terminal/general error with no retry state, while an invalid continuation token can be presented as a retryable OTP error.

**Recommendation:** Map `invalidGrant + invalidOneTimeCode` to `InvalidCode`, and keep invalid-continuation-token failures terminal/general.

### Severity: Medium – Post-reset sign-in cannot accept scopes and claims at invocation

**Issue:** Android exposes only parameterless `SignInAfterResetPasswordStateV2.signIn()`. Scopes are supplied when reset starts, and reset-start parameters do not accept a claims request. iOS accepts optional scopes and claims on the explicit token-acquiring sign-in operation.

Locations:
- `msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInAfterResetPasswordStateV2.kt:88-126`
- `msal/src/main/java/com/microsoft/identity/nativeauth/parameters/NativeAuthResetPasswordParameters.kt:30-36`
- `msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java:1264-1297`
- iOS references: `MSALNativeAuthSignInAfterResetPasswordParameters.swift:25-34` and `MSALNativeAuthSignInAfterResetPasswordState.swift:27-42`

**Impact:** Android apps cannot supply a claims challenge at the terminal sign-in step and must choose token scopes before SSPR begins.

**Recommendation:** Add terminal sign-in parameters containing optional scopes and claims, while retaining a parameterless convenience overload if desired.

### Severity: Medium – Cached-account precondition rejects otherwise valid SSPR

**Issue:** `resetPasswordV2` calls `verifyNoUserIsSignedIn()` before starting the reset. iOS starts the server-driven reset without a cached-account precondition.

Locations:
- `msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt:646-655,782-810`
- iOS references: `MSALNativeAuthPublicClientApplication.swift:302-312` and `MSALNativeAuthFlowController.swift:143-160`

**Impact:** Android cannot reset a signed-in account's password or start a reset for another identifier until the app signs out, and it uses a platform-specific legacy error path.

**Recommendation:** Remove the local cached-account precondition and project any service rejection through the unified v2 result contract.

### Severity: Medium – Authentication-method selection does not enforce email OTP

**Issue:** Android uses `methods.firstOrNull()` and carries that method into challenge/verify continuation. iOS explicitly selects an email method and rejects non-email challenge results because v2 SSPR currently supports email OTP only.

Locations:
- `common/common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/v2/NativeAuthV2ResponseParser.kt:211-252`
- `common/common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/v2/NativeAuthV2ContinuationState.kt:128-138`
- `msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/CodeRequiredStateV2.kt:246-254`
- iOS reference: `MSALNativeAuthFlowController.swift:168-184,780-833`

**Impact:** If a non-email method is advertised first, Android can challenge an unsupported method or expose the wrong channel even when a valid email method is available. The mismatch can recur on resend.

**Recommendation:** Select the first email-capable method, fail when none exists, and validate the email channel on initial and resent `CodeRequired` results.

djanardhan98:
@antrix1989 

Fixed in latest commit here and https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/3235

1. Invalid OTP recoverability: Common now maps  invalidGrant + invalidOneTimeCode  to retryable  InvalidCode ; invalid continuation tokens remain terminal.
This restores the retry state for mistyped OTPs and prevents retries with unusable continuation tokens.

2. Post-reset scopes and claims: Scopes and claims now belong to the explicit  signIn(parameters)  step rather than reset start.
MSAL passes them at invocation time, and Common uses the same effective values for token acquisition and cache persistence.

3. Cached-account precondition: The precondition is currently retained pending the Android API design decision.
Its failure now returns ResetPasswordErrorV2 through the unified v2 result contract instead of escaping as an exception. We need an explicit API decision on whether SSPR should be independent of cached sign-in state before making that behavioral change. Android v1 requires users to be signed out before starting password reset, returning a client error when an account is already cached. The v2 behavior is pending while we decide whether to preserve v1 compatibility or intentionally allow SSPR independently of cached sign-in state. @spetrescu84 to chime in.

4. Email OTP selection: Common now selects the first email-capable method and rejects responses without a supported email method. The selected method and channel are validated consistently for initial and resent code-required results.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/CodeErrorsV2.kt
```diff
@@ -39,6 +40,7 @@ import com.microsoft.identity.nativeauth.statemachine.NativeAuthFlowScenarioV2
  * @param errorCodes a list of specific error codes returned by the authentication server.
  * @param subError the sub error returned by the authentication server.
  * @param exception an internal unexpected exception that happened.
+ * @param nextState populated for recoverable errors (e.g. invalid code) so the caller can retry the same step without restarting the flow; null for terminal errors.
```
spetrescu84: This is the public class yes? We don't return a state here or up to here, the developer must use the state from before. We removed this on iOS from the interface, if still there on Android it can be removed. It's a divergence from the v1 way.
See here - https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2547/changes#diff-99ee3883428c080d3d88ac32b562f178c49bcd51c0ab91a60eb7d16f9298337c

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/NativeAuthErrorV2.kt
```diff
@@ -43,6 +44,7 @@ import com.microsoft.identity.nativeauth.statemachine.results.NativeAuthResultV2
  * @param scenario identifies which part of the Native Auth V2 surface produced this error.
  * @param errorCodes a list of specific error codes returned by the authentication server.
  * @param exception an internal unexpected exception that happened.
+ * @param nextState populated for recoverable errors (e.g. invalid code, password policy violation) so the caller can retry the same step without restarting the flow; null for terminal errors.
```
spetrescu84: not actually needed, developer can use from before

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/ResetPasswordErrorsV2.kt
```diff
@@ -66,6 +67,7 @@ class ResetPasswordErrorV2(
  * @param errorCodes a list of specific error codes returned by the authentication server.
  * @param subError the sub error returned by the authentication server.
  * @param exception an internal unexpected exception that happened.
+ * @param nextState populated for recoverable errors (e.g. password policy violation) so the caller can retry the same step without restarting the flow; null for terminal errors.
```
spetrescu84: not required see above

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/ResetPasswordErrorsV2.kt
```diff
@@ -75,8 +77,9 @@ class SubmitNewPasswordErrorV2(
     scenario: NativeAuthFlowScenarioV2,
     errorCodes: List<Int>? = null,
     val subError: String? = null,
-    exception: Exception? = null
-) : NativeAuthErrorV2(errorType, error, errorMessage, correlationId, scenario, errorCodes, exception) {
+    exception: Exception? = null,
```
spetrescu84: not required see above

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/CodeRequiredStateV2.kt
```diff
@@ -77,7 +111,96 @@ class CodeRequiredStateV2 internal constructor(
             correlationId = correlationId,
             methodName = "${TAG}.submitCode(code: String)"
         )
-        return notImplemented()
+        val state = continuationState ?: return notImplemented()
```
spetrescu84: Not sure if this is best error here, it's a general error saying internal sdk error or something similar

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/CodeRequiredStateV2.kt
```diff
@@ -104,7 +227,77 @@ class CodeRequiredStateV2 internal constructor(
...
+                    config,
+                    config.oAuth2TokenCache,
+                    state
+                )
+                val command = NativeAuthV2ResendCodeCommand(
+                    parameters,
+                    NativeAuthV2FlowController(),
+                    PublicApiId.NATIVE_AUTH_V2_RESET_PASSWORD_RESEND_CODE
+                )
+                ensureActive()
```
spetrescu84, line 241: This is how threading was done in v1 also yes?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NativeAuthBaseStateV2.kt
```diff
@@ -25,41 +25,95 @@ package com.microsoft.identity.nativeauth.statemachine.states
...
 import com.microsoft.identity.nativeauth.statemachine.errors.ErrorTypes
 import com.microsoft.identity.nativeauth.statemachine.results.NativeAuthResultV2
+import com.microsoft.identity.nativeauth.utils.serializable
 
 /**
  * Base class for the Native Auth V2 states. Each concrete state exposes only the methods that are
  * valid at that point in the flow.
  *
- * @property continuationToken Continuation token passed in the next request.
+ * @property continuationToken Legacy placeholder used only by states that do not yet carry the
```
spetrescu84, line 43: Just seeing this now, there's no way to make NativeAuthV2ContinuationState inherit the v1 one as they both contain `continuationToken`. I know that requires another PR on common core but I think we should do it if feasible, doesn't need to be done as part of this PR

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NativeAuthBaseStateV2.kt
```diff
@@ -25,41 +25,95 @@ package com.microsoft.identity.nativeauth.statemachine.states
...
+            NativeAuthErrorV2(
+                errorMessage = "Password reset completed but no sign-in result was returned.",
+                correlationId = result.correlationId,
+                scenario = scenario
+            )
+        }
+    }
+
+    /**
+     * Writes every field of this base class, including [continuationState]. Subclasses must not
```
spetrescu84: This comment should read.
```
Subclasses must not
     * override this method to append base fields; each subclass's `Parcel` constructor is required
     * to read the fields in exactly this order - continuationToken, correlationId, scenario, config,
     * continuationState - otherwise the reads desynchronise from the writes. Subclasses that never
     * carry a continuation state should read config via [readConfigAndSkipContinuationState].
 ```
Strange about the order, didn't know

File: msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NativeAuthBaseStateV2.kt [hunk 12]
```diff
@@ -25,41 +25,95 @@ package com.microsoft.identity.nativeauth.statemachine.states
...
         parcel.writeString(scenario.name)
         parcel.writeSerializable(config)
+        parcel.writeSerializable(continuationState)
     }
 
     override fun describeContents(): Int = 0
 }
+
+/**
+ * Reads the config field and then drains the trailing continuationState field written by
```
spetrescu84: This comment should read:
```
Used by states that never carry a continuation state, so
 * that the parcel read order stays symmetric with the write order.
```
Rest is implied

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NewPasswordRequiredStateV2.kt
```diff
@@ -26,35 +26,78 @@ package com.microsoft.identity.nativeauth.statemachine.states
...
         scenario = NativeAuthFlowScenarioV2.valueOf(parcel.readString() ?: NativeAuthFlowScenarioV2.UNKNOWN.name),
-        config = parcel.serializable<NativeAuthPublicClientApplicationConfiguration>() as NativeAuthPublicClientApplicationConfiguration
+        config = parcel.serializable<NativeAuthPublicClientApplicationConfiguration>() as NativeAuthPublicClientApplicationConfiguration,
+        continuationState = parcel.serializable<NativeAuthV2ContinuationState>()
     )
 
     interface SubmitNewPasswordCallback : Callback<NativeAuthResultV2>
 
+    /**
+     * Submits a new password to the server; callback variant.
```
spetrescu84, line 93: This comment should read - 
```
     * Submits a new password to the server; callback variant.
     *
     * @param password The new password to submit.
     * @param callback [com.microsoft.identity.nativeauth.statemachine.states.NewPasswordRequiredStateV2.SubmitNewPasswordCallback] to receive the result on.
```

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NewPasswordRequiredStateV2.kt
```diff
@@ -71,13 +114,130 @@ class NewPasswordRequiredStateV2 internal constructor(
         }
     }
 
+    /**
```
spetrescu84, line 114:
```suggestion
    /**
    * Submits a new password to the server; Kotlin coroutines variant.
     *
     * @param password The new password to submit.
     * @return The results of the submit new password action.
     */
```

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NewPasswordRequiredStateV2.kt
```diff
@@ -71,13 +114,130 @@ class NewPasswordRequiredStateV2 internal constructor(
...
         LogSession.logMethodCall(
             tag = TAG,
             correlationId = correlationId,
             methodName = "${TAG}.submitNewPassword(password: CharArray)"
         )
-        return notImplemented()
+        try {
+            return submitNewPasswordInternal(password)
+        } finally {
+            // Common only clears the array once its interactor starts processing the request, so
```
spetrescu84: No need for this comment here, unless present in v1 also

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInAfterResetPasswordStateV2.kt
```diff
@@ -0,0 +1,230 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.ensureActive
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+/**
+ * State reached once a password reset has completed server-side. The token exchange and cache
+ * persistence are deferred until the app explicitly invokes [signIn].
+ */
```
spetrescu84, line 58:
```suggestion
/**
 * State reached once a password reset has completed server-side.
 */
```

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInAfterResetPasswordStateV2.kt
```diff
@@ -0,0 +1,230 @@
...
+        continuationToken = parcel.readString(),
+        correlationId = parcel.readString() ?: "UNSET",
+        scenario = NativeAuthFlowScenarioV2.valueOf(parcel.readString() ?: NativeAuthFlowScenarioV2.UNKNOWN.name),
+        config = parcel.serializable<NativeAuthPublicClientApplicationConfiguration>() as NativeAuthPublicClientApplicationConfiguration,
+        continuationState = parcel.serializable<NativeAuthV2ContinuationState>()
+    )
+
+    interface SignInCallback : Callback<NativeAuthResultV2>
+
+    /**
```
spetrescu84, line 90:
```suggestion
/**
 * Explicit app-invoked sign-in step following a completed password reset flow.
 */
```

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInAfterResetPasswordStateV2.kt
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /**
+     * Explicit app-invoked sign-in step following a completed password reset flow. This is the only
+     * method that triggers the token exchange and cache persistence for the reset flow; callback variant.
+     *
+     * @param parameters parameters used for the sign-in-after-reset-password operation. Scopes and
+     * claims supplied here take precedence over any supplied at the start of the reset flow.
+     * @param callback [SignInCallback] to receive the result on.
+     */
```
spetrescu84, line 115:
```suggestion
    /**
     * Explicit app-invoked sign-in step following a completed password reset flow, callback variant.
     *
     * @param parameters parameters used for the sign-in-after-reset-password operation. Scopes and
     * claims supplied here take precedence over any supplied at the start of the reset flow.
     * @param callback [SignInCallback] to receive the result on.
     */
```

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInAfterResetPasswordStateV2.kt
```diff
@@ -0,0 +1,230 @@
...
+                Logger.error(TAG, "Exception thrown in signIn", e)
+                callback.onError(e)
+            }
+        }
+    }
+
+    /**
+     * Explicit app-invoked sign-in step following a completed password reset flow. This is the only
+     * method that triggers the token exchange and cache persistence for the reset flow.
+     */
```
spetrescu84, line 135:
```suggestion
    /**
     * Explicit app-invoked sign-in step following a completed password reset flow.
     */
```

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInAfterResetPasswordStateV2.kt
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /**
+     * Explicit app-invoked sign-in step following a completed password reset flow. This is the only
+     * method that triggers the token exchange and cache persistence for the reset flow; Kotlin
+     * coroutines variant.
+     *
+     * @param parameters parameters used for the sign-in-after-reset-password operation. Scopes and
+     * claims supplied here take precedence over any supplied at the start of the reset flow.
+     */
```
spetrescu84, line 151:
```suggestion
    /**
     * Explicit app-invoked sign-in step following a completed password reset flow, Kotlin coroutines      
     * variant.
     *
     * @param parameters parameters used for the sign-in-after-reset-password operation. Scopes and
     * claims supplied here take precedence over any supplied at the start of the reset flow.
     */
```

msal/src/main/java/com/microsoft/identity/nativeauth/utils/CoroutineExtensions.kt
```diff
@@ -0,0 +1,38 @@
...
+//  THE SOFTWARE.
+
+package com.microsoft.identity.nativeauth.utils
+
+import com.microsoft.identity.common.java.result.FinalizableResultFuture
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.runInterruptible
+
+internal suspend fun <T> FinalizableResultFuture<T>.getCancellable(): T {
```
spetrescu84, line 31: Where is this used?

spetrescu84 (dismissed):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NewPasswordRequiredStateV2.kt
djanardhan98: Removed the inline cleanup comment to match the V1 style.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NativeAuthBaseStateV2.kt
djanardhan98: Updated the `writeToParcel` KDoc to the requested wording.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NewPasswordRequiredStateV2.kt
djanardhan98: Updated the callback KDoc to the requested text.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NativeAuthBaseStateV2.kt
djanardhan98: Shortened the helper KDoc to describe only the parcel-order requirement.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/CodeRequiredStateV2.kt
djanardhan98: Yes, the core threading model matches V1: callback APIs launch on the shared IO-backed `pcaScope`, and suspend APIs execute command work inside `withContext(Dispatchers.IO)`. The V2 implementation intentionally improves cancellation, though: V1 waits with plain `Future.get()`, whereas V2 uses `getCancellable()` and propagates `CancellationException`, allowing coroutine cancellation to interrupt the wait and signal Common to cancel the active HTTP request. So the execution dispatcher is consistent with V1, while cancellation handling is stronger in V2.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NativeAuthBaseStateV2.kt
djanardhan98: Created item to track 

https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3739969

msal/src/main/java/com/microsoft/identity/nativeauth/utils/CoroutineExtensions.kt
djanardhan98: It is used by all five new V2 SSPR command waits: reset-password start, submit code, resend code, submit new password, and sign-in after reset. It makes the blocking `FinalizableResultFuture.get()` coroutine-cancellable and forwards cancellation to Common so active HTTP work is stopped. Existing Native Auth paths use plain `.get()`, which does not provide that behavior.

djanardhan98: **The helper is used and is necessary only to make the new V2 SSPR suspend APIs cooperatively cancellable.**

`getCancellable()` performs two distinct cancellation operations:

1. `runInterruptible` interrupts the coroutine’s thread blocked in `Future.get()`.

2. `cancelSignal()` tells Common to stop the underlying command/HTTP connection.



Both matter: interrupting `get()` alone does not stop network work, while signalling Common alone may not promptly unblock the waiting coroutine. It then rethrows `CancellationException`, preserving coroutine semantics.


The existing Native Auth V1 code uses `withContext(Dispatchers.IO) { future.get() }`. That prevents main-thread blocking but is **not cancellable while `get()` is waiting**. The surrounding `ensureActive()` calls only detect cancellation before and after the wait; they cannot interrupt it. Also, `ResultFuture.cancel(true)` cannot be reused because its implementation returns `false` and does nothing.


The helper was added in the follow-up commit titled **“Make Native Auth V2 command waits cancellable”**. Its regression test cancels each of the five operations and verifies both that the coroutine completes promptly and that Common’s HTTP cancellation signal is activated. Common introduced that signal specifically to disconnect in-flight HTTP work and avoid abandoned worker activity.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/CodeRequiredStateV2.kt
djanardhan98: Let me know if that sounds okay!

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/CodeErrorsV2.kt
djanardhan98: Removed `nextState` from `SubmitCodeErrorV2`. The caller can retry using the existing `CodeRequiredStateV2`, which keeps this aligned with the intended V1/iOS contract.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/ResetPasswordErrorsV2.kt
djanardhan98: Removed the `nextState` documentation because submit-new-password errors no longer return retry state.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/NativeAuthErrorV2.kt
djanardhan98: Removed `nextState` from the base `NativeAuthErrorV2` type. Retry state is no longer exposed through the V2 error hierarchy.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/CodeRequiredStateV2.kt
djanardhan98: Changed this fallback from `NOT_IMPLEMENTED` to `INVALID_STATE`. Both submit-code and resend-code now return actionable guidance to restart the flow when the continuation state is unavailable.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/ResetPasswordErrorsV2.kt
djanardhan98: Removed `nextState` from `SubmitNewPasswordErrorV2` and removed its producer assignments. Callers retain and reuse their existing `NewPasswordRequiredStateV2`.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NewPasswordRequiredStateV2.kt
```diff
@@ -71,13 +111,123 @@ class NewPasswordRequiredStateV2 internal constructor(
...
+     */
     suspend fun submitNewPassword(password: CharArray): NativeAuthResultV2 {
         LogSession.logMethodCall(
             tag = TAG,
             correlationId = correlationId,
             methodName = "${TAG}.submitNewPassword(password: CharArray)"
         )
-        return notImplemented()
+        try {
+            return submitNewPasswordInternal(password)
```
fadidurah, line 127: Kind of a nit, but we could use password.copyOf() here. If the coroutine is cancelled after the Common worker starts,  getCancellable()  returns immediately and MSAL clears that same array in `finally` , while Common may still be serializing it for the network request. This race can submit an empty or partially zeroed password while the client reports cancellation. We would however also need to clear the second password object created by `copyOf()` in common. I think this is fairly minor, and we can skip it, but figured i'd point it out.

fadidurah (approved):

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/SignInAfterResetPasswordStateV2.kt
```diff
@@ -0,0 +1,225 @@
...
+
+    private suspend fun internalSignIn(signInParameters: NativeAuthSignInContinuationParameters?): NativeAuthResultV2 {
+        val state = continuationState ?: return notImplemented()
+        return withContext(Dispatchers.IO) {
+            try {
+                val parameters = CommandParametersAdapter.createNativeAuthV2SignInAfterResetPasswordCommandParameters(
+                    config,
+                    config.oAuth2TokenCache,
+                    state,
+                    signInParameters?.scopes,
```
fadidurah, line 170: Sanity check: Wondering if this is intentional. We use the parameter-supplied scopes at this layer, but in NativeAuthV2FlowController.completeFlow() they are replaced with `val scopes = state.scopesForTokenRequest()`. This was a pretty impressive catch by ai, to point this out. just making sure this is intentional, we might be ending up with a token that doesn't match the scopes requested by the caller.

fadidurah: Ahh, seems that's what the second PR is doing. Ok, please resolve this if we've already addressed this fully.

djanardhan98: Confirmed this was not intentional. Common PR #3235 now derives the effective scopes from the terminal sign-in parameters, adds the default scopes, and uses the same scope set for both token acquisition and cache persistence.

https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/3235

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/NewPasswordRequiredStateV2.kt
djanardhan98: Good catch, thanks for flagging it. I dug into it and I think we're safe here, so I'm going to skip the copy — but the reason is a bit subtle, so writing it down:


Common doesn't check for cancellation before it serializes the password — `NativeAuthV2Interactor:386` reads the array with no guard. What saves us is a check further down, in `UrlConnectionHttpClient.executeHttpSend:371`. That check happens after the body is built but before anything is actually sent.


Since we always call `cancelSignal()` before MSAL clears the array, any timing that could corrupt the request body also guarantees that check sees the cancellation and bails out with `TIMED_OUT`. So we might build a bad body, but we can never send one.

spetrescu84 (approved):

fadidurah (approved):

PR #2563, opened 2026-08-31
Implement Native Auth V2 sign-in with password and email OTP MFA, Fixes AB#3729067

djanardhan98:
## Summary

Implements Native Auth V2 sign-in in MSAL for password first-factor flows and password followed by email OTP MFA, aligned with the current iOS V2 service contract.

## Supported flows

- Username with password supplied at `signInV2` entry
- Username followed by deferred password submission through `PasswordRequiredStateV2`
- Password followed by explicit email OTP MFA method selection and OTP verification
- Invalid MFA code recovery by requesting a fresh challenge
- Authorization-code exchange, token caching, requested scopes/claims, correlation IDs, browser-required results, Kotlin suspend APIs, and Java callbacks

## Behavior and error handling

- Entry-supplied wrong password maps to `SignInErrorV2.isInvalidCredentials()`.
- Deferred wrong password maps to `SubmitPasswordErrorV2.isInvalidPassword()`.
- Recoverable failures return result values without embedding a next state; callers retry using the retained state.
- Missing or invalid continuation state maps to `INVALID_STATE`.
- Unsupported MFA channels map to `NOT_IMPLEMENTED`.
- Preserves Android V1 account-cache behavior by rejecting `signInV2` while an account is already signed in.
- Password buffers are copied and cleared in `finally`; passwords are not retained in Parcelable or continuation state.

## Scope exclusions

- Email OTP as the first factor
- SMS OTP, passkeys, sign-up, JIT/strong-auth registration, and silent-refresh changes
- MFA returned only from the final token exchange
- Repeated same-account or switch-account sign-in

## Validation

- MSAL V2 unit suites: **85 tests passed, 0 failures**, including 26 new sign-in scenarios.
- Full MSAL unit-suite comparison against `dev`: **no new failures**. Remaining failures in both runs are confined to existing network tests that require the CI mock-API endpoint.
- Common V2 suites: **245 tests passed, 0 failures**.
- Common full `:common4j:test`: **1,757 tests passed, 0 failures**.
- `:common4j:spotbugsMain`: **0 violations**.

## E2E coverage

Adds eight public-API E2E scenarios in `SignInV2EmailPasswordTest`. They compile but remain `@Ignore`-gated because the V2 authorize-challenge sign-in endpoints are not available on the current test slice and the secure lab configuration is only available in CI. The tests can be enabled unchanged when a V2-capable slice is available.

Fixes [AB#3729067](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3729067)

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
@@ -1348,7 +1348,7 @@ public class CommandParametersAdapter {
      * @param claimsRequest (Optional) claims request to send with the token request
      * @return Command parameter object
      */
-    public static SignInV2StartCommandParameters createSignInV2StartCommandParameters(
+    public static SignInV2StartCommandParameters createNativeAuthV2SignInCommandParameters(
             @NonNull final NativeAuthPublicClientApplicationConfiguration configuration,
             @NonNull final OAuth2TokenCache tokenCache,
             @NonNull final String username,
--- a/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
@@ -66,9 +66,12 @@ import com.microsoft.identity.common.java.nativeauth.authorities.NativeAuthCIAMA
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.MFAChallengeAuthMethodCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.MFASubmitChallengeCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.NativeAuthV2ResendCodeCommandParameters;
+import com.microsoft.identity.common.java.nativeauth.commands.parameters.NativeAuthV2SelectMFAMethodCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.NativeAuthV2SignInAfterResetPasswordCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.NativeAuthV2SubmitCodeCommandParameters;
+import com.microsoft.identity.common.java.nativeauth.commands.parameters.NativeAuthV2SubmitMFAChallengeCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.NativeAuthV2SubmitNewPasswordCommandParameters;
+import com.microsoft.identity.common.java.nativeauth.commands.parameters.NativeAuthV2SubmitPasswordCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.ResetPasswordV2StartCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.ResetPasswordResendCodeCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.ResetPasswordStartCommandParameters;
@@ -78,6 +81,7 @@ import com.microsoft.identity.common.java.nativeauth.commands.parameters.SignInR
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.SignInStartCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.SignInSubmitCodeCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.SignInSubmitPasswordCommandParameters;
+import com.microsoft.identity.common.java.nativeauth.commands.parameters.SignInV2StartCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.SignInWithContinuationTokenCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.SignUpResendCodeCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.SignUpStartCommandParameters;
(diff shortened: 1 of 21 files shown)


msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/MFAVerificationRequiredStateV2.kt
```diff
@@ -26,36 +26,78 @@ package com.microsoft.identity.nativeauth.statemachine.states
...
 
 /**
  * State that requires the user to submit a multi-factor authentication challenge.
+ *
+ * A wrong code is a recoverable [MFASubmitChallengeErrorV2] with
+ * [MFASubmitChallengeErrorV2.isInvalidChallenge]; the caller retries on this same state instance,
+ * or requests a fresh challenge from the [MFARequiredStateV2] it still holds. No error result
+ * carries a next state.
  */
 class MFAVerificationRequiredStateV2 internal constructor(
```
spetrescu84, line 67: We are missing resend challenge here. That's mostly because how I split this into 2 states. So you can add a new method here to submitChallenge to the same one as before

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/MFAVerificationRequiredStateV2.kt
```diff
@@ -72,13 +114,100 @@ class MFAVerificationRequiredStateV2 internal constructor(
...
+                    parameters,
+                    NativeAuthV2FlowController(),
+                    PublicApiId.NATIVE_AUTH_V2_SIGN_IN_SUBMIT_MFA_CHALLENGE
+                )
+                ensureActive()
+                val rawCommandResult = CommandDispatcher.submitSilentReturningFuture(command).getCancellable()
+                ensureActive()
+                when (val result = rawCommandResult.checkAndWrapCommandResultType<NativeAuthV2SubmitMFAChallengeCommandResult>()) {
+                    is NativeAuthV2CommandResult.Complete -> {
+                        mapCompleteResult(result)
```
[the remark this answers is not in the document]
djanardhan98: Fixed in `718f8f54c`. The shared completion mapper now reports the scenario-neutral message `Native Auth V2 flow completed but no authentication result was returned.` Regression coverage exercises the MFA sign-in completion path.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/SignInErrorsV2.kt
```diff
@@ -78,5 +78,25 @@ class SubmitPasswordErrorV2(
...
+
+    /**
+     * Always false for this error.
+     *
+     * A deferred password rejection is reported as [isInvalidPassword], matching MSAL iOS/macOS:
+     * the account was already accepted at the sign-in entry point, so only the password was wrong.
+     * Invalid credentials are reported by
+     * [com.microsoft.identity.nativeauth.INativeAuthPublicClientApplication.signInV2] as a
+     * [SignInErrorV2] instead. Retained only for source compatibility.
+     */
```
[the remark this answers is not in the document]
djanardhan98: Fixed in `718f8f54c`. The documentation now matches the implementation: `isInvalidCredentials()` classifies an `INVALID_CREDENTIALS` error, while SDK-produced deferred-password failures are surfaced through `isInvalidPassword()`.

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -599,7 +604,167 @@ class NativeAuthPublicClientApplication(
             correlationId = null,
             methodName = "${TAG}.signInV2(parameters: NativeAuthSignInParameters)"
         )
-        return notImplementedV2(NativeAuthFlowScenarioV2.SIGN_IN)
+        return withContext(Dispatchers.IO) {
+            var correlationId = DiagnosticContext.INSTANCE.threadCorrelationId ?: "UNSET"
+            // Copied so the caller keeps ownership of its own buffer, while this copy is cleared
+            // below on every exit path, including cancellation. A password is never stored in a
+            // Parcelable state or in the opaque continuation state.
+            val passwordCopy = parameters.password?.copyOf()
```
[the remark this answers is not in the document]
djanardhan98: Fixed in `718f8f54c`. The callback entry point now snapshots the password synchronously before launching asynchronous work. The owned snapshot is cleared by a job completion handler, including cancellation before the coroutine body starts; tests cover immediate caller mutation and cleanup.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/MFAVerificationRequiredStateV2.kt
djanardhan98: Implemented in `9a5c1768a`. `MFAVerificationRequiredStateV2` now exposes callback and suspend `resendChallenge()` APIs. Resend uses the verification state's current continuation and returns a fresh verification state with the refreshed continuation and challenge metadata. Success, error, callback, and cancellation paths are covered.

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -1331,6 +1335,213 @@ public static NativeAuthV2SignInAfterResetPasswordCommandParameters createNative
...
+                        .authority(authority)
+                        .authenticationScheme(authenticationScheme)
+                        .challengeType(configuration.getChallengeTypes())
+                        .requestInterceptor(configuration.getRequestInterceptor())
+                        .capabilities(configuration.getCapabilities())
+                        .username(username)
+                        .password(password)
+                        .scopes(scopes)
+                        .claimsRequestJson(claimsRequestJson)
+                        .correlationId(DiagnosticContext.INSTANCE.getThreadCorrelationId())
```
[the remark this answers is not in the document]
djanardhan98: Fixed in `718f8f54c` and completed in `e164fa50a`. `CommandParametersTest` now verifies username/password, scopes, claims, correlation ID, continuation state, MFA method/code, and token cache across sign-in start and follow-up adapters. The final coverage pass also found and fixed claims propagation for the resend adapter.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/MFARequiredStateV2.kt
```diff
@@ -26,37 +26,101 @@ package com.microsoft.identity.nativeauth.statemachine.states
...
     scenario: NativeAuthFlowScenarioV2,
-    config: NativeAuthPublicClientApplicationConfiguration
-) : NativeAuthBaseStateV2(continuationToken, correlationId, scenario, config) {
+    config: NativeAuthPublicClientApplicationConfiguration,
+    continuationState: NativeAuthV2ContinuationState? = null,
+
+    /**
+     * The authentication methods the server offered for this multi-factor step.
+     */
+    val authMethods: List<AuthMethod> = emptyList()
```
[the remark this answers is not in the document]
djanardhan98: Fixed in `718f8f54c`. Both `MFARequiredStateV2` and `NativeAuthResultV2.MFARequired` now store defensive copies wrapped with `Collections.unmodifiableList`. Regression tests verify source-list mutation cannot alter the exposed methods and Java-style mutation is rejected.

msal/src/test/java/com/microsoft/identity/nativeauth/v2/NativeAuthV2SignInTest.kt
```diff
@@ -0,0 +1,785 @@
...
+            ),
+            NativeAuthV2SignInStartCommand::class
+        )
+
+        val callerPassword = "Password123!".toCharArray()
+        application.signInV2(signInParameters(password = callerPassword))
+
+        // signInV2 copies the buffer, so the caller's own array is untouched while the copy handed
+        // to Common is cleared. A password is never stored in a state or continuation state.
+        assertEquals("Password123!", String(callerPassword))
```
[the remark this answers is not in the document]
djanardhan98: Fixed in `718f8f54c`. The tests now capture the password arrays owned by dispatched commands, verify they are distinct from caller-owned buffers, and assert they are overwritten after completion and cancellation. Separate helper coverage verifies cleanup when cancellation occurs before coroutine execution.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/PasswordRequiredStateV2.kt
```diff
@@ -72,13 +117,118 @@ class PasswordRequiredStateV2 internal constructor(
...
+                errorType = ErrorTypes.INVALID_PASSWORD,
+                errorMessage = "Password cannot be empty.",
+                correlationId = correlationId,
+                scenario = scenario
+            )
+        }
+        return withContext(Dispatchers.IO) {
+            // Copied so the caller keeps ownership of its own buffer, while this copy is cleared
+            // below on every exit path, including cancellation.
+            val passwordCopy = password.copyOf()
```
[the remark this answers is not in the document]
djanardhan98: Fixed in `718f8f54c`. `submitPassword(..., callback)` copies the password at the synchronous callback boundary and transfers ownership to a launch helper that clears it on every job completion path, including cancellation before execution. The command-bound copy remains separately cleared in `finally`.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/PasswordRequiredStateV2.kt
```diff
@@ -72,13 +117,118 @@ class PasswordRequiredStateV2 internal constructor(
...
+                    NativeAuthV2FlowController(),
+                    PublicApiId.NATIVE_AUTH_V2_SIGN_IN_SUBMIT_PASSWORD
+                )
+                ensureActive()
+                val rawCommandResult = CommandDispatcher.submitSilentReturningFuture(command).getCancellable()
+                ensureActive()
+                when (val result = rawCommandResult.checkAndWrapCommandResultType<NativeAuthV2SubmitPasswordCommandResult>()) {
+                    is NativeAuthV2CommandResult.Complete -> {
+                        mapCompleteResult(result)
+                    }
```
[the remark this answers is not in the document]
djanardhan98: Fixed in `718f8f54c`. The shared missing-authentication-result diagnostic is now scenario-neutral rather than password-reset-specific, with regression assertions for both password and MFA sign-in completion paths.

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
```diff
@@ -1331,6 +1336,213 @@ public static NativeAuthV2SignInAfterResetPasswordCommandParameters createNative
...
+     * @param configuration PCA configuration
+     * @param tokenCache token cache for storing results
+     * @param username username of the account signing in
+     * @param password (Optional) password to submit for the password first factor. The caller's
+     *                 array is passed through unchanged so the same buffer can be cleared later.
+     * @param scopes (Optional) scopes to request for the resulting access token
+     * @param claimsRequest (Optional) claims request to send with the token request
+     * @return Command parameter object
+     */
+    public static SignInV2StartCommandParameters createSignInV2StartCommandParameters(
```
fadidurah: "createNativeAuthV2SignInCommandParameters"? Just to match the naming of the other method

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/MFAErrorsV2.kt
```diff
@@ -78,5 +78,9 @@ class MFASubmitChallengeErrorV2(
     exception: Exception? = null
 ) : NativeAuthErrorV2(errorType, error, errorMessage, correlationId, scenario, errorCodes, exception) {
 
-    fun isInvalidChallenge(): Boolean = this.errorType == ErrorTypes.INVALID_CHALLENGE
```
fadidurah, line 81: This wasn't being used?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/PasswordRequiredStateV2.kt
```diff
@@ -26,59 +26,210 @@ package com.microsoft.identity.nativeauth.statemachine.states
...
+                    is NativeAuthV2CommandResult.MFARequired -> {
```
fadidurah, line 166: My understanding was that we are adopting V2 because it was moving away from the whole error code matching style, and sort of relying on getting specific error codes from server to route flows through the library code. is this not the case? What's the main benefit of v2 otherwise?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/MFAErrorsV2.kt
```diff
@@ -78,5 +78,9 @@ class MFASubmitChallengeErrorV2(
...
+    /**
+     * Returns true when the submitted MFA one-time code was rejected and the caller can retry on
+     * the same [com.microsoft.identity.nativeauth.statemachine.states.MFAVerificationRequiredStateV2].
+     */
+    fun isInvalidCode(): Boolean = this.errorType == ErrorTypes.INVALID_CODE
```
spetrescu84: I know invalid code comes back from the server but here we should surface `isInvalidChallenge` as we actually have `submitChallenge` on the state so the normal naming is `isInvalidChallenge`

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/SignInErrorsV2.kt
```diff
@@ -78,5 +78,25 @@ class SubmitPasswordErrorV2(
...
+    /**
+     * Returns true if the password submitted through
+     * [com.microsoft.identity.nativeauth.statemachine.states.PasswordRequiredStateV2.submitPassword]
+     * was rejected. The caller can retry on the same state instance.
+     */
+    fun isInvalidPassword(): Boolean = this.errorType == ErrorTypes.INVALID_PASSWORD
+
+    /**
+     * A deferred password rejection is reported as [isInvalidPassword], matching MSAL iOS/macOS:
+     * the account was already accepted at the sign-in entry point, so only the password was wrong.
```
spetrescu84:
```suggestion
     * A deferred password rejection is reported as [isInvalidPassword], the account was already accepted 
     * at the sign-in entry point, so only the password was wrong.
```

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/results/NativeAuthResultV2.kt
```diff
@@ -145,8 +146,10 @@ interface NativeAuthResultV2 : Result {
     class MFARequired(
         override val nextState: MFARequiredStateV2,
         override val scenario: NativeAuthFlowScenarioV2,
-        val authMethods: List<AuthMethod>
-    ) : Result.SuccessResult(nextState = nextState), NativeAuthResultV2
+        authMethods: List<AuthMethod>
+    ) : Result.SuccessResult(nextState = nextState), NativeAuthResultV2 {
+        val authMethods: List<AuthMethod> = Collections.unmodifiableList(ArrayList(authMethods))
```
spetrescu84: I don't think this is used right?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/MFARequiredStateV2.kt
```diff
@@ -26,37 +26,103 @@ package com.microsoft.identity.nativeauth.statemachine.states
...
+        continuationToken = null,
+        correlationId = continuationState.correlationId,
+        scenario = scenario,
+        config = config,
+        continuationState = continuationState,
+        authMethods = authMethods
+    )
 
     private constructor(parcel: Parcel) : this(
-        continuationToken = parcel.readString() ?: "",
```
spetrescu84, line 50: What's happening here and why did these constructors change. Also looking at the code why this method doesn't exist for all states?

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
```diff
@@ -608,9 +774,14 @@ class NativeAuthPublicClientApplication(
             correlationId = null,
             methodName = "${TAG}.signInV2(parameters: NativeAuthSignInParameters, callback: NativeAuthV2Callback)"
         )
-        pcaScope.launch {
+        val parametersSnapshot = NativeAuthSignInParameters(parameters.username).also {
+            it.password = parameters.password?.copyOf()
+            it.scopes = parameters.scopes?.toList()
+            it.claimsRequest = parameters.claimsRequest
+        }
+        pcaScope.launchOwningPasswordSnapshot(parametersSnapshot.password) {
```
spetrescu84, line 781: I don't think this is needed. Password sensitivity is no different than in submitPassword or submitNewPassword. It should not be stored past the initial submission but that can be checked in the continuation state as it can't live anywhere else

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/MFAErrorsV2.kt
djanardhan98: Updated in 88034f629. `MFASubmitChallengeErrorV2` now exposes `isInvalidChallenge()` again, matching `submitChallenge` and the base API, while still comparing against `ErrorTypes.INVALID_CODE` so the V2 server classification remains correct. Updated the V2 MFA docs and tests accordingly.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/PasswordRequiredStateV2.kt
djanardhan98: Hey, yeah great question
So the idea behind v2 is that it's server driven yes, and the errors are returned by the server, but the client is still responsible for parsing those errors and mapping to client specific codes and surface appropriate UX, as well as returning a suitable error for a public SDK contract basically
So the raw server errors are then mapped to client error states, and the actions to take are given by the server as well but the client is still responsible for using that and base the UI lifecycle and state retention based on it. For eg there's codes which either allow restart flow or retry the same step like wrong OTP entered for eg, and the client would still be responsible for owning that lifecycle based on server's response 
In v1 the client would be responsible for determining the error scenario itself to deciding the next step in the flow. For eg step 1 -> server has some response based on which client determines whether its an error or not and what error it is -> client decides what the next step should be  and does that. Now its step 1 -> server tells us exactly what the error is if any and what the next step is -> client maps to public SDK error states and just performs the action. The server tells us for eg that "the password is incorrect", but now the sdk still needs to know 
So something like PasswordRequiredStateV2 is mapping an internal server result to a public MSAL api type. This is mainly because this mapping is contextual based on the flow which the sdk owns

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/MFAErrorsV2.kt
djanardhan98: Just renamed

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
djanardhan98: This was based on comment from copilot review 

https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2563#discussion_r3907478734

Without the snapshot:

1. The app gives MSAL a password array.
2. MSAL schedules the sign-in work and immediately returns.
3. The app clears or changes the password array.
4. MSAL’s background work starts and reads the now-cleared or changed password.
5. Sign-in fails because MSAL sends the wrong password.

The snapshot fixes this by copying the password before returning to the app. MSAL’s background work then uses its own copy, so the app can safely clear its original password array immediately.

Let me know if you think this makes sense or is even possible in a real scenario

msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
djanardhan98: Renamed this to `createNativeAuthV2SignInCommandParameters` in `4efdf8684` and updated its call site and adapter test. This now follows the neighboring Native Auth V2 adapter naming.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/MFAErrorsV2.kt
djanardhan98: This was an intermediate rename rather than an unused helper. The original V2 API exposed `isInvalidChallenge()` with `INVALID_CHALLENGE`; the resend update temporarily changed it to `isInvalidCode()` with the V2 service value `INVALID_CODE`; `88034f629` restored the established public MFA name while retaining the correct V2 mapping. The final `isInvalidChallenge() == (errorType == INVALID_CODE)` combination is intentional.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/results/NativeAuthResultV2.kt
djanardhan98: I changed the way we're using it to avoid duplicating the authMethods list, did you mean something else?

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/MFARequiredStateV2.kt
djanardhan98: Compared with `dev`, `SelectAuthMethodCallback` and both `selectAuthMethod` overloads already existed; the suspend overload previously returned `notImplemented()`. This PR implements that existing operation. The constructor changes retain the opaque `NativeAuthV2ContinuationState` and server-offered methods, add a production constructor that derives the correlation ID, update parcel restoration, and add `writeToParcel` for the methods. Those changes are needed now because the implemented selection command must follow the server and per-method HAL links; the old token-only state was sufficient only while this was a stub. The production continuation-state constructor pattern is used by the password, MFA-selection, and MFA-verification states implemented in this increment. Other states retain their existing constructors until their server-driving operations are implemented. Each state exposes only actions valid at that point, so method selection belongs on MFA/strong-auth selection states rather than every state.

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
djanardhan98: AI summary : This snapshot was added in `718f8f54c` in response to the callback-boundary race identified in [this review thread](https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2563#discussion_r3907478734). The callback API schedules a coroutine and returns before its body necessarily starts; if the caller immediately clears or mutates the supplied `CharArray`, the later internal copy can capture the cleared/changed value. Taking the snapshot synchronously preserves the submitted password, and the completion hook clears it even when cancellation occurs before the coroutine body runs. The same issue and fix apply to `PasswordRequiredStateV2.submitPassword` ([related thread](https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/2563#discussion_r3907478937)). The continuation state contains no password, but that addresses cross-state retention rather than this pre-execution race. Older `submitNewPassword` code does not have the same protection, so behavior is currently inconsistent; I recommend retaining this unless the API contract explicitly requires callers to keep the buffer unchanged until asynchronous consumption.

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/results/NativeAuthResultV2.kt
djanardhan98: My understanding is that the app would use  result.authMethods  to show the available MFA options, then use  nextState  to submit the selected one. Could you confirm if that’s the intended usage?

spetrescu84: Yes, only distinction is that we want to align with iOS/JS when only one method is possible (but only for MFA/SSPR) then we automatically invoke.

msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
spetrescu84: Ok, then we should retain this and also apply to SSPR/SignUp if needed

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/errors/SignInErrorsV2.kt
```diff
@@ -78,5 +78,25 @@ class SubmitPasswordErrorV2(
...
+     * [com.microsoft.identity.nativeauth.INativeAuthPublicClientApplication.signInV2] as a
+     * [SignInErrorV2] instead. This method returns true only when this error was explicitly
+     * constructed with [SignInErrorTypes.INVALID_CREDENTIALS], and is retained for source
+     * compatibility.
+     */
+    @Deprecated(
+        message = "A password rejected by PasswordRequiredStateV2.submitPassword is an invalid password, not invalid credentials. Use isInvalidPassword() instead.",
+        replaceWith = ReplaceWith("isInvalidPassword()")
+    )
     fun isInvalidCredentials(): Boolean = this.errorType == SignInErrorTypes.INVALID_CREDENTIALS
```
spetrescu84: Remove this completely, no need for the comment, the only error here should be `isInvalidPassword`, my PR was incorrect

djanardhan98: Removed SubmitPasswordErrorV2.isInvalidCredentials() entirely, including its KDoc and annotation, in 8ab480ad9. isInvalidPassword() remains; SignInErrorV2.isInvalidCredentials() is unchanged.

djanardhan98:
@copilot resolve the merge conflicts in this pull request

msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/results/NativeAuthResultV2.kt
djanardhan98: This change spans Common orchestration and MSAL’s public API. To avoid splitting the behavior across multiple partial updates, I propose keeping this PR scoped to sign-in with explicit MFA selection and delivering the complete MFA/SSPR auto-selection and unified selection API in coordinated Common and MSAL follow-up PRs under AB#3747379.

djanardhan98: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3747379

PR #2567, opened 2026-09-04
Temporarily support current and incoming common4j continuation-state constructors in MSAL tests, Fixes AB#3749428

djanardhan98:
### Summary

MSAL Native Auth V2 tests construct `NativeAuthV2ContinuationState` reflectively because common4j keeps its constructor private. The tests selected a constructor with exactly seven parameters, so common4j PR 3239— which expands it to nine parameters—causes 27 unrelated MSAL test failures in the **Assemble consumers (Validate MSAL)** stage.

This change replaces the duplicated reflection code with one shared test helper that:

- skips Kotlin's synthetic `DefaultConstructorMarker` constructor;
- supports only the current 7-parameter shape and the incoming 9-parameter shape required by common4j consumer validation;
- fails once with a clear error for an unknown future shape instead of producing widespread `NoSuchElementException` failures.

The helper remains test-only. MSAL still treats the continuation state as an opaque common4j-owned object and only verifies that it survives Parcelable transport with its public correlation ID and scopes intact.

### Verification

`msal:testLocalDebugUnitTest` for both affected classes: **32 passed, 0 failed** against:

- currently pinned common4j (`a809870a3`);
- common4j PR 3239 (`e5fd150e0`).

This unblocks common4j 3239 without adding a compatibility constructor to production code. Future constructor shapes still require an explicit helper update, by design, to avoid silently supplying unsafe defaults.

### TODO

Replace this temporary 7/9-parameter compatibility helper with a stable common4j test-fixture factory, then remove constructor reflection from MSAL. Tracked by [AB#3749428](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3749428).

### DIFF
--- a/msal/src/test/java/com/microsoft/identity/nativeauth/v2/NativeAuthV2StatesTest.kt
+++ b/msal/src/test/java/com/microsoft/identity/nativeauth/v2/NativeAuthV2StatesTest.kt
@@ -59,6 +59,7 @@ import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
 import java.util.concurrent.ExecutionException
 import java.util.concurrent.TimeUnit
+import java.lang.reflect.Constructor
 
 /**
  * Unit tests for the Native Auth V2 states, covering Parcelable serialization and the
@@ -156,20 +157,8 @@ class NativeAuthV2StatesTest {
         }
     }
 
-    private fun createContinuationState(): NativeAuthV2ContinuationState {
-        val constructor = NativeAuthV2ContinuationState::class.java.declaredConstructors
-            .single { it.parameterCount == 7 }
-        constructor.isAccessible = true
-        return constructor.newInstance(
-            "opaque-token",
-            emptyMap<String, String>(),
-            listOf("scope"),
-            null,
-            correlationId,
-            NativeAuthV2LinkRelation.RESET_PASSWORD.value,
-            NativeAuthV2FlowScenario.RESET_PASSWORD
-        ) as NativeAuthV2ContinuationState
-    }
+    private fun createContinuationState(): NativeAuthV2ContinuationState =
+        newContinuationState(correlationId)
 
     private fun assertCallbackNotImplemented(action: (ResultFuture<NativeAuthResultV2>) -> Unit) {
         val future = ResultFuture<NativeAuthResultV2>()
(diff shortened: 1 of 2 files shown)


msal/src/test/java/com/microsoft/identity/nativeauth/v2/NativeAuthV2ContinuationStates.kt
```diff
@@ -0,0 +1,122 @@
...
+    val links = emptyMap<String, String>()
+    val methodLinks = emptyMap<String, Map<String, String>>()
+    val claimsRequestJson: String? = null
+    // NativeAuthV2LinkRelation is a Kotlin value class, so it is erased to String on the JVM and
+    // reflection has to be handed the underlying value rather than the boxed relation.
+    val entryRelation = NativeAuthV2LinkRelation.RESET_PASSWORD.value
+    val scenario = NativeAuthV2FlowScenario.RESET_PASSWORD
+    val authenticationFactor: String? = null
+
+    val arguments: Array<Any?> = when (constructor.parameterCount) {
```
shahzaibj: the arg list isn't really derived from the constructor shape here, it's still pinned per arity (7/8/9) with the field order hard coded in each branch. so the next field common4j adds takes this to 10, hits the else, and all 32 fail again.. just with a clearer message this time. isn't the real win here centralizing to one spot + failing loudly, rather than actually being arity agnostic like the description says?

msal/src/test/java/com/microsoft/identity/nativeauth/v2/NativeAuthV2StatesTest.kt
```diff
@@ -434,3 +423,62 @@ class NativeAuthV2StatesTest {
...
+    val links = emptyMap<String, String>()
+    val methodLinks = emptyMap<String, Map<String, String>>()
+    val scopes = listOf("scope")
+    val claimsRequestJson: String? = null
+    // Value classes are erased to their underlying type, so reflection needs the raw String here.
+    val entryRelation = NativeAuthV2LinkRelation.RESET_PASSWORD.value
+    val scenario = NativeAuthV2FlowScenario.RESET_PASSWORD
+    val authenticationFactor: String? = null
+
+    val arguments: Array<Any?> = when (constructor.parameterCount) {
```
shahzaibj: the arg list isn't really derived from the constructor shape here, it's still pinned per arity (7/8/9) with the field order hard coded in each branch. so the next field common4j adds takes this to 10, hits the else, and all 32 fail again.. just with a clearer message this time. also is the 8 branch even reachable? per the description we went 7 -> 9 directly (the 8-param shim was rejected), so maxByOrNull never lands on it. isn't the real win here centralizing to one spot + failing loudly, rather than actually being arity agnostic like the description says?

spetrescu84 (approved):

djanardhan98:
Fixes AB#[3749428](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3749428)

PR #2574, opened 2026-09-09
Enable instrumented test validation for pull requests, Fixes AB#3751467

p3dr0rv:
## Summary

- Enable the MSAL instrumented-test Azure Pipeline to run for pull requests by removing `pr: none` from `pr-msal-instrumented.yml`.
- Keep `trigger: none` so the pipeline does not run for regular branch pushes.
- Remove the legacy GitHub Labeler workflow and its configuration because the generated path-based labels are no longer used.

## Why

The instrumented-test pipeline is intended to validate every pull request, but the explicit `pr: none` setting prevented PR-triggered runs. Removing that setting allows the existing PR validation configuration to queue the pipeline.

The Labeler removal is separate cleanup: its automatically generated labels are not part of the current review or validation process, so keeping the workflow adds maintenance without providing value.

## Work Item

[AB#3751467](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3751467)


djanardhan98:
Title is slightly misleading - we intend to ensure that instrumented tests run on every PR right? Wanted to understand if we want to trigger the tests on every PR without any condition, even if target branch is not dev, for eg?

djanardhan98:
Are we removing labeler step because we don't find value in the labels it's generating?

p3dr0rv:
Update description and title

djanardhan98 (approved):

PR #2579, opened 2026-09-10
Add Native Auth V2 SMS password reset support, Fixes AB#3754595

spetrescu84:
## Summary
- expose reset-password method selection when the service offers both email and SMS
- add a parcelable state with callback and coroutine APIs for selecting an offered method
- map the selected SMS challenge into the existing reset-password code-required flow
- preserve method validation, cancellation, errors, code length, channel, and masked destination
- update the Common submodule to `631fa691d`

## Testing
- Native Auth V2 interface, result, and state unit tests

Depends on #2578 and AzureAD/microsoft-authentication-library-common-for-android#3254.

Fixes [AB#3754595](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3754595)

### DIFF
--- a/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
+++ b/msal/src/main/java/com/microsoft/identity/client/internal/CommandParametersAdapter.java
@@ -67,6 +67,7 @@ import com.microsoft.identity.common.java.nativeauth.commands.parameters.MFAChal
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.MFASubmitChallengeCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.NativeAuthV2ResendCodeCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.NativeAuthV2SelectMFAMethodCommandParameters;
+import com.microsoft.identity.common.java.nativeauth.commands.parameters.NativeAuthV2SelectResetPasswordMethodCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.NativeAuthV2SignInAfterResetPasswordCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.NativeAuthV2SignInAfterSignUpCommandParameters;
 import com.microsoft.identity.common.java.nativeauth.commands.parameters.NativeAuthV2SubmitAttributesCommandParameters;
@@ -1663,6 +1664,52 @@ public class CommandParametersAdapter {
                         .build();
 
         return commandParameters;
+    }
+
+    /**
+     * Creates command parameters for selecting a V2 reset-password first-factor method.
+     *
+     * @param configuration PCA configuration
+     * @param tokenCache token cache for storing results
+     * @param methodId server-issued ID of the authentication method the app selected
+     * @param continuationState opaque continuation state from the reset-password start response
+     * @return Command parameter object
+     */
+    public static NativeAuthV2SelectResetPasswordMethodCommandParameters createNativeAuthV2SelectResetPasswordMethodCommandParameters(
+            @NonNull final NativeAuthPublicClientApplicationConfiguration configuration,
+            @NonNull final OAuth2TokenCache tokenCache,
+            @NonNull final String methodId,
+            @NonNull final NativeAuthV2ContinuationState continuationState) throws ClientException {
+
+        final NativeAuthCIAMAuthority authority = ((NativeAuthCIAMAuthority) configuration.getDefaultAuthority());
+        final AbstractAuthenticationScheme authenticationScheme = AuthenticationSchemeFactory.createScheme(
+                AndroidPlatformComponentsFactory.createFromContext(configuration.getAppContext()),
+                null
+        );
+
+        return NativeAuthV2SelectResetPasswordMethodCommandParameters.builder()
+                .platformComponents(AndroidPlatformComponentsFactory.createFromContext(configuration.getAppContext()))
+                .applicationName(configuration.getAppContext().getPackageName())
+                .applicationVersion(getPackageVersion(configuration.getAppContext()))
+                .clientId(configuration.getClientId())
+                .isSharedDevice(configuration.getIsSharedDevice())
+                .redirectUri(configuration.getRedirectUri())
+                .oAuth2TokenCache(tokenCache)
+                .requiredBrokerProtocolVersion(configuration.getRequiredBrokerProtocolVersion())
+                .sdkType(SdkType.MSAL)
+                .sdkVersion(PublicClientApplication.getSdkVersion())
+                .powerOptCheckEnabled(configuration.isPowerOptCheckForEnabled())
+                .authority(authority)
+                .authenticationScheme(authenticationScheme)
+                .challengeType(configuration.getChallengeTypes())
+                .requestInterceptor(configuration.getRequestInterceptor())
+                .capabilities(configuration.getCapabilities())
+                .methodId(methodId)
+                .continuationState(continuationState)
+                .scopes(continuationState.scopesForTokenRequest())
+                .claimsRequestJson(continuationState.claimsRequestJsonForTokenRequest())
+                .correlationId(continuationState.getCorrelationId())
+                .build();
     }
 
     /**
--- a/msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
+++ b/msal/src/main/java/com/microsoft/identity/nativeauth/NativeAuthPublicClientApplication.kt
@@ -96,6 +96,7 @@ import com.microsoft.identity.nativeauth.statemachine.states.AwaitingMFAState
 import com.microsoft.identity.nativeauth.statemachine.states.Callback
 import com.microsoft.identity.nativeauth.statemachine.states.RegisterStrongAuthState
 import com.microsoft.identity.nativeauth.statemachine.states.ResetPasswordCodeRequiredState
+import com.microsoft.identity.nativeauth.statemachine.states.ResetPasswordMethodRequiredStateV2
 import com.microsoft.identity.nativeauth.statemachine.states.SignInCodeRequiredState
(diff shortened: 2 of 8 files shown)


antrix1989:
## Comparison with MSAL iOS `dev`

Compared this PR at `3872211ec2bdbcac55d2d64104581081de35ebf1`, including its Common pin `631fa691d6df16d04d2b6ea2bf9aeeb89d94cbbb`, against `AzureAD/microsoft-authentication-library-for-objc` **`dev` at `6f7cbda36d251d1c6e6e0f40334b16059d5b1f20`**. The comparison uses iOS's actual **V2 HAL flow controller**, not the legacy reset-password controller.

**Result: six parity findings below. These are non-blocking behavioral/API differences, not six proven Android bugs.** Some are inherited from the existing V2 implementations and are identified as such; they should not be attributed to newly added lines in this PR.

### 1. SMS risk-verification handling differs

**Difference:** The Common update used here follows `RiskVerificationRequired` with `performRiskVerification()` before returning `CodeRequired`. iOS's V2 selection path sends the selected method challenge and directly handles verification-required; it has no corresponding explicit risk-verification hop in this controller/parser.

**Impact:** A service response requiring that intermediate step can proceed on Android but is not handled equivalently by the inspected iOS V2 flow. Ordinary direct code-required responses still align.

**Recommendation:** Keep the Android support; track the iOS gap separately and use the same SMS challenge/risk/verify response sequence in cross-platform contract tests.

Evidence: [Android challenge/risk handling][android-challenge], [iOS selection][ios-select], [iOS challenge-result handling][ios-challenge].

### 2. A completion response at the method-challenge stage has different outcomes

**Difference:** Android's challenge helper accepts code-required or redirect and maps an unexpected `ReadyToComplete` to an API error. iOS's corresponding handler accepts `.readyToComplete` and returns the sign-in-after-reset state.

**Impact:** The same completion response at this stage produces an error on Android versus a continuation toward sign-in on iOS. Android explicitly documents rejection before a credential has been proven; this is not evidence that its guard should be removed.

**Recommendation:** Confirm the service contract for completion at this exact stage and record the intentional divergence or align the appropriate implementation. Do not relax Android's guard merely to copy iOS. This is an inherited contract difference exercised by the new selector.

Evidence: [Android guard rationale and flow][android-start], [Android challenge handling][android-challenge], [iOS completion branch][ios-challenge].

### 3. Unknown method types can prevent reset on iOS but be ignored on Android

**Difference:** Android parses a well-formed method with any nonblank type, normalizes the type, and then filters the reset list to email/SMS. iOS's method parser recognizes email/SMS/password and rejects the whole response for an unknown type before reset-specific filtering.

**Impact:** For a well-formed list containing one email method plus an unknown future method type, Android automatically challenges email; iOS returns a general error. Both platforms agree for supported email/SMS lists and for filtering a recognized password method.

**Recommendation:** Document or align the forward-compatibility policy and add a shared mixed-supported/unknown-method response fixture. The parser behavior predates this PR; this is a dependency/platform follow-up, not a request to broaden the Android diff.

Evidence: [Android parser][android-method-parser], [Android reset filtering][android-start], [iOS parser][ios-method-parser].

### 4. Missing verification-code metadata is tolerated differently

**Difference:** Android requires a positive `codeLength` and a destination hint before producing `CodeRequired`. iOS's V2 parser defaults a missing length to `0` and a missing hint to `""`.

**Impact:** With an otherwise valid email/SMS verification response and verify link, omitted length or destination produces an Android error but can still produce the iOS code-entry state. For complete service responses, both preserve the returned length/channel/masked destination.

**Recommendation:** Agree which fields the service guarantees and cover omitted metadata in shared response fixtures. Keep Android's validation unless the contract explicitly permits those omissions. This is inherited parser behavior, not a newly introduced metadata-loss bug.

Evidence: [Android verification parser][android-code-parser], [iOS verification parser][ios-code-parser].

### 5. Selection errors have different public classifications

**Difference:** An unoffered method ID returns Android `ResetPasswordErrorV2` with `INVALID_STATE`; iOS returns a unified flow error classified as `generalError`. Android's generic selection API-error mapping leaves `errorType` unset while retaining the server error string; iOS assigns `generalError` and does not expose that raw server-code string in its public error model.

**Impact:** Cross-platform clients cannot use identical error classifications/recovery predicates even for the same selection failure. This is a representation difference, not proof that either platform reliably supports more recovery cases.

**Recommendation:** Document the mapping and cover unknown-ID and service-error outcomes for the new Android selection API. Do not infer missing blocked-method/contact behavior from iOS helper names: although iOS declares those helpers, its inspected V2 parser does not assign those classifications either.

Evidence: [Android selection guards and mappings][android-selection], [iOS selection guards][ios-select], [iOS actual server-error classification][ios-errors], [iOS public error shape][ios-error-model].

### 6. The new public API is platform-specific rather than structurally identical

**Difference:** Android adds reset-specific `ResetPasswordMethodRequiredStateV2` / `ResetPasswordMethodRequired`; iOS reuses `MSALNativeAuthAuthMethodSelectionRequiredState` for reset and MFA. The new Android state also follows Android's existing callback/coroutine and `Parcelable` conventions, whereas the inspected iOS V2 surface uses delegate callbacks and does not expose equivalent public state persistence or a cancellable task handle.

**Impact:** Shared app abstractions need platform adapters. Android coroutine callers can observe propagated cancellation; iOS's public selection API does not provide the equivalent coroutine contract. These are API/lifecycle differences, not reasons to require Swift and Kotlin implementations to be identical.

**Recommendation:** Document the platform mapping, cancellation behavior, and persistence expectations alongside the new result/state.

Evidence: [Android new state][android-selection], [iOS shared state][ios-selection-state], [iOS state execution][ios-runner].

### Confirmed parity / not findings

- Both filter to email/SMS, automatically select exactly one supported method, and wait for explicit selection when multiple supported methods remain.
- Both select by the offered **ID**, not full-object equality. Both ultimately follow the stored method-specific HAL challenge URL; Android's internal `methodId` parameter is **not** a difference in the challenge wire body.
- Both expose browser-required rather than automatically opening a browser after method selection.
- Neither current code-required state exposes method reselection. Keeping an old selection object does not establish that the server will accept its stale token.

**Review summary:** New parity findings: 6. Resolved findings: N/A, first pass. Confirmed blocking Android defects from this comparison: none. Source and existing tests were inspected; builds, tests, and live-service flows were not executed.

[android-start]: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/631fa691d6df16d04d2b6ea2bf9aeeb89d94cbbb/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/v2/NativeAuthV2FlowController.kt#L128-L234
[android-challenge]: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/631fa691d6df16d04d2b6ea2bf9aeeb89d94cbbb/common/src/main/java/com/microsoft/identity/common/nativeauth/internal/controllers/v2/NativeAuthV2FlowController.kt#L265-L294
[android-method-parser]: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/631fa691d6df16d04d2b6ea2bf9aeeb89d94cbbb/common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/v2/NativeAuthV2ResponseParser.kt#L245-L285
[android-code-parser]: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/631fa691d6df16d04d2b6ea2bf9aeeb89d94cbbb/common4j/src/main/com/microsoft/identity/common/java/nativeauth/providers/responses/v2/NativeAuthV2ResponseParser.kt#L330-L377
[android-selection]: https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/3872211ec2bdbcac55d2d64104581081de35ebf1/msal/src/main/java/com/microsoft/identity/nativeauth/statemachine/states/ResetPasswordMethodRequiredStateV2.kt#L58-L235
[ios-select]: https://github.com/AzureAD/microsoft-authentication-library-for-objc/blob/6f7cbda36d251d1c6e6e0f40334b16059d5b1f20/MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift#L434-L499
[ios-challenge]: https://github.com/AzureAD/microsoft-authentication-library-for-objc/blob/6f7cbda36d251d1c6e6e0f40334b16059d5b1f20/MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowController.swift#L1160-L1190
[ios-method-parser]: https://github.com/AzureAD/microsoft-authentication-library-for-objc/blob/6f7cbda36d251d1c6e6e0f40334b16059d5b1f20/MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift#L188-L268
[ios-code-parser]: https://github.com/AzureAD/microsoft-authentication-library-for-objc/blob/6f7cbda36d251d1c6e6e0f40334b16059d5b1f20/MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift#L137-L162
[ios-errors]: https://github.com/AzureAD/microsoft-authentication-library-for-objc/blob/6f7cbda36d251d1c6e6e0f40334b16059d5b1f20/MSAL/src/native_auth/network/responses/v2/parser/MSALNativeAuthV2ResponseParser.swift#L268-L302
[ios-error-model]: https://github.com/AzureAD/microsoft-authentication-library-for-objc/blob/6f7cbda36d251d1c6e6e0f40334b16059d5b1f20/MSAL/src/native_auth/public/state_machine/error/MSALNativeAuthError.swift#L27-L65
[ios-selection-state]: https://github.com/AzureAD/microsoft-authentication-library-for-objc/blob/6f7cbda36d251d1c6e6e0f40334b16059d5b1f20/MSAL/src/native_auth/public/state_machine/v2/state/MSALNativeAuthAuthMethodSelectionRequiredState.swift#L27-L66
[ios-runner]: https://github.com/AzureAD/microsoft-authentication-library-for-objc/blob/6f7cbda36d251d1c6e6e0f40334b16059d5b1f20/MSAL/src/native_auth/controllers/v2/MSALNativeAuthFlowInternalState.swift#L36-L58

djanardhan98 (approved):

diegojerezba (approved):

